PR #87, opened 2013-11-17
Hot-Key (WinKey/Super) support for application launching, switching, etc

anprogrammer:
I've added support for the "Super" key.  Pressing Super+1 switches to (or launches) the first program on the dash, Shift+Super+1 performs the custom shift action when available, and Ctrl+Super+1 always launches a new copy of the program.  This works for Super+1 -> Super+0 going down the dash, much like the similar keyboard shortcuts on Win7.

This is the first time I've worked on a Gnome extension before, so let me know if there are any corrections you would like me to make.

### DIFF
--- a/dockedDash.js
+++ b/dockedDash.js
@@ -26,6 +26,7 @@ function dockedDash(settings) {
 }
 
 dockedDash.prototype = {
+    _numHotkeys: 10,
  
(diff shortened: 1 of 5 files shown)


micheleg:
This feature was merged into master ( [3a55d50] and previous commit).

PR #118, opened 2014-11-11
Fix issue with small gap between expanded dock and top panel

michalmuskala:
When using this extension I found a problem with expanded dock - it wasn't aligned to top panel:
screenshot from 2014-11-11 19 36 43
Strangely, the gap was gone after resetting the expansion option. I figured out there must be some problem with timing. Unfortunately I know nothing about GTK (or any other desktop environment for that matter), but the delay in updating position seems to help with the issue. This was found through some trials and errors, there may be some cleaner way of fixing this, but I think it's beyond my current abilities.

I'll be happy to answer any questions.

### DIFF
--- a/dockedDash.js
+++ b/dockedDash.js
@@ -875,8 +875,13 @@ const dockedDash = new Lang.Class({
         this.actor.move_anchor_point_from_gravity(anchor_point);
         this.actor.x = position;
 
-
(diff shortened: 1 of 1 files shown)


micheleg:
I think the bug the pull request was trying to address is the same reported in #241, thus solved in current master.

PR #125, opened 2014-11-28
an idea for embedding dash-to-dock in the top bar

altgokul:
You may or may not have considered doing this already. I don't consider this complete. It is just something that works for me (not fully but good enough). 

I simply reoriented and re-parented the elements so that they are placed at the centre on the gnome-shell top-bar. I prefer this layout because it saves space and top-bar is closer to the regular window controls and tabs and title bar etc. So placing the dash there seems more space and movement efficient.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 # Basic Makefile
 
-UUID = dash-to-dock@micxgx.gmail.com
+UUID = dash-to-topbar@nowhere.com
 BASE_MODULES = extension.js stylesheet.css metadata.json COPYING README.md
-EXTRA_MODULES = convenience.js dash.js docking.js appIcons.js windowPreview.js intellihide.js prefs.js theming.js utils.js Settings.ui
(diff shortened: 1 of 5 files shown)


user501254:
@altgokul Some screenshots would be nice. :-)

micheleg:
This was the original proposed behaviour. 

I think this is a bit out of the scope of this extension.

jderose9:
I found that ZorinOS had taken this pretty far. Unfortunately they had stripped all the config settings out (to be fair, most of them needed at minor adjustment or even a complete re-implementation). I forked their implementation, fixed a few bugs, added a few things, and started merging the configuration back in..

https://github.com/jderose9/dash-to-panel

It seems like it would be nice to have this feature in dash-to-dock rather than maintain the core functionality separately, but there's a good bit of divergence when it comes to styling and positioning, Plus you need to move the centerBox out of the way and such, and then a lot of features that I think are expected when you integrate with the top bar (re-styling everything to match and being able to reorder all of the panel elements, etc.). Trying to get everything into a single settings UI that makes sense seems pretty ambitious.

micheleg:
I don't think it is impossible to integrate such functionality in a maintainable way, but it would require to reorganize the code and a lot of work which I'm not personally interested in doing right now.

PR #133, opened 2015-01-12
Add support to 3.14.3

Hernou:
It dont work if not on 3.14.3...

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,5 +1,5 @@
 {
-"shell-version": ["3.14"],
+"shell-version": ["3.14", "3.14.3"],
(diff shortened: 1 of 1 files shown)


micheleg:
Does adding the "3.14.3" string really makes a difference between working
and not working for you? This should not happen, as all minor versions
3.14.x should be included if "3.14" is specified. Please confirm it wasn't
another error solved by a restart rather than your changes.

luisfpg:
To me this extension also doesn't work on Gnome shell 3.14.3, on Arch Linux.
Used to work before the upgrade...

PR #137, opened 2015-01-16
Change wording

6ahodir:
The previous message makes sense if the dock is vertical, but not
when it's horizontal. The new message is clearer no matter what
the direction of the dock is.


micheleg:
Merged. Thanks!

PR #175, opened 2015-05-21
Update de.po

prdiewelt:
I have inserted new translations and modified some old ones. May still be improved, also some strings are untranslatable (“Behaviour when clicking…”, “Cycle through windows”, “Webpage”).

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -9,8 +9,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Dash to Dock\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2015-04-17 21:33+0100\n"
-"PO-Revision-Date: 2015-05-21 19:20+0100\n"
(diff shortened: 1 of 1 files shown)


micheleg:
Thank you. Merged into master and gnome-3.14 branches. Can't you really express those concepts anyhow?

prdiewelt:
The problem is that these strings are not in the .po-file and I don't know, how to add them…

micheleg:
If you run make```make mergepo``` the po file should be updated with all newly added strings.

prdiewelt:
Does it work like this? Webpage is still not found, but this is not so important…

micheleg:
My fault, that string was not marked as translatable. It's fixed now in master. Any other string missing? If you send me the updated version I'll merge it.

prdiewelt:
Now everything should be OK.

micheleg:
Perfect. Merged.

PR #205, opened 2015-07-29
Add Serbian Cyrillic and Latin translation (files sr.po and sr@latin.po)

Faenriis:
Serbian translation in cirillic and latin variant (sr.po and sr@latin.po). Please merge.


micheleg:
Thank you for the translations. Everything has been merged.

Issue #202, opened 2015-07-27
Gnome Shell API changes causes breakage error


nikolowry:
Running the lastest git master of Dash-to-dock, GJS, Gnome-Shell, Clutter and Mutter and was saddened to see an API change has now broken Dash-to-dock.  The error being emitted is "no 'maximize' on object ShellWM"

I had to comment out the 'maximize' and 'unmaximize' from https://github.com/micheleg/dash-to-dock/blob/master/intellihide.js#L76 to get it back functioning.

micheleg:
Thank you very much for testing the extension on the upcoming versions of gnome. You save me from a lot of work.... I'll have a look at it. Usually I'm not supporting the developing version until at least the beta version comes out.

These seem to be relevant upstream commits:
https://git.gnome.org/browse/mutter/commit/?id=9f04c58ffed2f5c59731831ddc6823caef5d59eb
https://git.gnome.org/browse/gnome-shell/commit/?id=7305466765395e555278a3b26e9017387edb3a74
https://git.gnome.org/browse/gnome-shell/commit/?id=fe265554a7563d91d00525e35e43a3ff3e57effe

micheleg:
Solved in the gnome-3.17 branch.

nikolowry:
Quick work! Confirming everything is working.

PR #206, opened 2015-08-01
Update intellihide.js for gnome shell 3.17

Haringat:
This should solve #202 but will likely break compatibility for 3.16 (and lower).


micheleg:
Thank you very much. I can't test the patch but it  looks ok according to the upstream changes. I've created a gnome-3.17 branch that will become master after gnome-3.18 release and merged the pull request there.

PR #207, opened 2015-08-03
Update ru.po

vantu5z:
Updated Russian translation and fixed some misprints.


micheleg:
Merged into master [8206b33332]. Thank you.

PR #240, opened 2015-10-16
Add explicit GNOME shell 3.18.1 support.

delcypher:


micheleg:
As far as I know it is not necessary to mark all point versions for stable release, "3.18"  should already include it. Did you experience the need of explicitly declare it? Can you confirm it?

delcypher:
I just upgraded to Gnome 3.18.1 on my Arch Linux on one of my boxes. I tried uninstalling the extension and installing it again via the Gnome Extensions website but it complained that the extension was incompatible too.

So I cloned this repo then installed the extension but it still  didn't work. Doing this seemed fixed the issue.

The really weird thing is I have another Arch Linux machine which I did the same upgrade on and the extension kept working fine.

I think I may of have forgotten to reload the shell one the machine that extension didn't work on. Let me check...

delcypher:
My mistake. Looks like I forgot to reload GNOME shell when I was experimenting on my box. The version in the master branch works fine when I install it and reload the shell.

Sorry for the noise.

PR #250, opened 2015-11-15
Update Simplified Chinese translation.

shlinux:
make myDash.js can be translated.
Fix a spelling mistake. 
Update Simplified Chinese translation.
Thanks.

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -1242,7 +1242,7 @@
                             <property name="can_focus">False</property>
                             <property name="hexpand">True</property>
                             <property name="xalign">0</property>
-                            <property name="label" translatable="yes">Few customizations meant to integrate the dock wih the default GNOME theme. Alternatively, specific options can be enabled below.</property>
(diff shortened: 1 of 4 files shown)


micheleg:
Hi,

Sorry for the late reply and thanks for spotting the problem with the translations with the strings in myDash.js. 

I decided to solve the issue by keeping the strings in myDash.js translated with the default domain, i.e. reusing the translations already provided by gnome-shell. This is more for simplicity and practical reasons to avoid mixing gettext domains in the same files (or, I can't find a  quick and clean way to do it now).

I merged your fix and translation (db0cc2eae0bf2112ef87503dc127c0f6f964ff78, af12e74c6bf112b5cb80d03631e3f78b5d781d7b, f76b875b4015035ae9d2b825f8d23242f528bc67), I hope nothing has been lost while rebasing.

shlinux:
That is all right.
I'll check and remove the "fuzzy" in the po file.
And I think maybe it's better like this
```javascript
let item = this._appendMenuItem("Dash to Dock " + _("Settings"));
```
What do you think about it ?

micheleg:
Yes, I updated the menu label as you suggested [593b6c53e]

PR #282, opened 2016-01-30
Fix restore default settings in intellihide mode

user501254:
This is a patch to allow restoring the `pressure-threshold` back to its default value.


micheleg:
Perfect, thank you! Merged into master [093a77f] and backported [27d29f3, 1b3b05b].

PR #284, opened 2016-02-15
Enable in fullscreen 316

dmarrazzo:
Hi, please add this feature!
Thank you


micheleg:
Hi,

thanks for the suggestion and code. There is a little problem. When  windows are full-screen the dash visibility is not updated. This has to be solved before merging. Temporary adding an visibility update when windows are in fullscreen - there might be a signal to connect to somewhere - should be enough.

micheleg:
It should work ok now, can you test it?

micheleg:
I  decided to disable this behaviour by default and added an option in the intellihide setttings. Almost ready to merge.

dmarrazzo:
Ok great news! Thank you

micheleg:
Merged in master [63c6f9f17], 3.16 [92d8728]  and 3.14 [e034700] branches.

PR #299, opened 2016-03-22
Update french translation

Sancho-Panza:


micheleg:
Thanks a lot. Merged into master [4005eac].

PR #300, opened 2016-03-27
Code style cleanup, support for window stealing, support for right-click for settings

tliron:
This seems like a very massive pull request, but that's only because of the code-style cleanup. :) Actually, the change is just adding one file and a few smaller interventions.

**Cleanup involved:**

* Using more and smaller JS files, organized by "feature" (to be more like the Shell's codebase).
* Standardizing on one specific JS style (there were many different styles before: different use of spaces, indents; sometimes single line blocks had brackets and sometimes not; different styles of comments; different standards for camel-case; various conventions for GObject class names; and more).

**Window stealing feature:**

IMPORTANT: Let me say that I'll be happy to continue supporting this feature. And actually generally help you fix bugs in Dash to Dock, now that I know the codebase well. I'm not just dumping a feature on you, I really want to collaborate and help. :)

* The feature is disabled by default, but users can enable it in prefs, in the "behavior" pane.
* When enabled: an extra "Window Stealing" entry is added to the app icon popup menu. It entry will open a simple configuration dialog box that lets you enter a list of WM_CLASS names (or leave empty to remove stealing).
* Once set, those windows will now be "stolen" by this icon -- the icon show in the active style if any of these windows are open, and will have the correct number of dots. Click to minimize/maximize will also take these windows into consideration. Relatedly, the extra running app icons will not appear for those windows.

* Known bug: if you de-configure window stealing for an icon, the icon will immediately stop stealing, but the extra running icons for those windows will not reappear until restarting the Shell. This is rather hard to solve, and is also special edge case that would likely only appear during testing, so I don't think it should affect releasing the feature.

* Another tiny bug: window minimizing/maximizing animations originate at the location of the now-invisible stolen app icon on the dash, rather than the stealing one. This doesn't affect usability, of course, but is something I will try to improve.

The code is all in windows.js, however it also affects some code in dash.js and icons.js. Specifically, instead of calling app.get_windows, we are now calling Windows.getAllWindows which knows how to handle window stealing (if enabled). We are additionally adding some simple hooks to setting changes to make sure to refresh the icons.

**Right-click for settings:**

This adds to ability to right-click anywhere on the dash (outside the icons) and get a simple popup menu that lets you run the settings. This is very useful for users who don't have the applications icon enabled (otherwise they have no easy easy to get to settings).

The code is in dash.js line 81. Very straightforward!

### DIFF
--- a/convenience.js
+++ b/convenience.js
@@ -70,7 +70,9 @@ function getSettings(schema) {
         throw new Error('Schema ' + schema + ' could not be found for extension '
                         + extension.metadata.uuid + '. Please check your installation.');
 
-    return new Gio.Settings({ settings_schema: schemaObj });
+    return new Gio.Settings({
+        settings_schema: schemaObj
+    });
 }
 
 /**
@@ -84,11 +86,10 @@ const BasicHandler = new Lang.Class({
         this._storage = new Object();
     },
 
-    add: function(/*unlimited 3-long array arguments*/) {
-
-        // convert arguments object to array, concatenate with generic
+    add: function(/* unlimited 3-long array arguments */) {
+        // Convert arguments object to array, concatenate with generic
         let args = Array.concat('generic', Array.slice(arguments));
-        // call addWithLabel with ags as if they were passed arguments
+        // Call addWithLabel with ags as if they were passed arguments
         this.addWithLabel.apply(this, args);
     },
 
@@ -97,38 +98,39 @@ const BasicHandler = new Lang.Class({
             this.removeWithLabel(label);
     },
 
-    addWithLabel: function( label /* plus unlimited 3-long array arguments*/) {
-
-        if(this._storage[label] == undefined)
+    addWithLabel: function(label /* plus unlimited 3-long array arguments*/) {
+        if (this._storage[label] == undefined)
             this._storage[label] = new Array();
 
-        // skip first element of the arguments
-        for( let i = 1; i < arguments.length; i++ ) {
-            this._storage[label].push( this._create(arguments[i]) );
+        // Skip first element of the arguments
+        for (let i = 1; i < arguments.length; i++) {
+            this._storage[label].push( this._create(arguments[i]));
         }
-
     },
 
     removeWithLabel: function(label) {
-
-        if(this._storage[label]) {
-            for( let i = 0; i < this._storage[label].length; i++ ) {
+        if (this._storage[label]) {
+            for (let i = 0; i < this._storage[label].length; i++)
                 this._remove(this._storage[label][i]);
-            }
 
             delete this._storage[label];
         }
     },
 
-    /* Virtual methods to be implemented by subclass */
-    // create single element to be stored in the storage structure
+    // Virtual methods to be implemented by subclass
+
+    /**
+     * Create single element to be stored in the storage structure
+     */
     _create: function(item) {
-      throw new Error('no implementation of _create in ' + this);
+        throw new Error('no implementation of _create in ' + this);
     },
 
-    // correctly delete single element
+    /**
+     * Correctly delete single element
+     */
     _remove: function(item) {
-      throw new Error('no implementation of _remove in ' + this);
+        throw new Error('no implementation of _remove in ' + this);
     }
 });
 
@@ -140,17 +142,16 @@ const GlobalSignalsHandler = new Lang.Class({
     Extends: BasicHandler,
 
     _create: function(item) {
+        let object = item[0];
+        let event = item[1];
+        let callback = item[2]
+        let id = object.connect(event, callback);
 
-      let object = item[0];
-      let event = item[1];
-      let callback = item[2]
-      let id = object.connect(event, callback);
-
-      return [object, id];
+        return [object, id];
     },
 
     _remove: function(item) {
-       item[0].disconnect(item[1]);
+         item[0].disconnect(item[1]);
     }
 });
 
@@ -163,14 +164,13 @@ const InjectionsHandler = new Lang.Class({
     Extends: BasicHandler,
 
     _create: function(item) {
+        let object = item[0];
+        let name = item[1];
+        let injectedFunction = item[2];
+        let original = object[name];
 
-      let object = item[0];
-      let name = item[1];
-      let injectedFunction = item[2];
-      let original = object[name];
-
-      object[name] = injectedFunction;
-      return [object, name, injectedFunction, original];
+        object[name] = injectedFunction;
+        return [object, name, injectedFunction, original];
     },
 
     _remove: function(item) {
--- a/convenience.js
+++ b/convenience.js
@@ -3,17 +3,15 @@
 /*
  * Part of this file comes from gnome-shell-extensions:
  * http://git.gnome.org/browse/gnome-shell-extensions/
- * 
  */
-
 
 const Gettext = imports.gettext;
 const Gio = imports.gi.Gio;
+const Clutter = imports.gi.Clutter;
 const Lang = imports.lang;
 
 const Config = imports.misc.config;
 const ExtensionUtils = imports.misc.extensionUtils;
-
 
 /**
  * initTranslations:
@@ -27,7 +25,7 @@ function initTranslations(domain) {
 
     domain = domain || extension.metadata['gettext-domain'];
 
-    // check if this extension was built with "make zip-file", and thus
+    // Check if this extension was built with "make zip-file", and thus
     // has the locale files in a subfolder
     // otherwise assume that extension has been installed in the
     // same prefix as gnome-shell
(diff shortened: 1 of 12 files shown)


micheleg:
Thanks. I wanted to try it, and I noticed that you didn't update the Makefile. Just list the required files, changing those you renamed and adding the additional ones in the EXTRA_MODULES  variable, to start. That should be enough.

tliron:
Oops, you're right. :/

tliron:
Fixed.

tliron:
I see that you just made some major changes! I'd be happy to do the work to merge them into my pull request, but I want to make sure you actually have interest in it. I do think it really does a lot to clean up your codebase and make it friendlier to contributors.

micheleg:
Hi,

I just fix the inevitable bug after the latest release, no major changes, nor any is planned to land soon. I've just had the time to run your version (Settings.ui is still missing in the Makefile by the way). I'm looking right now to your changes, but it will take me some time. Although merging the few changes should be trivial, there's no need for you to do it right now. 

Regarding your changes, the first impact with the "windows stealing" functionality has been "now what do I write here? What's a the WM_CLASS...", while I'm not able to access the settings right clicking on the dash.

Michele

convenience.js
```diff
@@ -3,18 +3,16 @@
 /*
  * Part of this file comes from gnome-shell-extensions:
  * http://git.gnome.org/browse/gnome-shell-extensions/
- * 
  */
 
-
 const Gettext = imports.gettext;
 const Gio = imports.gi.Gio;
+const Clutter = imports.gi.Clutter;
```
micheleg: Should be before (in alphabetic order).

convenience.js
```diff
@@ -176,3 +180,17 @@ const InjectionsHandler = new Lang.Class({
         object[name] = original;
     }
 });
+
+/**
+ * Return the actual position reverseing left and right in rtl
+ */
+function getPosition(settings) {
```
micheleg: Not sure this is the place for this function

dash.js
```diff
@@ -779,17 +650,17 @@ const myDash = new Lang.Class({
...
     },
 
-    _updateAppIconsGeometry: function() {
+    _updateAppIcons: function() {
         let appIcons = this._getAppIcons();
-        appIcons.forEach(function(icon){
-            icon.updateIconGeometry();
+        appIcons.forEach(function(icon) {
+            icon.onWindowsChanged();
         });
```
micheleg: * _updateAppIconsGeometry was the name in the windows-lits extension, because it chenge the icon_geometry property of the metawindow. Why changing the name?

* Is there a reason for calling onWindowsChange? I prefer not to call something which is clearly meant to be a signal callback from here.

icons.js
```diff
@@ -0,0 +1,547 @@
+// -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
```
micheleg: Maybe this file should be named appIcons.js?

icons.js
```diff
@@ -0,0 +1,547 @@
...
+        // If (for unknown reason) the actor is not on the stage the reported size
+        // and position are random values, which might exceeds the integer range
+        // resulting in an error when assigned to the a rect. This is a more like
+        // a workaround to prevent flooding the system with errors.
+        if (this.actor.get_stage() == null)
+            return
+
+        let rect = new Meta.Rectangle();
+
+		if (Windows.isStolen(this.app)) {
```
micheleg: Use spaces, not tabs.

micheleg:
At the moment I'm trying to split the code reorganization (which once refined I want to merge as soon as possible) from the additional features (which might require more thinking). This will also help backporting to previous gnome shell versions.

The code really required some clean up and a better consistency: very nice work. Out of curiosity, did you use any tool to enforce a specific code style?

icons.js
tliron: Sure thing. But since we don't have any other kinds of icons, I thought of a simpler name. :)

convenience.js
tliron: Also not sure. But it is used by a few others files, so it must be in some shared third file.

dash.js
tliron: It is necessary to call onWindowsChange here because window stealing could result in changes beyond just geometry. (Without it, there are bugs with window stealing.) I changed the name because I added the onWindowsChange call, and so it was no longer just about geometry.

tliron:
Hey, I see that you are commenting on actually an older version ... I might various fixes since then, please check my latest master. Not sure if github is having a bug here...

I did all the code style by hand because I didn't want to mess anything up with an automatic tool. It was a lot of work. :/

The question about how window stealing works: you can use the `xprop WM_CLASS` to find out the class of a window and experiment with that. Ideally I would want that incorporated into the dialog, but that's more work for the future.

For example, if you are say launching Minecraft with a shell script, then you need to steal the resulting window, otherwise you get a mess on your dock. For the latest version of Minecraft the WM_CLASS is "Minecraft 1.9". I've also tested it with running remote applications using `ssh -X`.

micheleg:
Hi,

A quick update: I think I have condensed your style and reorganization changes (included your latest ones) in this commit https://github.com/micheleg/dash-to-dock/commit/6a651a7da13876b6094c99640f1f7203c9567f26, isolating the changes related to additional features in the following commits.

I still have to double check I haven't left dirt around and that there are not major unwanted changes left, and consider few other style changes. After that this will be merged into master (after proper rebase) and I will look at your additional features in more detail.

tliron:
Looks good! I guess I should have done it this way, too, but the way it happened was that working on one thing turned into another...

intellihide.js
```diff
@@ -221,39 +200,36 @@ const Intellihide = new Lang.Class({
...
-
                 this._topApp = this._tracker.get_window_app(topWindow);
                 // If there isn't a focused app, use that of the window on top
                 this._focusApp = this._tracker.focus_app || this._topApp
 
                 windows = windows.filter(this._intellihideFilterInteresting, this);
 
-                for(let i=0; i< windows.length; i++) {
-
+                for (let i in windows.length) {
```
micheleg: It tool me a while to find the source of a bug: can you confirm that this is unintended?

tliron: Correct, unintended! Oops, I really did not intend to change any code. Probably it was a bad copy/paste. I apologize for wasting your time trying to find the bug. :/

micheleg: No problem. The reason It's taking a while to merge the changes is that I'm trying to to be sure to not introduce bugs with these huge commits, as it will be very difficult to spot the difference in the future.

micheleg:
Sorry if it's taking me a long time to merge, I haven't had much time lately.

tliron:
There is no rush at all. :) Please note that I made some more small fixes to master. Basically, anything that you want to fix just let me know and I can do that, so that in the end you can just push "merge" and have it done.

micheleg:
I've merged into master the first big code refactoring commit [f9b4bab12b67179fbdd325b1822dfbdd7392a576]. I'm confident everything is consistent and there are no bugs! I know I will be proven wrong, but I don't want to leave it unused for too long or it will be lost. Thanks again for the great work. Additional changes will follow.

tliron:
Terrific! I'm traveling now but will double check next week.

tliron:
Well, it's been a long time since I touched this. But I wonder if you're still interested at all in this feature? Right now there are various launchers that don't behave properly in the dock: you launch them with one icon, and they create new icons for every window. I don't think it's hard to find a solution by gathering all such windows by their window name. As it stands, this is a glaring bug in otherwise terrific software.

If you want to see an example of such a launcher, try installing the [itch.io client](https://itch.io/docs/itch/installing/linux/). Add itch.io to your favorites, and you'll see that when you launch it, it creates a new icon.

micheleg:
Hi @tliron,

sorry for the late reply and thanks for pinging on this issue. I understand the problem and also experience similar issues  on a daily basis with windows not being grouped correctly. 

I'm not quite sure if this is the solution to the problem, but then, what else can we do at the extension level? I'm wondering if we should just have a simple interface and some documentation for it. At the moment the GUI is nice up to the point it asks for the windows classes, which most people -- me included -- have no idea how to easily obtain. A nice user-friendly interface would probably require way too much effort, but it could be considered together with issue #445, as a general way to workaround this and other windows management bugs.

I don't have much time to dedicate to this issue now, so I can't make promises.  I would appreciate if you could just isolate your patches and possibly rebase on master so that in future it will make it  easier for me to go back and work on it (right now they still mixed with your  work on code clean up).

tliron:
Sounds good. OK, so I think it should be separated into two PRs:

1) One PR that implements the feature _without_ a GUI. To enable/test it you'll need to edit the gsettings directly.
2) Assuming everything works with that, we can move to a second PR that adds a nice GUI.

micheleg:
That would be ideal, thank you.

Issue #314, opened 2016-05-15
Customize border opacity following the dash background value.


mankurt:
Dock border should get same opacity as the rest of the dock, it would look much better IMHO.

Sancho-Panza:
Which border are you talking about ? I don't see any border:

![capture du 2016-05-15 20-14-10](https://cloud.githubusercontent.com/assets/16472988/15275944/b3c9449e-1ad9-11e6-8282-bdced2b280ca.png)

mankurt:
![c2f6af4c-c845-11e5-8ff9-26c13fc7d591](https://cloud.githubusercontent.com/assets/15964583/15275965/a2410618-1ae2-11e6-9ec6-78a41d34ba3a.png)

Sancho-Panza:
Ah yep, you're right, I tried with 0% opacity and saw the border, my bad !

Sancho-Panza:
As a workaround, you can edit the file:

`~/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/stylesheet.css`

On line 4 you should have something like this:

`border:1px;`

Just replace it with:

`border:0px;`

Then disable/enable the extension.

Result:

![capture du 2016-05-15 20-48-38](https://cloud.githubusercontent.com/assets/16472988/15276112/88ef6dd4-1ade-11e6-8646-1b473dac2ad2.png)

mankurt:
Will do, thanks mate!

Sancho-Panza:
You're welcome.

@micheleg: I guess a better way to fix this would be to add a new line after line 84 to set border-color with the same value than background-color ; but idk if everyone would like this feature, so, it means adding a GUI option but I don't know the extension well enough to propose a patch...

micheleg:
I guess it would make sense to set the same opacity value used for the bakground to the border as well. From a very quick test it seems this should work.  It should not be too complicated, it's a matter of extending the code which retrieve and update the background color for the border too. The relevant code is in https://github.com/micheleg/dash-to-dock/blob/master/theming.js if someone wants to propose a patch.

Sancho-Panza:
I'll propose a patch next week.

franglais125:
I took the liberty of looking into this and prepared a patch. https://github.com/franglais125/dash-to-dock/commit/0393fb9e2ea5cd62940b417f9e952ea13831c418

It's in fact just two lines. I use the same information from the background colour to set the border colour.
[...]

PR #316, opened 2016-05-20
Customize border opacity following the dash background value

passingthru67:
This is the modified patch started by @franglais125 to fix the border opacity issue #314.


franglais125:
For what it's worth, I tested his patches and it works well, as intended.

micheleg:
Merged. Thank you.

Issue #313, opened 2016-05-11
How to change background color?


alsoijw:
How to change background color?

micheleg:
Hi,

there's no option to change the dash background. I'm leaving the styling to the theme authors. However, you can tweak the color manually with a css rule . Have a look at the penultimate line here:  https://github.com/micheleg/dash-to-dock/blob/a82c45213d799f27c68f2c02417b418d9303659e/stylesheet.css

This file is installed in:
```
~/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/

```
You can change the color there and reload the extension.

alsoijw:
How about add GUI setting for this?

micheleg:
It's not in my priority list, as I'm inheriting the color from the theme, but I could add this functionality in the future. But I can't promise anything.

SeventhSonOfASeventhSon:
An option allowing changing of the background colour would be nice.

Sancho-Panza:
Coming soon, see [this PR](https://github.com/micheleg/dash-to-dock/pull/318). ;)

micheleg:
This functionality has now been merged into master [7db860c326567b4d7d17dc3c3fc655543e84e61b].

PR #318, opened 2016-05-22
Implementation to set the dash background color.

franglais125:
This adds the ability to set the background color for the dash from the
settings.

Issue reference: https://github.com/micheleg/dash-to-dock/issues/313


franglais125:
screenshot from 2016-05-22 14-42-42

I made it compatible with the opacity setting, so one can independently use both of them.

micheleg:
Hi,

thank you for your pull request. It seems to work very well and it is not invasive neither from the code  nor from the ui point of view. I would made just a change, moving the background color settings ui before the opacity one. After that, I'm ready to merge it. 

Would you mind rebasing the pull request to current master and updating the settings ui?

Thank you,
Michele

franglais125:
Hi Michele,

Thanks for looking into this patch. I rebased and now the branches are now compatible.

I moved the option for the color above the opacity one, as requested!

I know it's a bit off-topic here, but while we are at it... Do you want me to open a PR for the Workspace isolation issue?
I have actually prepared three branches for that, in case you want to backport to Gnome 3.14 and 3.16.

Regards

screenshot from 2016-05-27 18-42-29

franglais125:
Oops, looking at the previous screenshot I realized the switches didn't have an even spacing. Patch b2b3eef562a22969201466334811c032ee4ba361 takes care of that!

screenshot from 2016-05-27 18-47-25

micheleg:
Merged [7db860c326567b4d7d17dc3c3fc655543e84e61b]. Thank you.

Issue #204, opened 2015-07-27
Show only windows of current workspace


2i:
I think it would improve the usability of the extension if it had an option (in the settings) to only show applications on the current workspace in the dock.

The scenario which is inconvenient without the feature is: when you click on an icon of an application which is already running but has no window on the current workspace then you will be navigated to the the workspace which contains such a window. But if you use workspaces to work on separate things you rather would like to open a new window of the application (for example a browser window). The navigation to another workspace is distracting and you have to find your previous workspace again.

To accomplish the desired behavior without the feature you have to right click on the application icon and click on “new window” if the application is not already open on the current workspace. Otherwise you have just to left click on the icon. Currently you cannot see in the dock whether the application is already open on the current workspace, so you have to remember.

The feature would also reduce the number of icons shown in the dock and so improve lucidity.

Favorite applications in the dock instead should be shown on every workspace in the dock in my opinion.

micheleg:
As far as I know this extension is compatible with dash to dock and does part of what you are asking https://extensions.gnome.org/extension/887/workspace-isolated-dash/.

The main problem I see with your suggestion is that there's no way to know if clicking will bring up an older window or a new one. I f you want a new window you can always use ctrl+click.

This is partialy a duplicate of #194.

2i:
In my opinion clicking on an icon of an application should open a new window if (and only if) there is no window on the current workspace of the same application. In the other case, an existing window of the application on the current workspace should be brought to front (like it is already if there is only one workspace).

Thanks for the link to the extension. Principally it is exactly what I was think of (except the icons changing the order sometimes, but this is another bug). But I think, bringing an option into D2D would improve it even more. ;)

fabrizio11:
I agree with this feature request. 

[Workspace-isolated-dash](https://extensions.gnome.org/extension/887/workspace-isolated-dash/]workspace-isolated-dash) works fine **but** it keeps showing opened windows of other workspaces **if** those are windows of a favorite application. So for me isn't a nice solution at all :stuck_out_tongue_closed_eyes:

micheleg:
Hi,

I'm open to consider this option, provided that the feature can be implemented in a clean and maintainable way. However I have no time to work on it now. I'm happy to consider a pull request and to provide guidance for the implementation if anybody wants to look into it.  The  Workspace-isolated-dash extension might be a good starting point to know which things to touch but, as noticed by @fabrizio11, more details have to be tweaked.

[...]

PR #325, opened 2016-05-30
Workspace isolation

franglais125:
Hi,

Here is the PR to introduce workspace isolation (issue https://github.com/micheleg/dash-to-dock/issues/204). As you can see from the work, I tried to keep this branch up to date with yours to avoid problems merging.

I took some time before creating a PR, as I wanted to test the code quite a bit before that. The patches included here work reliably on Gnome Shell 3.18 and 3.20.

I took the time of updating two more branches, so that if you want to back-port this feature it goes faster for you:

- for 3.14: https://github.com/franglais125/dash-to-dock/tree/gnome-3.14 (tested extensively)
- for 3.16: https://github.com/franglais125/dash-to-dock/tree/gnome-3.16 (work in progress, @nsrosenqvist has been helping with the tests).

Regards, and thanks for looking into this!


appIcons.js
```diff
@@ -166,7 +166,13 @@ const MyAppIcon = new Lang.Class({
 
     _updateRunningStyle: function() {
         this.parent();
```
franglais125: We hide the dots for apps that are not active in the current workspace.
Then we need to tell the AppSystem that changes were introduced.

appIcons.js
```diff
@@ -500,7 +510,7 @@ const MyAppIconMenu = new Lang.Class({
             this._quitfromDashMenuItem = this._appendMenuItem(quitFromDashMenuText);
             this._quitfromDashMenuItem.connect('activate', Lang.bind(this, function() {
                 let app = this._source.app;
-                let windows = app.get_windows();
```
franglais125: This change modifiess the "Quit" button, to only quit windows in the current workspace

dash.js
```diff
@@ -699,7 +699,20 @@ const MyDash = new Lang.Class({
     _redisplay: function() {
         let favorites = AppFavorites.getAppFavorites().getFavoriteMap();
 
-        let running = this._appSystem.get_running();
+        let running;
+        if (this._dtdSettings.get_boolean('isolate-workspaces')) {
+            running = this._appSystem.get_running().filter(function(_app) {
```
franglais125: The filtering for the running Apps looks a bit messy.
The reason is that we first check if nautilus is being used for Desktop icons. If we don't do this, when a Nautilus window is open on a given workspace (and it's a favorite), it will be shown as active everywhere.

docking.js
```diff
@@ -1375,6 +1376,47 @@ const DockedDash = new Lang.Class({
             else
                 return false;
         }
+    },
+
+    /**
+     * Isolate overview to open new windows for inactive apps
+     */
```
franglais125: This extension of the Overview is simply to open a new window of a given App, when this App is already active in a different workspace. Example of the problem without this extension:

   - open gedit, then move to another workspace
   - open the overview and type "gedit"
   - hit enter to open the app
   - you will be shown the previous gedit window (in the other workspace).

micheleg:
Hi. I've quickly tested your patches and they seems to be working well. Great work. Few random comments:

* You can probably can use the helper class to manage overwritten methods and function: https://github.com/micheleg/dash-to-dock/blob/master/convenience.js#L158. I see that you have to overwrite this behaviour.

* I'm not sure any more if the current settings UI makes sense, I need to think about it

* for the backports: I guess you are actively using 3.14 (I am too). I don't think there should be major differences related to this feature, beside the codebase which is different bacause it has not been reorganized as in master. I'm not totally sure I will backports this future in older versions. In any case, this will be done after backporting all other changes now in master. Therefore it's probably not worth putting too much effort in  backporting it now. In any case, the work is appreciated as it could reveal some incompatibility.

I'll try to have a look more deeply to you patches soon.

franglais125:
Hi Michele,
Thanks for the suggestion!

- With this patch [518f0a76771006abe2b8d58b6afb4efd1f3ecf25] I now make use of the injectionsHandler. It looks nicer. Also, while preparing this I realized the overwritten function was lost when disabling/enabling the Dash-to-dock extension. Now it's fixed!

- Let me know if you want to change something about the UI.

- As for the backports, it probably makes sense not to backport this. I was just doing it preemptively. I use Debian stable (Gnome 3.14), so I might be biased here :).

Cheers

appIcons.js
micheleg: I'm not sure I understand:
1. Shouldn't dots be already hidden as no interesting windows are returned by ```getInterestingWindows```?
2. I totally don't understand the purpose of ```Shell.AppSystem.get_default().emit('installed-changed');```
3. I also don't understand the reason why focus should not be affected

franglais125: Hi, sorry it took so long to answer.

1. That is correct in the case where we are using the windows counter (or the built-in theme option). If we are not using them, we still need to hide the "dots" (or bars, glow, etc...). Here is a video of what happens if we don't hide
them:
https://www.youtube.com/watch?v=1nuSk8IcenI

2. We need to tell the system that there was a change in the app list when we switch to another workspace. If we don't, then the list of running apps (non-favorites) is not refreshed properly. Here is a video of what happens: https://www.youtube.com/watch?v=E2vbn6vXd2k

3. When switching workspace, the focus is not refreshed unless we force it. This happens only for non-favorite apps.
Here is a video showing the problem:
https://www.youtube.com/watch?v=5o7s6VFAX20

I prepared all videos on Debian Testing (Gnome Shell 3.20). However, I tested the code (or variations of it, due to changes in the code!) on Gnome Shell 3.14, 3.20 and 3.18 to some extent.

Some of the fixes might not be super elegant, but from the use I had, not including them really feels buggy.

Thanks for looking into it!

franglais125: Hi again,

I went ahead and analyzed a bit more the signal handling... Here is a patch [f744dd66b5fdf04653a5d58da916483014b74a22] to better address this issue. I think it is more organized this way (and I now handle signals properly...).

Effectively, I moved `Shell.AppSystem.get_default().emit('installed-changed');` to docking.js, so that it has a proper initialization, and disconnection.

appIcons.js
```diff
@@ -166,6 +166,11 @@ const MyAppIcon = new Lang.Class({
 
     _updateRunningStyle: function() {
         this.parent();
+        if (this._dtdSettings.get_boolean('isolate-workspaces')) {
```
micheleg: Ok. Can you please add he explanation of what we are doing with a comment? 

Also, to be more explicit I think this should be (please check the code is correct):

```
if (this._dtdSettings.get_boolean('isolate-workspaces')) {
        if (this.app.state != Shell.AppState.STOPPED && this.app.is_on_workspace(global.screen.get_active_workspace())
            this._dot.show();
        else
            this._dot.hide();
        this._onFocusAppChanged();
}
else {
        this.parent();
}

File: dash.js (hunk 3 above)
Comment (micheleg) [inline dash.js]: Can't we reuse the `appIcons.getInterestingWindows()` function and filter for `windows!== null`?

Like:

```
let running = this._appSystem.get_running();

if (this._dtdSettings.get_boolean('isolate-workspaces')) {
    running = running.filter(function (.... ) {
           return appIcons.getInterestingWindows !== null;
     })
}

docking.js
```diff
@@ -1375,6 +1376,67 @@ const DockedDash = new Lang.Class({
...
+    _optionalWorkspaceIsolation: function() {
+
+        let OverviewLabel = 'optionalWorkspaceIsolation';
```
micheleg: I don't understand. Why do we need two "namespaces" for the ingjections?

franglais125: We don't need that really. I kept them separate in case you wanted to create two different functions, that way it was easier to spread the code into 2.

Patch [fe2b3151fc2fb545b62121c67f97a8b85167d26a] fixes that!

dash.js
franglais125: Absolutely. Thanks for the suggestion. I had to redeclare 'settings' to use them. See patch [85cfc760aba7f3a5a2c1545d45bed8f1c0d94577].

I was thinking that maybe it would be nicer to write `getInterestingWindows()` so that it takes the boolean instead of the settings.
That way if at some point you need to get ALL windows no matter the settings, you can simply pass `getInterestingWIndows(app, true)`.

In this case, we would write:
`return AppIcons.getInterestingWindows(_app, true).length != 0`

What do you think? Both ways it's fine, just a detail I guess.

appIcons.js
franglais125: I included your lines, they work better (totally missed the `dot.show()`). See patch [8c6f89738e13c14a3d3514af37afdfedbf32337a]

I have a further comment regarding `this._onFocusAppChanged()`. I actually moved it out of the `if`, so that it is always activated.

The problem doesn't show often, but it is present even if we don't use workspace isolation. I don't claim that this is the nicest solution, but for now it works.
I tested this on Gnome-Shell 3.20. If you use the built-in theme (so that the focus style is used) and you open a non-favorite app, focus is not shown (until you change to another window and back).
The same happens when using workspace-isolation and changing workspaces: non-favorite apps don't show focus at first.

Both problems are fixed when including `this._onFocusAppChanged()`. It could actually be moved to this line https://github.com/micheleg/dash-to-dock/blob/master/appIcons.js#L108. That way it would be clearer that we do it for onWindowsChanged.

What do you think?

dash.js
micheleg: I think that passing the settings is more clear than a current parameter and I would keep it the way it is.

micheleg:
Merged into master [7c5aa417a8aff4ea0535542a012dab72e6b1f2fe]. Thank you.

PR #348, opened 2016-07-11
Add an option setting the direction of switching workspace

demokritos:
Workspace-to-dock extension allows the workspaces to lay in horizontal direction. If workspaces are in horizontal direction, switching workspace by scrolling on the dock does not work. This PR adds a checkbox for horizontal workspace switching in the preference and a switcher popup source code from the Workspace-to-dock extension.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 UUID = dash-to-dock@micxgx.gmail.com
 BASE_MODULES = extension.js stylesheet.css metadata.json COPYING README.md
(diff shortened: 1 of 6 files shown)


micheleg:
Hi,

thanks for your contribution. I had a look at the code and everything seems to make sense. However, I'm a bit puzzle about the inclusion of such feature. I'm wondering if there's a way for workspace-to-dock to be more compatible without requiring change in other extensions.

I'll have to think a bit about it.

passingthru67:
@demokritos, I probably should have made the workspaces-to-dock horizontal switching an alternative instead of the default. Anyway, I can add an option that will give users a choice.

demokritos:
OK.
Thanks for your time.

Issue #312, opened 2016-05-09
Rename click action "do nothing" settings to something meaningfull.


micheleg:
The do nothing option really means default behaviour. As reported in #310, it's quite confusing at the moment.

feeef:
Thank you!

micheleg:
Fixed in current master [08bc3629dceaa].

Issue #148, opened 2015-03-24
Expand the 'Customize actions on mouse click' to include middle click to quit and double-click to duplicate (currently undocumented middle click)!!


yeehaw2015:
Hi!!

I'm loving and hugely appreciating your awesome dash to dock, thanks for the current active dev't on it! I'm on gnome 3.14 in deb jessie.

A request: I love the 'middle click to close' idea in computing and have used it to close tabs in browsers and as many proggies I can get my hands on, from my early windows days to my OS X transition and now my enlightened times of Linux :). In Windows I would use tools to modify the taskbar (7+ Taskbar tweaker) to bring the 'middle click to close' action to the taskbar too - it was awesome. Of course, windows is different to the gnome/OS X dock, which groups them per application so on windows it would close 'windows' but here clearly it would have to be 'Quit'.

I see the current behavior for middle click (I'm using v40 from e.g.o), is to duplicate the window ('new instance'), not close.

Might you consider expanding the 'Customize actions on mouse click' area to include a drop-down selection of options for middle click (offering the same sort of offering of options as the current single left click one)? [I guess there's no need to introduce code headache in not having any overlap - user can just have one click action mapped to whatever in the list so it can be easy!] 

For me I'd love middle click to close, and I'm using left click to 'Cycle through application windows' (which is AWESOME btw!!!), and yeah as mentioned in title, other cool idea, if it's not too buggy of a hassle to implement: how about double click as a customiazble action too (and in my case, I'd assign the current 'New Window' action to it, where middle click is currently)?

Oh and if we really want to go to town, how about a fourth action to map to the same master list of actions as the three above - to left double click? 

Thanks micheleg, this is the best extension ever!

~

Also, I have another idea - but maybe it should be a separate request thread: with a customizable timeout setting (from 0.whatever seconds to say 10 seconds), it would be awesome to extend the idea of 'middle click to close', to hovering the mouse on an app icon, and having the windows expand out from it as thumbnail previews (exactly like the things on the overview), where, of course, selecting a thumbnail box will open that window, but especially in the case of my idea, middle click also can close that individual window! So that way you can add granularity to the 'middle click to quit' action by hovering a little longer, and do a less violent 'close by window at a time' thing instead.

[...]

PR #349, opened 2016-07-11
Implementing more options for the different clicks

franglais125:
Hi @micheleg ,

With respect to issues: https://github.com/micheleg/dash-to-dock/issues/148 and https://github.com/micheleg/dash-to-dock/issues/312

With this set of patches I changed the following:
 - Added the "Quit" option
 - Added more options for shift-click
 - Added customization for middle-click and shift+middle-click
 - Changed the name "Do nothing" to "Raise window"

I wanted to bring your attention to two questions before merging this:
 - Is the "Quit" behavior what you expect? This is not a feature I use (e.g. on a Mac), so I don't know what a user should expect. The current implementation closes all windows on all workspaces (except if using workspace isolation).
 - What do you think of the defaults for the middle-click options? I chose "Launch new instance" as I think that's what the Shell does if we don't change anything.

I will add some comments in the code as well.


appIcons.js
```diff
@@ -231,14 +232,32 @@ const MyAppIcon = new Lang.Class({
         let event = Clutter.get_current_event();
         let modifiers = event ? event.get_state() : 0;
         let openNewWindow = modifiers & Clutter.ModifierType.CONTROL_MASK &&
-                            this.app.state == Shell.AppState.RUNNING ||
-                            button && button == 2;
```
franglais125: I removed `button && button == 2` as the behavior might not be "openNewWindow" anymore. Even if it is, the "Launch new instance" will open a new window now anyway..

appIcons.js
```diff
@@ -247,38 +266,39 @@ const MyAppIcon = new Lang.Class({
...
             }
             else {
                 // Activate all window of the app or only le last used
-                if (this._dtdSettings.get_enum('click-action') == clickAction.CYCLE_WINDOWS && !Main.overview._shown) {
+                if (_clickAction == clickAction.CYCLE_WINDOWS && !Main.overview._shown) {
                     // If click cycles through windows I can activate one windows at a time
                     // However, when using isolation, we need to open a new
                     // window if there are no windows in the current WS
                     let windows = getInterestingWindows(this.app, this._dtdSettings);
-                    if (windows.length == 0)
```
franglais125: I had added these lines when implementing workspace isolation.
However, we now check that windows.length > 0 in the (new) line 260, so this is not necessary anymore.

schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
```diff
@@ -6,6 +6,27 @@
...
+    <value value='3' nick='cycle-windows'/>
+    <value value='4' nick='quit'/>
+  </enum>
+  <enum id='org.gnome.shell.extensions.dash-to-dock.shiftMiddleClickAction'>
+    <value value='0' nick='skip'/>
+    <value value='1' nick='minimize'/>
+    <value value='2' nick='launch'/>
+    <value value='3' nick='cycle-windows'/>
+    <value value='4' nick='quit'/>
+  </enum>
```
micheleg: I think there's no need of defining multiple times the enumeration. I think `clickAction` can be used for all of them.

franglais125: Ah, you are right! I just tested it without the repetition.

Fixed in [0cd2bb009f54dd1acb84594fb4b3254a3f239bc0]

micheleg:
I'm running your code and it seems to be working well. To answer your questions:

1. The expected behaviour for click is to close all the windows of the apps, therefore closing the app itself (unless it stays alive as a daemon, like geary for example). I think the current implementation is ok. Which other behaviour woulld you expect (MacOsX?)?
2. The default action of the middle click should be to launch a new instance

I will review your code as soon as possible.

franglais125:
1. I wasn't expecting any other behavior in particular. But since the functions minimizeWindow and activateAllWindows have a workspace filter, I thought that maybe you'd want something like this. For me, it is fine as it is now!
2. Awesome. That's what we have now.

Cheers

Settings.ui
```diff
@@ -1020,19 +1188,41 @@
...
+                            <property name="valign">center</property>
+                            <child>
+                              <object class="GtkImage" id="middle_click_image">
+                                <property name="visible">True</property>
+                                <property name="can_focus">False</property>
+                                <property name="icon_name">emblem-system-symbolic</property>
+                              </object>
+                            </child>
+                            <style>
+                              <class name="prefs-button"/>
```
micheleg: Should have class `circular` (see a6679def3bbafc61).

franglais125: Fixed! [6a6ef41fdf7ee50718f7776df403fafce972cac5]

appIcons.js
```diff
@@ -426,6 +446,15 @@ function activateAllWindows(app, settings) {
...
 
+/**
+ * This closes all windows of the app.
+ */
+function closeAllWindows(app, settings) {
+    let windows = getInterestingWindows(app, settings);
+    for (let i = 0; i < windows.length; i++)
+        windows[i].delete(global.get_current_time());
+}
+
```
micheleg: Now that we have this function, it could be used  also for the quit menu (line 549), and so the `_closeWindowInstance` method (line 528) could be removed.

appIcons.js
```diff
@@ -231,14 +232,32 @@ const MyAppIcon = new Lang.Class({
...
         let focusedApp = tracker.focus_app;
 
         if (this.app.state == Shell.AppState.STOPPED || openNewWindow)
             this.animateLaunch();
 
-        if(button && button == 1 && this.app.state == Shell.AppState.RUNNING) {
+        // We check what type of click we have and if the modifier SHIFT is
+        // being used. We then define what _clickAction should be for this
+        // event.
+        let _clickAction = 0;
```
micheleg: I would not use the `_` prefix or the variable name.

appIcons.js
franglais125: Great idea to simplify the code. Implemented here: [9cb1c49b91ce08cb509e1da9d8497fe5faf69f34]

appIcons.js
franglais125: Yeah, I put the prefix `_` since `clickAction` is already a global variable containing the different actions (line 32).

In this patch [e3a72723ca4750a75c689e894eaf956b60d3754d] I renamed it to `buttonAction`, since after all we first check what kind of button it is.

micheleg:
I found the current logic inside the activate function inpossible to follow after to many additions and I tried to rework it from scratch. You can find it in the branch https://github.com/micheleg/dash-to-dock/commits/click_options. Please rebase on this branch if you are doing additional changes.

Please have a look at it. I might have changed the behaviour. Unless there are bugs in my implementation I'm satisfied with the current code and ready to merge. 

The UI need still a bit of work though:
* I would move  the "shift+click" action to the advanced panel too. 
* Also, I would make cycle through windows the default option for the click.
* The gear button should be placed closer to the swicth (as the others)
* I have some problem settings the quit option in the advance settings, it just doesn't get saved.

franglais125:
Yeah, I fully agree. The logic was becoming unbearable. I had to think in terms of "what happens if the click is ..." in my head.
I'm glad you are proposing this. Next time, if I prepare some PR like this, I'll try to be more proactive (and less shy) about proposing simplification/improvements.

I'll rebase and work out the remaining problems after some testing!

franglais125:
Regarding your last comment:
> The UI need still a bit of work though:
> - I would move the "shift+click" action to the advanced panel too.

Fixed here: 80f36246eb9ebcf640c0a3f39c49c80de75a7cc5.
I added the property `max-width-chars` since the description was too long, making the dialog too wide.

> - Also, I would make cycle through windows the default option for the click.

This is already the default behavior.

> - The gear button should be placed closer to the swicth (as the others)

Fixed here: 8c60228f65aed5c8dd4dbf1bccd08b7ae780cec3. I added a GtkBox for this.

> - I have some problem settings the quit option in the advance settings, it just doesn't get saved.

Fixed here: 7b4b3e40726e43f3dfa7151f6273dc8a2dae6eec. Forgot one element in the `enum`.

appIcons.js
```diff
@@ -271,15 +271,18 @@ const MyAppIcon = new Lang.Class({
             case clickAction.MINIMIZE:
                 // In overview just activate the app, unless the acion is explicitely
                 // requested with a keyboard modifier
-                if (!Main.overview._shown || !modifiers){
-                    if (this.app == focusedApp)
+                if (!Main.overview._shown || modifiers){
```
franglais125: I think you meant `modifiers` and not `!modifiers`? I changed this to get the behaviour described in the comment just above.

appIcons.js
```diff
@@ -271,15 +271,18 @@ const MyAppIcon = new Lang.Class({
...
+                    // If we have button=2 or Shift, allow minimization even if
+                    // the app is not focused
+                    if (this.app == focusedApp || button == 2
+                        || modifiers & Clutter.ModifierType.SHIFT_MASK)
```
franglais125: Here I added `button == 2 || modifiers & Clutter.ModifierType.SHIFT_MASK` as options in the `if`. Not sure if you actually want this, but this is what the behaviour was like before: when NOT using the regular click, one can minimize windows even if they are not focused.

Below I just adjust the indentation to be consistent with the rest of the code.

franglais125:
@micheleg I added another patch [4253164e7aff1eb34fd8b290ce12732d144d85d2]. It removes some invalid properties that were causing warnings in `/var/log/user.log`.

Hopefully the last one!

franglais125:
Well, actually, I just bumped into a minor issue.

In this patch [529304bc07b7c03b9f69b7ca37bb80ea9e27e01b], is there a specific reason we don't check for the `SHIFT_MASK` modier, instead of just `modifiers`?

The problem I'm having is when trying to implement hotkeys on top of this PR (e.g. https://github.com/franglais125/dash-to-dock/tree/experimental_merging ). If using the option "Minimize" for left click, I can't use "Super+number" to un-minimize the app, as the modifier "Super" is activated.

franglais125:
@micheleg Just a ping!

micheleg:
So I've finally merged all commits into master [08bc3629dceaa], I hope I haven't lost anything. 

Thanks a lot again for  your help.I've seen you have quite a few branches with various features. In the future, please, instead of merging the master branch into your changes, try to rebase your development branch onto master (or the relevant branch). This will make the history linear and more importantly allow me to keep track of the changes much more easily.

franglais125:
Thanks a lot for the review and for merging!

I have two remaining branches to review: hotkeys and mult_monitor. I can try and rebase them before trying a new merge.

Perhaps this issue can be closed now: https://github.com/micheleg/dash-to-dock/issues/312.

franglais125:
Hi @micheleg 

There is a small change in how the "Minimize" action behaves, compared to before this merge.

When using the "Minimize" option for the Left Click, the whole app (all windows) used to be minimized. This behaviour was lost, and instead now each window is minimized one by one (except if using double-click). That corresponded to the "Shift+Click" minimize option.

I prepared this patch (rebased on your master), to re-implement this [a96ebd9dff902cacd3ad11ddeb5e2b63c5b60d28], (new branch just for this issue https://github.com/franglais125/dash-to-dock/tree/minimize_action).

Note that I also included the `& Clutter.ModifierType.SHIFT_MASK` bit (I mentioned that in the Hotkeys PR, as "Super" is a modier and it causes problems). Anyway, I'm not sure which behavior you prefer!

I don't use "Minimize" as an option, but for some people it might feel like a regression.

micheleg:
Thanks for spotting the problem. I have restructured the code a bit.

1. I have kept the generic modifier because it should work with any modifiers, not only shift.
2. I have excluded the super case by runsetting it from the modifiers. Does this solve the hotkeys issue as well?

Can you check the ```development``` branch ?

micheleg:
Also the UI shoudl change, but I still don't know how (the help suggesting middle click should be shown for all actions, and only when the minimize option is set I think)

franglais125:
>  1)  I have kept the generic modifier because it should work with any modifiers, not only shift.

I agree with the idea, but from my testing it is a bit hard to predict what Clutter will interpret as a modifier in an event. For instance, having Caps Lock activated triggers the same problem, which seems really odd.
I added this patch [2693e5147284b69cbcbd91c8937037f134f09a07], based on your `development` branch.

>  2)  I have excluded the super case by runsetting it from the modifiers. Does this solve the hotkeys issue as well?
Unfortunately, it didn't solve the issue. In the same patch as mentioned above, you can see the changes.

It took me a while to make it work (https://developer.gnome.org/clutter/stable/clutter-Events.html#ClutterModifierType), essentially trial and error. In the end, I had to remove the `MOD4_MASK` for the Super+Num to work as a hotkey, and not be taken as a modifier (same patch as above).

Does this make sense to you? Can you reproduce the problem, i.e. removing `SUPER_MASK` doesn't work?
I cherry-picked the hotkey patch (so you can checkout my `development` branch to try this).

micheleg:
1.) I have realized that the only modifier we are using is indeed SHIFT, so let's just user your previously suggested patch
2.) Let's just whitelist the modifiers we like (CTRL and SHIFT) instead of excluding the ones we don't want

I have updated the development branch.

PR #351, opened 2016-07-17
changed popup menu max width to 250px

ItamarShDev:


ItamarShDev:
cancel
this fix makes the calendar top panel to be cut

Issue #20, opened 2012-08-14
Feature request: Add support for showing notification count for applications.


OlegOAndreev:
It would be useful to (optionally) show the number of notifications, which were emitted by each application. An example can be seen in the following screenshot: http://cloudfront.omgubuntu.co.uk/wp-content/uploads/2011/04/screenshotae_thumb.png

I'm willing to propose a patch if you like the general idea.

zeraien:
What happened with this? Any plans on implementing, is it worth it taking a crack at the code?

micheleg:
@zeraien This was never finalized and  merged and I don't think I am note personally motivated to work on it, as I don't feel the need to overhaul the default Gnome behaviour. Should someone want to try to improve this feature, i'm open to consider its integration, yet I can't promise much of my attention given the time I can dedicate to this project.

gochev:
so sad I would love such feature :( @OlegOAndreev  is gone :) and I guess this will never be rewritten and merged ? :( This is like the biggest thing I miss from my Mac when using Gnome. Is there any other similar extension or something that supports notification badge counts on apps ?

Feichtmeier:
Awesome extension!
Though I miss this really. Progress bars of downloads and unread mail count from geary. That would make it perfect. @gochev you could use plank, like I do, but it just feels renundant to install another dock when there could be this one

micheleg:
Also, download progressbar and unread messages count (vs unread notifications count) is not a freedesktop standard and is supported by ubuntu/libunity and derivative (elementary) only. This makes the notifications a bit less helpful. I personally don't want to target a specific platforms (which i don't run). I have no idea what will happen to unity support in applications now that the project has been deprecated.

Alexander-Wilms:
I file a bug report here: https://bugzilla.gnome.org/show_bug.cgi?id=781368

If Gnome could reuse the Unity LauncherAPI, that'd be great.

k3a:
Yes, this is probably the feature I miss the most, too. I used Plank on Linux which has some kind of "notification" dot but I can't use Plank when in Wayland. Other than that this extension works perfectly for me.

owais:
I wouldn't hold my breath on Gnome accepting any unity. Unity wouldn't exist if that was possible :)

Anyway, there has been some discussion in Gnome bug tracker around this and it looks like Gnome might be open to accept a NEW API to implement things like these. https://bugzilla.gnome.org/show_bug.cgi?id=701402#c13. 

[...]

PR #360, opened 2016-08-12
Hotkey implementation

franglais125:
Hi,

I created 2 new branches to rebase more easily. Here is the pull request for the hotkey feature.

I have 3 comments:
1. When using "minimize" as the click action, we can't use Super+Num to unminimize, due to the use of modifiers (Super). You specifically changed that in this commit [529304bc07b7c03b9f69b7ca37bb80ea9e27e01b], so I didn't change it back. Is there a specific reason? Can I add the `SHIFT_MASK` again?
2. I am not sure that the setting option fits well where it is (Setting.ui, in the Behavior Tab). I just added it somewhere for testing purposes. Let me know if we should put it somewhere else!
1. This line (https://github.com/micheleg/dash-to-dock/blob/master/appIcons.js#L278) is emitting errors due to th use of `event.get_click_count()`. This is logical since we don't have real clicks when pressing Super+Num. Anyway we can avoid that? Just so we don't spam the logs.

Obviously, credit to @anprogrammer! (Reference: https://github.com/micheleg/dash-to-dock/pull/87)

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -107,6 +107,7 @@ const MyAppIcon = new Lang.Class({
         }));
         this._optionalScrollCycleWindows();
 
+        this._numberOverlay();
     },
 
     _onDestroy: function() {
@@ -513,6 +514,61 @@ const MyAppIcon = new Lang.Class({
         Clutter.cairo_set_source_color(cr, bodyColor);
         cr.fill();
         cr.$dispose();
+    },
+
+    _numberOverlay: function() {
+        // Add label for a Hot-Key visual aid
+        this._numberOverlayLabel = new St.Label();
+        this._numberOverlayBin = new St.Bin({child: this._numberOverlayLabel });
+        this._numberOverlayStyle = 'background-color: rgba(0,0,0,0.8);'
+        this._numberOverlayOrder = -1;
+        this._numberOverlayBin.hide();
+
+        // iconBox is a Shell.GenericContainer from IconGrid.BaseIcon
+        this.iconBox = new Shell.GenericContainer();
+        this._iconContainer.add_child(this.iconBox);
+        // We have to add allocation of counterBin.
+        this.iconBox.connect('allocate', Lang.bind(this, this._allocateIconBox));
+
+        this.iconBox.add_actor(this._numberOverlayBin);
+
+        this._updateNumberOverlay();
+    },
+
+    _allocateIconBox: function(actor, box, flags) {
+        let childBox = new Clutter.ActorBox();
+        let naturalHeight = this.icon.iconSize;
+        // Partial size for the overlay:
+        naturalHeight *= 0.6;
+
+        childBox.x1 = box.x2 - naturalHeight/2;
+        childBox.x2 = box.x2 + naturalHeight/2;
+        childBox.y1 = box.y2 - naturalHeight/2;
+        childBox.y2 = box.y2 + naturalHeight/2;
+
+        this._numberOverlayBin.allocate(childBox, flags);
+    },
+
+    _updateNumberOverlay: function() {
+        let size = this.icon.iconSize;
+        // Set the font size to something smaller than the whole icon so it is
+        // still visible. The border radius is large to make the shape circular
+        let font_size = Math.round(0.4 * size);
+        this._numberOverlayBin.set_style(this._numberOverlayStyle +
+                                         'font-size: ' + font_size + 'px;' +
+                                         'border-radius: ' + size + 'px;');
+    },
+
+    setNumberOverlay: function(number) {
+        this._numberOverlayOrder = number;
+        this._numberOverlayBin.get_child().set_text(number.toString());
+    },
+
+    toggleNumberOverlay: function(activate) {
+        if (activate && this._numberOverlayOrder > -1)
+           this._numberOverlayBin.show();
+        else
+           this._numberOverlayBin.hide();
     }
 });
 
--- a/appIcons.js
+++ b/appIcons.js
@@ -524,15 +524,15 @@ const MyAppIcon = new Lang.Class({
         this._numberOverlayOrder = -1;
         this._numberOverlayBin.hide();
 
-        // iconBox is a Shell.GenericContainer from IconGrid.BaseIcon
-        this.iconBox = new Shell.GenericContainer();
-        this._iconContainer.add_child(this.iconBox);
+        // genericContainer is a Shell.GenericContainer from IconGrid.BaseIcon
+        let genericContainer = new Shell.GenericContainer();
+        this._iconContainer.add_child(genericContainer);
         // We have to add allocation of counterBin.
-        this.iconBox.connect('allocate', Lang.bind(this, this._allocateIconBox));
+        genericContainer.connect('allocate', Lang.bind(this, this._allocateIconBox));
 
-        this.iconBox.add_actor(this._numberOverlayBin);
+        genericContainer.add_actor(this._numberOverlayBin);
 
-        this._updateNumberOverlay();
+        this.updateNumberOverlay();
     },
 
     _allocateIconBox: function(actor, box, flags) {
@@ -549,7 +549,7 @@ const MyAppIcon = new Lang.Class({
         this._numberOverlayBin.allocate(childBox, flags);
     },
 
-    _updateNumberOverlay: function() {
+    updateNumberOverlay: function() {
         let size = this.icon.iconSize;
         // Set the font size to something smaller than the whole icon so it is
         // still visible. The border radius is large to make the shape circular
--- a/dash.js
+++ b/dash.js
@@ -860,6 +860,37 @@ const MyDash = new Lang.Class({
 
         // This is required for icon reordering when the scrollview is used.
         this._updateAppsIconGeometry();
+
+        // This will update the size, and the corresponding number for each icon
+        this._updateNumberOverlay();
+    },
+
+    _updateNumberOverlay: function() {
+        let appIcons = this._getAppIcons();
+        let counter = 1;
+        appIcons.forEach(function(icon) {
+            if (counter < 10){
+                icon.setNumberOverlay(counter);
+                counter++;
+            }
+            else if (counter == 10) {
+                icon.setNumberOverlay(0);
+                counter++;
+            }
+            else {
+                // No overlay after 10
+                icon.setNumberOverlay(-1);
+            }
+            icon._updateNumberOverlay();
+        });
+
+    },
+
+    toggleNumberOverlay: function(activate) {
+        let appIcons = this._getAppIcons();
+        appIcons.forEach(function(icon) {
+            icon.toggleNumberOverlay(activate);
+        });
     },
 
     setIconSize: function(max_size, doNotAnimate) {
--- a/dash.js
+++ b/dash.js
@@ -881,7 +881,7 @@ const MyDash = new Lang.Class({
                 // No overlay after 10
                 icon.setNumberOverlay(-1);
             }
-            icon._updateNumberOverlay();
+            icon.updateNumberOverlay();
         });
 
     },
--- a/docking.js
+++ b/docking.js
@@ -522,7 +522,7 @@ const DockedDash = new Lang.Class({
 
         // Remove keybindings
         this._disableHotKeys();
-        this._disableNumberOverlay();
+        this._disableExtraShortcut();
     },
 
     _bindSettingsChanges: function() {
--- a/docking.js
+++ b/docking.js
@@ -30,8 +30,6 @@ const Theming = Me.imports.theming;
 const MyDash = Me.imports.dash;
 
 const DOCK_DWELL_CHECK_INTERVAL = 100;
-const NUMBER_OVERLAY_INTERVAL = 2000;
-const SUPER_KEY_ID = 64;
 
 const State = {
     HIDDEN:  0,
@@ -1575,35 +1573,76 @@ const DockedDash = new Lang.Class({
     },
 
     _optionalNumberOverlay: function() {
-        if (this._settings.get_boolean('hot-keys'))
-            this._enableNumberOverlay();
+        this._shortcutIsSet = false;
+        // Enable extra shortcut if either 'overlay' or 'show-dock' are true
+        if (this._settings.get_boolean('hot-keys') &&
+           (this._settings.get_boolean('hotkeys-overlay') || this._settings.get_boolean('hotkeys-show-dock')))
+            this._enableExtraShortcut();
 
         this._signalsHandler.add([
             this._settings,
             'changed::hot-keys',
-            Lang.bind(this, function() {
-                    if (this._settings.get_boolean('hot-keys'))
-                        this._enableNumberOverlay();
-                    else
-                        this._disableNumberOverlay();
-            })
+            Lang.bind(this, this._checkHotkeysOptions)
+        ], [
+            this._settings,
+            'changed::hotkeys-overlay',
+            Lang.bind(this, this._checkHotkeysOptions)
+        ], [
+            this._settings,
+            'changed::hotkeys-show-dock',
+            Lang.bind(this, this._checkHotkeysOptions)
+        ], [
+            this._settings,
+            'changed::shortcut-text',
+            Lang.bind(this, this._setShortcut)
         ]);
     },
 
-    _enableNumberOverlay: function() {
-        Main.wm.addKeybinding('number-overlay-key', this._settings,
-                              Meta.KeyBindingFlags.NONE,
-                              Shell.ActionMode.NORMAL | Shell.ActionMode.OVERVIEW,
-                              Lang.bind(this, this._showOverlay));
-
+    _checkHotkeysOptions: function() {
+        if (this._settings.get_boolean('hot-keys') &&
+           (this._settings.get_boolean('hotkeys-overlay') || this._settings.get_boolean('hotkeys-show-dock')))
+            this._enableExtraShortcut();
+        else
+            this._disableExtraShortcut();
     },
 
-    _disableNumberOverlay: function() {
-        Main.wm.removeKeybinding('number-overlay-key');
+    _enableExtraShortcut: function() {
+        let shortcut_is_valid = this._setShortcut();
+
+        if (shortcut_is_valid && !this._shortcutIsSet) {
+            Main.wm.addKeybinding('shortcut', this._settings,
+                                  Meta.KeyBindingFlags.NONE,
+                                  Shell.ActionMode.NORMAL | Shell.ActionMode.OVERVIEW,
+                                  Lang.bind(this, this._showOverlay));
+            this._shortcutIsSet = true;
+        }
+    },
+
+    _setShortcut: function() {
+        let shortcut_text = this._settings.get_string('shortcut-text');
+        let [key, mods] = Gtk.accelerator_parse(shortcut_text);
+
+        if (Gtk.accelerator_valid(key, mods)) {
+            let shortcut = Gtk.accelerator_name(key, mods);
+            this._settings.set_strv('shortcut', [shortcut]);
+            return true;
+        }
+        else {
+            this._settings.set_strv('shortcut', []);
+            return false;
+        }
+    },
+
+    _disableExtraShortcut: function() {
+        if (this._shortcutIsSet) {
+            Main.wm.removeKeybinding('shortcut');
+            this._shortcutIsSet = false;
+        }
     },
 
     _showOverlay: function() {
-        this.dash.toggleNumberOverlay(true);
+        if (this._settings.get_boolean('hotkeys-overlay'))
+            this.dash.toggleNumberOverlay(true);
 
         // Restart the counting if the shortcut is pressed again
         if (this._numberOverlayTimeoutId) {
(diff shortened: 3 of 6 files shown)


micheleg:
I'm testing the branch now. 

1. should be solved if I merge b0f314b416f7de2ef11297c1054c733af25a60f1
2. Yes, I don't like it. Still thinking to a better placement
3. I'm not sure what the best way to fix this. It is illogical to call a method meant to be a callback for some click event, although it might be the easier solution  What is the error you are gettting? I can't see anything in the logs.

franglais125:
1. Comments in a separate discussion: https://github.com/micheleg/dash-to-dock/pull/349#issuecomment-241099144
2. I'll see if I can think of something
3. Here is the error message I get when using hotkeys (and asking for get_click_count()):

`Aug 19 14:06:17 debian org.gnome.Shell.desktop[6601]: (gnome-shell:6601): Clutter-CRITICAL **: clutter_event_get_click_count: assertion 'event->type == CLUTTER_BUTTON_PRESS || event->type == CLUTTER_BUTTON_RELEASE' failed`

Maybe we can do something like this? I tested it and it seemed to work.

`let click_count = 0;`
`if (Clutter.EventType.CLUTTER_BUTTON_PRESS)`
`click_count = event.get_click_count();`

micheleg:
for 3 you r solution looks good.

franglais125:
Sounds good.

Should I rebase onto your updated `development` branch?

micheleg:
yes

franglais125:
So, I updated the branch.
The problem with the Super key is no longer present.
The warning with the get_click_count() is now gone.

Only remaining issue:
 - where is it better to place the option in UI?

franglais125:
Hi Michele,

I updated the UI [f304f18f396aa699384e87261d2435094c48d274], here is a screenshot:
screenshot from 2016-08-23 11-56-24

Let me know what you think.

I just realized the setting is 'true' by default. Perhaps it would be better to have it off?

franglais125:
Hi @micheleg 
Even if you don't include this PR soon, perhaps you could cherry-pick the first 4 commits? They are in fact just fixes for regressions: click regressions and choosing the proper modifiers

I'll rebase later on if you pick them up.

f5356a798d840dd42ea01ab7ba69fa53e4f28546
b0f314b416f7de2ef11297c1054c733af25a60f1
51825c66bdd0fa8ee1a8a5780886e5795030639c
606302d93cd7f58389b520647b79a5a77848f020

franglais125:
@micheleg I've rebased this PR onto your master branch.

I've squashed a few commits to make it look simpler. There are only 2 commits now:
[218ee3ed4f434fe4be5d705e20277198db67d03b] is the implementation
[7b93ebb4aa687e304274027f704583eb3b9eb51f] is the fix for "click events".

franglais125:
@micheleg I just rebased this branch after the window preview feature you added to master.

Cheers

micheleg:
This old implementation of a notification count #20 could be of inspiration for adding number labels.

franglais125:
Thanks a lot for the suggestions and pointers!

I actually won't have that much time this weekend, but I will definitely try to improve this in the coming weeks.

franglais125:
Hi Michele,
I actually never intended to close this PR. Github closed it on its own when I rebased the branch. So I am reopening it.

- I have implemented a "number overlay" to make the feature more discoverable. See [ef4df2852eac0d4ab329bf1ea19cfa15fd34e267]. Your suggestion was quite helpful!
- As for the settings, I will try and see if I can work something out. As it is now, the Settings window doesn't fit on a 720p screen...

appIcons.js
```diff
@@ -443,6 +445,61 @@ const MyAppIcon = new Lang.Class({
         Clutter.cairo_set_source_color(cr, bodyColor);
         cr.fill();
         cr.$dispose();
+    },
+
+    _numberOverlay: function() {
+        // Add label for a Hot-Key visual aid
+        this._numberOverlayLabel = new St.Label();
+        this._numberOverlayBin = new St.Bin({child: this._numberOverlayLabel });
```
franglais125: It would be nice if we could add a style here, so that it is easy to edit from stylesheet.css.
I'm not sure how to do it though.
This is not strictly necessary of course, just an extra feature.

docking.js
```diff
@@ -1512,6 +1516,60 @@ const DockedDash = new Lang.Class({
...
+
+    _enableNumberOverlay: function() {
+        this.keymap = Gdk.Keymap.get_default();
+
+        this._signalsHandler.removeWithLabel('number-overlay');
+        this._signalsHandler.addWithLabel('number-overlay', [
+            this.keymap,
+            'state_changed',
+            Lang.bind(this, function() {
+                    let state = this.keymap.get_modifier_state();
```
franglais125: I actually don't like 100% this approach. I tried to use the signal `key-press-event` as it's done here (https://extensions.gnome.org/extension/10/windownavigator/), but I couldn't get it to work.
Instead, I took this idea from: https://extensions.gnome.org/extension/975/keyboard-modifiers-status/

franglais125:
I have tentatively added [59333b8123a2fe80105bd264e5a36f2f4396aede].
I don't use intellihide/autohide, so I'm not sure how useful it can be. What this patch does is to show the dock in case it is hidden (after a delay, same as the overlay) so that one can quickly see the apps' numbers when using the hotkeys.

franglais125:
screenshot from 2016-10-29 11 36 38
screenshot from 2016-10-29 11 38 06

micheleg:
Nice work! I'll have a look at it in the following days.

micheleg:
I'm testing this now. I'll try to review and hopefully merge soon.

micheleg:
One quick comment: I'm using command + mouse clicks a lot to move and resize windows, so it seems a bit strange to see the dock appearing. But I don't know how this could be fixed.

franglais125:
Thanks for the feedback.

Meanwhile, your suggestion prompted me to test the use of other modifiers... and it seems that the press/release logic needs to be improved a bit.
Right now, if you long-press <Super> to get the overlay numbers, but then decide to press <Sup><Ctrl><Number> (without releasing throughout the procedure), the overlay isn't hidden after release.

I'll try to find a workaround or both cases.

franglais125:
@micheleg I committed some changes which hopefully address the remaining issues.
 - [27a36d114e4df3415d43c96ac69d12aba8a89602] to avoid inconsistencies with other modifiers
 - [1e40f4a271c02a7b98f0dbfd6e892b8f010761fd] prevents the dock from showing when moving windows with `<Sup>+Click`
 - [f7b2f9c2e63a3b5e76a2b657e96a976f3e5e1cb4] adding comments and better variable naming

Let me know how it goes!

By the way, I tried to keep the commits separate for easier review, but merge-squashing them might be better.

nicman23:
not sure if i should open an issue, but on a wayland session, nautilus does not play nice.

Specifically, with ctrl super and (in my case 2), gnome does not launch a new window.

Also, after interacting with nautilus (not sure if nautlius only problem) when trying to focus a window with super + num, (specifically gnome-terminal), it launch a new window instead.

This is on arch stable (up to date), wayland and with master cloned (make, make zip-file)

franglais125:
@nicman23 I think what you are experiencing is a couple of Gnome bugs, but I could be wrong.
I happened to encounter the same problems, and tried to get this fixed upstream:
https://bugzilla.gnome.org/show_bug.cgi?id=764849
https://bugzilla.gnome.org/show_bug.cgi?id=766808

In any case, please open a new issue if those are not the problems you are experiencing.

franglais125:
@micheleg back on topic (number overlays): the issue above had me test Wayland, and it seems the overlays are not showing. I'll try to investigate that...

Would you prefer to merge as-is, or would you rather find a fix for this first?

micheleg:
I'd like to treat the wayland session as a first class citizen. However, my understanding is that the shortcuts works but the numbers are not shown, so this could be fixed later once the feature is finalized.

I'm still unsure about how it should behave: just as a launcher (always launch new instance) or as a click (following application status and modifier).

franglais125:
As for Wayland: exactly, the hotkeys work, it's just the overlay that is not being shown for the moment. Apparently `Gdk.keymap` behaves differently (changes for each application or something of the sort, see https://bugzilla.gnome.org/show_bug.cgi?id=696339), so I'll have to dig a little deeper to get it right. Also, I was looking forward to using Wayland by default pretty soon, so I'm certainly eager to fix this :) ! I've been double-checking with other features, totally forgot to test Wayland with the number overlays.

As for the behavior: at least from my usage, using the hotkeys as a click is what makes most sense (I think this is also the behavior for Unity and Win10), also one can use modifiers to get the other actions. I wouldn't mind setting up a second level of options for hotkeys (as for intellihide) asking users if they want:
 - to quickly show the dock when pressing `<Sup>` (already an option)
 - show number overlays
 - act as a launcher

In any case, I'm just trying to give some feedback. Let me know what you prefer... I usually tend to let the users decide (i.e. more options), but I don't want to turn the Settings.ui into a nightmare.

franglais125:
**Regarding Wayland and number-verlays:**
I've been trying to read up on that and try to see how to properly use `Gdk.Keymap` (so that it works uniformly for X and Wayland), but I don't have good news. This is what I gathered, and from my understanding:
 - _You simply can't track keyboard status on Wayland_, and this is by design (security, eavesdropping, etc.). From a Gnome dev: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/6YWHCFTB4VOWY4LA5GG562TFBGSBBC5G/
 - `<Sup>+[somekey]` has been fixed in recent versions of Mutter on Wayland (https://bugzilla.gnome.org/show_bug.cgi?id=748526).

**What are we left with?**
 - I'd hate to have two different codepaths for X and Wayland (and I think you too). We could leave it as it is, but Wayland would be broken (and it would sort of work on Xwayland windows, which would be terribly inconsistent).
 - One thing that I think can work for both X and Wayland is to have a shortcut (say `<Sup>+z`) that purposefully shows the dock with the number overlays for a moment? Btw, using `<Sup>` alone (as we do now) won't work, as Mutter doesn't propagate the event until a second key is pressed.
 - It's too bad because I really liked the implementation I had done... But looking forward, I wouldn't want to introduce this and then deprecate it soon. Too many users might see it as a regression when the time comes.
 - Another possibility is that in a few releases a codepath is opened for this to work, but I wouldn't bet on that, nor would I want to rely on something like that.
 - In any case, I'll keep my eyes open in case I find a workaround at some point.

**For now:**
What I would do to get a consistent behavior is to add an optional shortcut (`<Sup>+z` or something else) that shows the overlay and the dock for a moment (instead of holding `Sup`). However, before doing it I wanted to know what your thoughts were on this.

franglais125:
I went ahead and committed [f7ecc05c1d59ac88e945d9933f60c1be2e90d7a4]. It changes the behavior to what I suggested (using Sup+z), so that you can test it.

I don't think it is ideal (especially for AZERTY users, or if someone already has Sup+Z set to something else. We can change it to something else if you prefer (or ask the user for input), but for now it showcases the implementation.

franglais125:
@micheleg just a ping about this.

Hotkeys were merged already. Here is a summary (since it's been two months, it's hard to remember):
 - Added number overlays for easier discoverability of the feature
 - Quickly show the dock is hidden
 - Use `<Sup> + z` to show the overlay
 - Reorganized Settings.ui to prevent cluttering

micheleg:
Thank for the summary, very appreciated :) As you see I'm quite discontinued in dedicating time to the project.

franglais125:
No problem, it's hard for myself to keep track of my own commits, can't imagine for you!

As usual, I have no problem rebasing if you merge other features before. Just let me know if additional changes are required, whenever you get the time.

franglais125:
I just rebased on your master, and condensed the work into two commits:
- [6dfe47be8eb6582e69e9cef23d6c26a11a128d56] complete implementation
- [45b7aa1de4315de7c23d76048e1283f60c738566] Settings.ui rework to avoid cluttering

franglais125:
Sorry for spamming this PR. I had to rebase again due to the latest developments. I edited the previous message with the new commit SHAs.

micheleg:
No worry. thanks for taking care of updating this branch. I quickly tested it now. It took me a while to realize I had to press ctrl+z to show the numbers and dock. I'm still not totally convinced of this solution... at least, we have to make the shortcut customizable.

The option "quickly show the dock if hidden" does not make much sense:  if it is triggered by a shortcut and not any more by the super long press it should be the default.

franglais125:
> It took me a while to realize I had to press ctrl+z to show the numbers and dock. I'm still not totally convinced of this solution... at least, we have to make the shortcut customizable.

It's not the solution I like best, but from my investigations, there seems to be no way around this for Wayland (assuming we want to have X and Wayland consistent).
I will make it customizable for now, and see how it goes.

> The option "quickly show the dock if hidden" does not make much sense: if it is triggered by a shortcut and not any more by the super long press it should be the default.

Hmm, you are right. I tried to keep the functionality working, and forgot to rethink the logic from settings.

I will try to update this soon. (I am rebasing multi-monitor as well).

franglais125:
I reworked a bit the logic, making it configurable as you suggested. I'll add a screenshot so it's easier to see the changes.
I switched the default to `<Sup>q` as it's closer to the numbers, but it's configurable anyway.

Let me know what you think!

screenshot from 2017-01-28 16 13 37

appIcons.js
```diff
@@ -513,6 +514,61 @@ const MyAppIcon = new Lang.Class({
...
+        this._numberOverlayStyle = 'background-color: rgba(0,0,0,0.8);'
+        this._numberOverlayOrder = -1;
+        this._numberOverlayBin.hide();
+
+        // iconBox is a Shell.GenericContainer from IconGrid.BaseIcon
+        this.iconBox = new Shell.GenericContainer();
```
micheleg: iconBox as a variable name is a bit confusing I think. Also, why is iconBox "public"?

dash.js
```diff
@@ -860,6 +860,37 @@ const MyDash = new Lang.Class({
...
+            }
+            else if (counter == 10) {
+                icon.setNumberOverlay(0);
+                counter++;
+            }
+            else {
+                // No overlay after 10
+                icon.setNumberOverlay(-1);
+            }
+            icon._updateNumberOverlay();
```
micheleg: _updateNumberOverlay() should be "public" then.

franglais125:
@micheleg I hope I addressed the issues you pointed out properly. [1822d165cdc7d601235bbcd6b8c4b3ae3babeb39].

Thanks for the review!

micheleg:
I just noticed that the extensionfail to reload after screen lock with this error
 ```
JS LOG: Reloading completed : Dash to Dock : dash-to-dock@micxgx.gmail.com
JS LOG: Extension "dash-to-dock@micxgx.gmail.com" had error: TypeError: this._disableNumberOverlay is not a function
```

franglais125:
I fixed that in [1822d165cdc7d601235bbcd6b8c4b3ae3babeb39], I'll double check!

micheleg:
Yes, sorry, I missed your latest commits.

micheleg:
I've made some changes here: 55ced85946ae261398413e7ace13426fe43de787. 

I've moved the label to the icon corner, and also got rid of an unnecessary container. I was thinking the label size could also be fixed in size, but it look a bit strange with small and large icons. 

What do you think?

franglais125:
Thanks for the code improvements!

Honestly, I think it looks nice either way. Your style choice is actually more readable, while being less "symmetric".

I tested various icons sizes and the label is always readable.

micheleg:
Any comment about this 1eed2c1?

franglais125:
As I don't use the intellihide feature by default, I don't have a strong opinion on that. I usually use it only when testing new features to make sure I don't break it.

That being said, the logic behind that variable was that if a user is quickly cycling through windows/apps, perhaps a fullscreen app hides the dock while the user is still looking at the numbers for feedback. Then the dock hides anyway after the timeout.

Perhaps this doesn't make much sense now... It was more logical with the "hold the Sup key" behavior: if the user is _still_ holding the key, we don't hide the dock to preserve the feedback. As soon as the key was released, the dock would hide.

Again, up to you, I don't feel strongly about it.

micheleg:
Ok, now I understand the reason for that, let me think about it. 

Few additional changes:

d7880f116a0748: just experimenting with the behaviour, what about this? It partially solve the above-mentioned issue: if the user keep cycling windows by shortcuts the dock stay visible

7997c52870: hope this doesn't break anything.

I'm thinking we could unify the sho2 overlays and show dock options in a single setting, but I'll keep this for later after we merge the feature.

franglais125:
> d7880f1: just experimenting with the behaviour, what about this? It partially solve the above-mentioned issue: if the user keep cycling windows by shortcuts the dock stay visible

Seems ok, I just wonder if people will find it distracting if they want to use the hotkey, but not necessarily want to see the overlay.

> 7997c52: hope this doesn't break anything.

I can't remember why I did it like that, but I remember doing it for some obscure corner case. Too bad I didn't document it back then :/. Your changes are fine though as they are. If we find an issue, we can always add it back later on.

franglais125:
As for the "forceShow" option from before, I remember another small issue. Again, a corner case (and keep in mind I don't use intellihide often, so I don't know how important this is).

Issue:
1) Put the mouse on the dock area, while the dock is hidden (so, the mouse close to the edge of the screen).
2) Trigger the overlay -> so now the dock is showing
3) Move the mouse away quickly. The 'un-hover' action now makes the dock disappear before the timeout ends.

micheleg:
>Seems ok, I just wonder if people will find it distracting if they want to use the hotkey, but not necessarily want to see the overlay.

Yes it could be, but users can disable it in the settings (also disabling the showing on shortcut thought). I could merge it and see how many people complain, or wait (I'd like to improve the keyboard interaction and it could be rethought)

>As for the "forceShow" option from before, I remember another small issue...
This doesn't look like a problem to me.

We are converging towards a mergeable status. Last things to consider is the settings default and wording. In particular, which shortcut? And also, what's happen if we pick a shortcut already used by the user for some other action?

franglais125:
> Last things to consider is the settings default and wording. In particular, which shortcut?

Heh, I am not sure I understand the question here. Do you mean in the settings dialog?

> And also, what's happen if we pick a shortcut already used by the user for some other action?

This is something I was wondering as well. I didn't want to add even more stuff to the branch. I haven't investigated if it's possible to check for already existing keybindings, and then alert the user of this problem.
As for typical shortcuts (say `<Sup>a` for apps), my hope was the user would quickly realize that this keybinding is already taken, and then choose something else. But I agree that some kind of feedback would be nicer than nothing.

micheleg:
>Wording: 
* "Apps" tab -> "Launchers" ?
* "hotkey" -> "keyboard shortcut" ?

> which shortcut

Which default shortcut do we pick. And do we enable it by default?
For the shortcut conflict, it seems I get inconsistent behaviours, overwriting or being overwritten by the default one. But let's hope on the user understanding for the moment, and try to improve it later eventually.

franglais125:
Ah, that makes sense.

- "Launchers" sounds better to me
- "Hotkeys" or "Keyboard shortcuts", no preference really!

As for the default value of the shortcut, it's a usability habit I guess. The only reason I picked `<Sup>q` as default is that it was the closest letter to the "1" on the keyboard (i.e. where the hotkeys start), but I do realize this might differ a lot from user to user.

franglais125:
I have a question in general (1), and then a comment in particular (2) for one of the commits you proposed.

1) **Debugging:**
I have seen in different issues here on github a proposal to check for errors using `gnome-shell --replace` from a terminal, and then see the output.
I think that this is perhaps not sufficient. For some reason, gnome-shell is logging quite a few things to `/var/log/kern.log` that are not visible from the command prompt. This is what I see on Debian Stretch, on Debian Jessie I simply looked at `/var/log/user.log`.
Which brings me to the second point.

2) Commit [55ced85946ae261398413e7ace13426fe43de787]
This commit make gnome-shell spit a lot of warnings on the log (but not on the terminal!). Here is a gist of it https://gist.github.com/franglais125/f1580bde98450251b02c429cb9736abd.
This is related to the removal of the icon container, I believe.

micheleg:
I found that error moving in different place over time, and it also varies from distribution to distribution so it is always difficult to recommend users a way to access the error log. I used to rely on the ```gnome-shell restart``` trick on debian Jessie and before. I've never know about ```/var/log/user.log```.  

Instead, in Stretch, I'm now relying on ```journalctl  -f /usr/bin/gnome-shell```,  which also seems to show your reported error. Thanks for spotting it. It's not clear to me where it originates.

micheleg:
72df8745739c4 Seems to fix the problem for me. Apparently calling ```get_preferred_width()``` on an actor not on the stage results in the attempt to access  the theme node.

franglais125:
Thanks, that seems to have fixed it for me too!

micheleg:
Good. I'm going to clean up a bit the history to prepare for merge.

micheleg:
Merged into master: [578481c1fb44..0aa48a1a3]. I'm still not totally satisfied by the settings though. In particular, I don't want to have a shortcut such as <super>q enabled by default, as it would confuse people who have it already mapped.

franglais125:
Thanks for the merge!

If you have a better idea for the settings, let me know and I'll try to help.
As for the default behavior, if both booleans `hotkeys-overlay` and `hotkeys-show-dock` are false by default, then the shortcut is not set (since it doesn't do anything).

We can perhaps turn them both off by default?

franglais125:
I'm closing this, as there is nothing new going now for now. ALl relevant commits were reworked and merged

aryasenna:
@franglais125  Hello, I just installed and love dash to dock, but I noticed the Super+<num> interfere with my existing shortcut to switch workspace? Is there any option to turn this behavior off, or am I stuck on maintaining my own fork..? :(

micheleg:
There's an option in the settings to disable the shortcuts (in the behavior tab). Doesn't that work?

aryasenna:
@micheleg Ah yes, it works.. It's 3.30 AM and I should sleep.  uh, Much thanks!

markstos:
I preferred the Unity behavior here, but the "Super Q" shortcut works well enough. Thanks for that. If only there was a way to discover that. 

(As a new Ubuntu 17.10 user, I'm not even sure how to open the related dock preferences.)

Issue #354, opened 2016-07-22
Dock not refreshing


tenten8401:
Running Ubuntu Gnome 16.04 and if the screen falls asleep, and I wake it back up and unlock, the dock no longer appears to refresh running apps until restarting

micheleg:
Sorry for the late reply. What version of Gnome Shell? Does it also occurs if you purposely lock the screen?

PR #363, opened 2016-08-19
Workspace Isolation improvement

franglais125:
Hi Michele,

This is a tiny PR.

1. I added a signal `'switch-workspace'`, since in Wayland sessions `'restacked'` is not emitted when changing workspace.
2. The update to the `RestackAction()` function is simply to improve efficiency.

I've used these changes enough (as usual, Debian stable and testing, so 3.14 and 3.20) that I'm confident they don't introduce problems.

Similar changes were introduced in the workspace-isolated-dash extension.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -1410,22 +1410,18 @@ const DockedDash = new Lang.Class({
             this._signalsHandler.addWithLabel(label, [
                 global.screen,
                 'restacked',
-                Lang.bind(this, RestackAction)
+                Lang.bind(this.dash, this.dash._queueRedisplay)
             ]);
             this._signalsHandler.addWithLabel(label, [
                 global.window_manager,
                 'switch-workspace',
-                Lang.bind(this, RestackAction)
+                Lang.bind(this.dash, this.dash._queueRedisplay)
             ]);
         }
 
         function disable() {
             this._injectionsHandler.removeWithLabel(label);
             this._signalsHandler.removeWithLabel(label);
-        }
-
-        function RestackAction() {
-            this.dash._queueRedisplay();
         }
 
         function IsolatedOverview() {
--- a/docking.js
+++ b/docking.js
@@ -1412,6 +1412,11 @@ const DockedDash = new Lang.Class({
                 'restacked',
                 Lang.bind(this, RestackAction)
             ]);
+            this._signalsHandler.addWithLabel(label, [
+                global.window_manager,
+                'switch-workspace',
+                Lang.bind(this, RestackAction)
+            ]);
         }
 
         function disable() {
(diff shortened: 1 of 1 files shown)


micheleg:
Do you think there is any drawback then at removing the restack event?

franglais125:
Not really, I wanted to keep the difference small, but we might as well remove `RestackAction()` and call the `this.dash._queueRedisplay` directly.

micheleg:
I'd prefer to keep X and Wayland the same. If we know restacked does not work on wayland, we should get rid of it.

franglais125:
100% agreed. For some reason I thought both were necessary, but after a quick test it seems `'restacked'` can be removed.

Let me test this a bit more, and get back to you.

I'll also remove `RestackAction()` if everything works out.

franglais125:
Hi,

**1)** Regarding the signal `'restacked'`, if this signal is not included, too often the dock fails to update the running icons. I don't know why workspace isolation would trigger this more than the non-isolated case. It feels similar to https://github.com/micheleg/dash-to-dock/issues/354.
With X, when switching to an empty (no windows) workspace, it seems that a `restacked` signal is emitted anyway, since the window is changing (to no window). Under Wayland, this doesn't seem to happen. This is why I added the `'switch-workspace'` signal, but I didn't intend to replace the other signal. Hence:
X: needs `'restacked'`
Wayland:  needs `'restacked'`and `'switch-workspace'`

**2)** Regarding removing the RestackAction function. When I call directly `_queueRedisplay`, I get an error and it fails. Do you know if there is anything wrong in the code/logic?
It essentially tells me that workId is invalid (https://github.com/micheleg/dash-to-dock/blob/master/dash.js#L426).
`this._signalsHandler.addWithLabel(label, [`
`                global.screen,`
`                'restacked',`
`                Lang.bind(this, this.dash._queueRedisplay)`
`            ]);`
**Error message**:
`Aug 20 10:31:40 debian gnome-session[9243]: (gnome-shell:9381): Gjs-WARNING **: JS ERROR: Invalid work id NaN: Error: Invalid work id NaN`
`Aug 20 10:31:40 debian gnome-session[9243]: queueDeferredWork@resource:///org/gnome/shell/ui/main.js:602`
`Aug 20 10:31:40 debian gnome-session[9243]: MyDash<._queueRedisplay@/home/franglais125/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/dash.js:421`


Let me know if you have suggestions or ideas. With the issues I found, I couldn't change anything to the patch above.

franglais125:
Regarding **2)**, I was simply missing a `.dash` :
`Lang.bind(this.dash, this.dash._queueRedisplay)`

So I removed the unnecessary RestackAction() function.

Cheers

micheleg:
Merged [c25ba2746], Thank you.

PR #377, opened 2016-09-17
Support Gnome shell 3.21

pade:
Hi

I test your great extension with new gnome shell 3.21. Works fine, I haven't seen any problem.
I suggest to update it for this new gnome shell release.

Thanks for this very useful extension !


zapashcanon:
3.21 is a development version, it shouldn't be added to the metadata.json, moreover, 3.21.92 disable version checks by default, so it won't be necessary anymore.

micheleg:
@pade  Thank for your suggestion. But yes, I never add development releases to the metadata. Now, as @zapashcanon reminded, it is totally unnecessary.

PR #409, opened 2016-11-27
Added support to control dash borders

gayanper:
A new setting is introduce to configure wether the dash borders must be rounded or straight

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -2011,12 +2011,13 @@
                         <property name="can_focus">False</property>
                         <property name="spacing">32</property>
                         <child>
-                          <object class="GtkLabel" id="enable_rounded_borders_label">
+                          <object class="GtkLabel" id="force_non_rounded_corner_label">
                             <property name="visible">True</property>
                             <property name="can_focus">False</property>
                             <property name="valign">center</property>
                             <property name="xpad">12</property>
-                            <property name="label" translatable="yes">Enable rounded borders</property>
+                            <property name="label" translatable="yes">Force non rounded coner
+</property>
                             <property name="xalign">0</property>
                           </object>
                           <packing>
@@ -2026,7 +2027,7 @@
                           </packing>
                         </child>
                         <child>
-                          <object class="GtkSwitch" id="enable_rounded_borders_switch">
+                          <object class="GtkSwitch" id="force_non_rounded_corner_switch">
                             <property name="visible">True</property>
                             <property name="can_focus">True</property>
                             <property name="valign">center</property>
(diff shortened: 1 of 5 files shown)


micheleg:
Hi,

thank you for your contribution. There are few problems with the current implementation

1. It overwrites the border rounded corner with the "Use built in theme" option. Like the other sub-options of the theming, it should be ignored in this case.

2. The button size in the settings panel is wrong
screenshot from 2016-11-29 22-19-18

If these are fixed I'm happy to consider the pull request for inclusion in master.

gayanper:
Sure i will look into those issues.

gayanper:
@micheleg Can you elaborate more on the first option a bit more. I'm not clear the expectation you have there.

gayanper:
I have fixed the switch height.

micheleg:
Sorry for the late reply. What I was pointing out with problem 1., is that if you select "use built-in theme" all the option below are ignored, and so should your one. Visually, they are grayed out for this reason.

gayanper:
@micheleg  Please review the latest commit to fix **Problem 1**

micheleg:
Thanks for your contribution. I had a look at your changes. The way it behaves now is quite inconsistent:

* shrink and built-in theme force no round corner. This could be fine but I'd like the built-in option to have the rounded corners.
* enable round corner force round corner unless shrink is selected.
* disabling round corner keep the theme defaults, which could be with round corner

We need a more clever way of arranging the options.  So, first of all, what is the problem we are trying to solve and what are we trying to achieve?

1. If the idea is customizing the corners we need three statuses:  a.) forced non-rounded corner,  b.)  default,  c.) forced rounded corner .
This could be implemented with a "customize corner" option plus the choice of the border radius

2. If we are trying to fix themes incompatibilities, that option should be renamed either "force round corner" of "force non-round corner" and implemented accordingly.

gayanper:
I would like to go with the second option for now and they may be implement the first option later as a new feature. Because the first option requires more modifications and code flow changes than the current solution. from end user point of view if we solve his/her problem first and then implement our best solution around it, it will be a cherry on the cake for end users. What do you think ?

micheleg:
Thanks. I've done few changes here: https://github.com/micheleg/dash-to-dock/tree/gayanper-rnd-borders

In particular, I think 8fcf13a48dbd2f71af44c00d1391bd898e16de1a fix the behaviour both with and without the "shrink" option enabled. The others are "cosmetic" changes.

Please have a look at it and see if it's still working as expected. If so I'll merge the feature into master.

gayanper:
@micheleg I tested your changes and it working as expected and looks good.

micheleg:
Merged into master [71daa570a74]. Thank you!

Issue #183, opened 2015-06-11
Feature request: Cycle trough windows


vincent-t:
Could you please add an option to cycle through open windows of an application via mouse wheel.
best regards

micheleg:
The scroll is used to scroll the dock itself at the moment, so I need to think of a way to make things coexist.

gabrielmoreira:
I have used Plank, Docky, Cairo-Dock and Dash-to-Dock.

But only Dash-to-Dock doesn't cycle open windows using mouse wheel.

micheleg:
@gabrielmoreira that's doesn't mean we can't do something different :) That said, the problem is the coexistence with the other scroll behaviour, namely workspace switching and more importantly dock scrolling with fixed icon option. I gess both Plank and Cairo-Dock always shrink in size to allocate more icons instead?

Deus-Rex:
Is it possible to be able to change the dock scrolling to either a click or hover to drag? I'm sure that's how Unity has it, which lets that be able to implement scroll to cycle through windows.

I'd very much appreciate being able to cycle by scrolling as well. It's the only thing this dock is missing.

micheleg:
Cycle by scrolling was implemented by #428.

PR #428, opened 2017-01-08
Add cycle through windows feature

gabrielmoreira:
Issue #183 

Added a new setting "Scroll action", with 3 options:
- Do nothing
- Cycle through windows
- Switch workspace

[screenshot](https://s29.postimg.org/pp1rhlu7r/Captura_de_tela_de_2017_01_08_00_43_54.png)

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -1,11 +1,250 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.20.0 -->
 <interface>
   <requires lib="gtk+" version="3.12"/>
   <object class="GtkAdjustment" id="animation_time_adjustment">
     <property name="upper">1</property>
     <property name="step_increment">0.050000000000000003</property>
     <property name="page_increment">0.25</property>
+  </object>
+  <object class="GtkBox" id="box_middle_click_options">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="margin_left">12</property>
+    <property name="margin_right">12</property>
+    <property name="margin_top">12</property>
+    <property name="margin_bottom">12</property>
+    <property name="orientation">vertical</property>
+    <child>
+      <object class="GtkFrame" id="frame_middle_click_options">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="label_xalign">0</property>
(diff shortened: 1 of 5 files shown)


micheleg:
Sorry for the late reply and thanks for your contribution. It seems to be working very well, and it totally makes sense to have this option. The settings look clear too.

I'll try to merge it as soon as possible, just let me review your code. I quickly skimmed through your changes. I have a concern about the changes in the setttings.ui file. By changing the file, it got updated to latest gtk version which might be a problem. At the moment the master version is meant for Gnome Shell 3.18-3.22. I'd ask you to isolate the changes related to the additional settings ui. I'll have to update the ui file sooner or later, but I prefer not to risk to break back-compatibility now.

gabrielmoreira:
Hi @micheleg ! I returned the Settings.ui to the previous version, and applied only the difference.

I have no idea if this feature is well-programmed. This was my first contact with gnome-extensions, and for me it was not easy. I could not find good documentation, complete examples, and it is not quick to understand the various concepts. The biggest difficulty was to test. My gnome-shell froze several times when there was a bug. But I was determined to replace the plank, so I struggled a little bit more :)

Since I sent the patch, I have used it on my home computer and my work (Manjaro - archlinux), replacing Plank, and I had no problem.

In the future, I think we can put a feature similar to the shift + click action, we could have the shift + scroll action for a second action.

Anyway, grateful for the attention.

micheleg:
There's no real documentation for extension development, and it doesn't help that it's often a fight to patch the upstream code, which can  makes some parts of the code awful.

Your implementation follows very much my code and look clean enough for merging. I'd like to rethink how this "optional feature" are organized but that will be for the future.

I've only changed the deadtime to 250 ms instead of 150 ms for consistency with the workspace scrolling (any thoughts on which is the optimal value?). The updated branch is here: https://github.com/micheleg/dash-to-dock/tree/scroll_cycle_windows.

I should merge the feature soon.

gabrielmoreira:
I do not know what timeout would be the best setting for deadtime, but if it's easy and few lines of code, I suggest making this value configurable by gsettings (without ui, just gsettings). 

Thanks again.

micheleg:
Merged in master [0040c2af8a76], Thank you!

micheleg:
I only now remembered about the incompatibility with the "fixed icon size" option. At the moment the feature conflict with the ability to scroll to show the additional launchers.

PR #438, opened 2017-01-23
Minor optimization

zapashcanon:
Just a minor optimization.

Before the loop:  avoid one multiplication and two subtractions but add two additions.

`h - 2 * n * r - (n - 1) * s`
= `h - n * (2 * r + s) + s`

In the loop: avoid one multiplication.

`(2 * i + 1) * r + i * s`
= `i * (2 * r + s) + r`

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -474,34 +474,34 @@ const MyAppIcon = new Lang.Class({
 
         switch (side) {
         case St.Side.TOP:
-            cr.translate((width - (2*n)*radius - (n-1)*spacing)/2, padding);
+            cr.translate((width - n * (2 * radius + spacing) + spacing)/2, padding);
             for (let i = 0; i < n; i++) {
                 cr.newSubPath();
-                cr.arc((2*i+1)*radius + i*spacing, radius + borderWidth/2, radius, 0, 2*Math.PI);
+                cr.arc(i * (2 * radius + spacing) + radius, radius + borderWidth/2, radius, 0, 2*Math.PI);
             }
             break;
 
         case St.Side.BOTTOM:
-            cr.translate((width - (2*n)*radius - (n-1)*spacing)/2, height - padding);
+            cr.translate((width - n * (2 * radius + spacing) + spacing)/2, height - padding);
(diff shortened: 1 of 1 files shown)


micheleg:
Thanks for the suggestion. Did you actually profile the code? I understand that in principle it could be faster, but is it actually the case?

zapashcanon:
No I didn't, I'm not very into javascript and have no idea how to do this.

But as it's an interpreted language, I think it's more likely to be the case.

micheleg:
I'm not an expert by any means, I'm not sure whoat the jit compiler does. I quickly try to run  throgh gjs these snippets:

```
let h = 12.0
let n=3
let r = 2.0
let s = 2
let result;
for (let i=0; i<10000000; i++) {
	result = h - n * (2 * r + s) + s;
} 
```
and this:
```
for (let i=0; i<10000000; i++) {
	result = h - 2 * n * r - (n - 1) * s;
} 
```
and I can't see any noticeable difference. Not a very scientific measurement, I agree.

zapashcanon:
I think it's because it's the wrong expression. ;)

`result` doesn't depends on `i` here. So, `result` always has the same value. I guess the compiler is smart enough to notice it and computes it before.

But even with the right expression, it wouldn't be easy to test, as giving predefined values to `r` and `s` would allow the compiler to compute part of the result before execution.

micheleg:
Yes, I actually included the ```i``` in the expression in my test but couldn't see any difference either. I'm still to my point that if you can't profile it (or are sure it's faster) it's not an optimization.

zapashcanon:
Yep' I understand. Will try to have a look at how to check if there's a difference, could be interesting. But probably won't find the time soon.

zapashcanon:
Just tried [this](https://godbolt.org/g/XDDXEW). Looks like gcc isn't smart enough to simplify it with `-O2` (the default). In this case, it's clear that there's an optimization. I also tried with `-O6`, assembler code produced is better too. :)

So, I'm not sure, but I think that if gcc isn't smart enough to simplify it, then jit - or whatever is used in gnome - isn't.

PR #453, opened 2017-03-19
Click to show dock when window maximized

gogi983:
Hi, 
Is there any possibility to add an option like in "Workspaces-to-Dock" extension, which requires, if enabled, to click over the area where the Dock should appear if a window is maximized?

I imagine that it shouldn't be so hard to implement, but I have no skills in programming to do it or suggest some code to do it.

Regards.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 UUID = dash-to-dock@micxgx.gmail.com
 BASE_MODULES = extension.js stylesheet.css metadata.json COPYING README.md
-EXTRA_MODULES = convenience.js dash.js docking.js appIcons.js intellihide.js prefs.js theming.js Settings.ui
+EXTRA_MODULES = convenience.js dash.js docking.js appIcons.js windowPreview.js intellihide.js prefs.js theming.js Settings.ui
 EXTRA_MEDIA = logo.svg
 TOLOCALIZE =  prefs.js
 MSGSRC = $(wildcard po/*.po)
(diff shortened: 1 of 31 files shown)


micheleg:
@gogi983 I guess the pull request was not intended? 

Would you mind pointing out at the option in workspace-to-dock to enable your desired behaviour? I quite don't understand what you are proposing.

gogi983:
Oh sorry no it wasn't intended, my bad... Instead of opening a new thread in "issues", even it is more a suggestion/request than an issue, I probably clicked on "pull request", sorry...

Well there is an option in "Workspaces-to-dock" which, when checked, requires mouse click (in addition to hovering) to show dock when focused window is maximized.
Can you implement same stuff in your extension, I guess it's possible and would be really helpful if you could do this?

micheleg:
Thanks, now I understand. I haven't looked at the implementation. I'm a bit concerned about having this as a default, because it might interfere with the user trying to click on some button of the application. It could be an option, but what is exactly the problem this feature solves?

gogi983:
I didn't mean having it as a default but as a possible option one could enable by checking it. 
About your concern, I don't know exactly how they implemented it but there must be some tweakable stuff to do in the code part for this, as in "Workspaces-to-dock" extension one have to move mouse pointer really on the edge screen and click to make it appear so the area for click should be very very thin, about  1 or 2px... 

It solves nothing but I really appreciate the behaviour when a focused window is maximized as it prevents the extension to appear when I "accidentally" move the mouse too fast, and so only appears when I voluntarily click on the edge of the part of the screen where it is configured to appear.

micheleg:
I see. Are you aware that you can disable intellihide with fullscreen applications? Would you mind opening an issue, so that I can close this pull request?

gogi983:
I opened an issue so you can close the pull.

micheleg:
Thanks.

PR #456, opened 2017-03-25
Add GNOME Shell 3.24 support.

dz0ny:


micheleg:
Merged [97f6a0bb95], thank you.  I'm a bit late with the explicit support of the new GNOME release. I will make a new release as soon as possible.

dz0ny:
:)

PR #457, opened 2017-03-27
Multi monitor

franglais125:
So, here we go... Commit hashes might change over time, as I'm often rebasing on current master branch.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -21,6 +21,7 @@ const PopupMenu = imports.ui.popupMenu;
 const Tweener = imports.ui.tweener;
 const Util = imports.misc.util;
 const Workspace = imports.ui.workspace;
+const LayoutManager = imports.ui.main.layoutManager;
 
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 const Convenience = Me.imports.convenience;
@@ -67,10 +68,9 @@ const MyAppIcon = new Lang.Class({
     Extends: AppDisplay.AppIcon,
 
     // settings are required inside.
-    _init: function(settings, app, monitorIndex, iconParams) {
+    _init: function(settings, app, iconParams) {
         // a prefix is required to avoid conflicting with the parent class variable
         this._dtdSettings = settings;
-        this._monitorIndex = monitorIndex;
         this._signalsHandler = new Convenience.GlobalSignalsHandler();
         this._nWindows = 0;
 
@@ -227,9 +227,10 @@ const MyAppIcon = new Lang.Class({
         [rect.x, rect.y] = this.actor.get_transformed_position();
         [rect.width, rect.height] = this.actor.get_transformed_size();
 
+        let monitorIndex = LayoutManager.findIndexForActor(this.actor);
+
         let windows = this.app.get_windows();
         if (this._dtdSettings.get_boolean('multi-monitor')){
-            let monitorIndex = this._monitorIndex;
             windows = windows.filter(function(w) {
                 return w.get_monitor() == monitorIndex;
             });
--- a/appIcons.js
+++ b/appIcons.js
@@ -21,7 +21,6 @@ const PopupMenu = imports.ui.popupMenu;
 const Tweener = imports.ui.tweener;
 const Util = imports.misc.util;
 const Workspace = imports.ui.workspace;
-const LayoutManager = imports.ui.main.layoutManager;
 
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 const Convenience = Me.imports.convenience;
@@ -68,9 +67,10 @@ const MyAppIcon = new Lang.Class({
     Extends: AppDisplay.AppIcon,
 
     // settings are required inside.
-    _init: function(settings, app, iconParams) {
+    _init: function(settings, app, monitorIndex, iconParams) {
         // a prefix is required to avoid conflicting with the parent class variable
         this._dtdSettings = settings;
+        this._monitorIndex = monitorIndex;
         this._signalsHandler = new Convenience.GlobalSignalsHandler();
         this._nWindows = 0;
 
@@ -227,10 +227,9 @@ const MyAppIcon = new Lang.Class({
         [rect.x, rect.y] = this.actor.get_transformed_position();
         [rect.width, rect.height] = this.actor.get_transformed_size();
 
-        let monitorIndex = LayoutManager.findIndexForActor(this.actor);
-
         let windows = this.app.get_windows();
         if (this._dtdSettings.get_boolean('multi-monitor')){
+            let monitorIndex = this._monitorIndex;
             windows = windows.filter(function(w) {
                 return w.get_monitor() == monitorIndex;
             });
--- a/appIcons.js
+++ b/appIcons.js
@@ -70,6 +70,7 @@ const MyAppIcon = new Lang.Class({
     _init: function(settings, app, iconParams) {
         // a prefix is required to avoid conflicting with the parent class variable
         this._dtdSettings = settings;
+        this._signalsHandler = new Convenience.GlobalSignalsHandler();
         this._nWindows = 0;
 
         this.parent(app, iconParams);
@@ -97,7 +98,11 @@ const MyAppIcon = new Lang.Class({
                    'custom-theme-running-dots-border-width'];
 
         keys.forEach(function(key) {
-            this._dtdSettings.connect('changed::' + key, Lang.bind(this, this._toggleDots));
+            this._signalsHandler.add([
+                this._dtdSettings,
+                'changed::' + key,
+                Lang.bind(this, this._toggleDots)
+            ]);
         }, this);
 
         this._toggleDots();
@@ -123,6 +128,8 @@ const MyAppIcon = new Lang.Class({
         // stateChangedId is already handled by parent)
         if (this._focusAppId > 0)
             tracker.disconnect(this._focusAppId);
+
+        this._signalsHandler.destroy();
 
         if (this._scrollEventHandler)
             this.actor.disconnect(this._scrollEventHandler);
--- a/appIcons.js
+++ b/appIcons.js
@@ -67,9 +67,10 @@ const MyAppIcon = new Lang.Class({
     Extends: AppDisplay.AppIcon,
 
     // settings are required inside.
-    _init: function(settings, app, iconParams) {
+    _init: function(settings, app, monitorIndex, iconParams) {
         // a prefix is required to avoid conflicting with the parent class variable
         this._dtdSettings = settings;
+        this._monitorIndex = monitorIndex;
         this._signalsHandler = new Convenience.GlobalSignalsHandler();
         this._nWindows = 0;
 
@@ -227,6 +228,12 @@ const MyAppIcon = new Lang.Class({
         [rect.width, rect.height] = this.actor.get_transformed_size();
 
         let windows = this.app.get_windows();
+        if (this._dtdSettings.get_boolean('multi-monitor')){
+            let monitorIndex = this._monitorIndex;
+            windows = windows.filter(function(w) {
+                return w.get_monitor() == monitorIndex;
+            });
+        }
         windows.forEach(function(w) {
             w.set_icon_geometry(rect);
         });
--- a/docking.js
+++ b/docking.js
@@ -1673,25 +1673,25 @@ const DockManager = new Lang.Class({
 
     _createDocks: function() {
 
-        this._preferredMonitor = this._settings.get_int('preferred-monitor');
+        this._preferredMonitorIndex = this._settings.get_int('preferred-monitor');
         // In case of multi-monitor, we consider the dock on the primary monitor to be the preferred (main) one
         // regardless of the settings
         // The dock goes on the primary monitor also if the settings are incosistent (e.g. desired monitor not connected).
         if (this._settings.get_boolean('multi-monitor') ||
-            this._preferredMonitor < 0 || this._preferredMonitor > Main.layoutManager.monitors.length - 1
+            this._preferredMonitorIndex < 0 || this._preferredMonitorIndex > Main.layoutManager.monitors.length - 1
             ) {
-            this._preferredMonitor = Main.layoutManager.primaryIndex;
+            this._preferredMonitorIndex = Main.layoutManager.primaryIndex;
         } else {
             // Gdk and shell monitors numbering differ at least under wayland:
             // While the primary monitor appears to be always index 0 in Gdk,
             // the shell can assign a different number (Main.layoutManager.primaryMonitor)
             // This ensure the indexing in the settings (Gdk) and in the shell are matched,
             // i.e. that we start counting from the primaryMonitorIndex
-            this._preferredMonitor = (Main.layoutManager.primaryIndex + this._preferredMonitor) % Main.layoutManager.monitors.length ;
+            this._preferredMonitorIndex = (Main.layoutManager.primaryIndex + this._preferredMonitorIndex) % Main.layoutManager.monitors.length ;
         }
 
         // First we create the main Dock, to get the extra features to bind to this one
-        let dock = new DockedDash(this._settings, this._preferredMonitor);
+        let dock = new DockedDash(this._settings, this._preferredMonitorIndex);
         this._mainShowAppsButton = dock.dash.showAppsButton;
         this._allDocks.push(dock);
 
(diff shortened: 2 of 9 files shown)


franglais125:
Here is a screenshot of the current look (24 and 23 inch monitors):
screenshot from 2017-03-27 16-57-38

SharpOB:
Very good work! I hope it gets merged.

appIcons.js
```diff
@@ -67,9 +67,11 @@ const MyAppIcon = new Lang.Class({
     Extends: AppDisplay.AppIcon,
 
     // settings are required inside.
-    _init: function(settings, app, iconParams, onActivateOverride) {
+    _init: function(settings, app, monitorNumber, iconParams, onActivateOverride) {
```
micheleg: little thing: maybe it should be named monitorIndex instead of monitorNumber, for consistency with the shell naming.

bigger thing. We are passing the monitor  number through all the classes. Could we retrieve the monitor internally with something like ```findMonitorForActor``` (https://git.gnome.org/browse/gnome-shell/tree/js/ui/layout.js#n891). I don't think it is a good idea (not sure if it it completely reliable for instance), but I wanted to point out the option.

appIcons.js
```diff
@@ -97,7 +99,11 @@ const MyAppIcon = new Lang.Class({
                    'custom-theme-running-dots-border-width'];
 
         keys.forEach(function(key) {
-            this._dtdSettings.connect('changed::' + key, Lang.bind(this, this._toggleDots));
+            this._signalsHandler.add([
+                this._dtdSettings,
+                'changed::' + key,
+                Lang.bind(this, this._toggleDots)
+            ]);
```
micheleg: We are keeping a single settings object, therefore we need the disconnect machinery. 

Just few considerations:
*  For simplicity, what about having a settings object per dock, and just destroy the settings object on dock remove?
* Even more, what about we completely reload the extension when monitor changes, or the multimonitor option is activated?

Would that make the extension code simpler and less bug prone?

intellihide.js
```diff
@@ -198,7 +199,7 @@ const Intellihide = new Lang.Class({
              * select a window in the secondary monitor.
              */
 
-            let monitorIndex = this._settings.get_int('preferred-monitor');
+            let monitorIndex = this._monitorIndex;
```
micheleg: This is not used, right? The tracked actor already bring information about the monitor. I guess you included for consistency, maybe we can skip it.

prefs.js
```diff
@@ -303,6 +303,14 @@ const Settings = new Lang.Class({
                             this._builder.get_object('windows_preview_button'),
                             'active',
                             Gio.SettingsBindFlags.DEFAULT);
+        this._settings.bind('hot-keys',
+                            this._builder.get_object('hot_keys_switch'),
+                            'active',
+                            Gio.SettingsBindFlags.DEFAULT);
```
micheleg: This is an unwanted change, some fix to the current settings or what?

docking.js
```diff
@@ -44,6 +44,14 @@ const scrollAction = {
...
 };
 
+/* Array of all the docks created */
+let allDocks = [];
+
+/* status variable: true when the overview is shown through the dash
+ * applications button.
+ */
+let forcedOverview = false;
+
```
micheleg: If we have a dockManager object maybe these should be inside the dockManager (or some other class, not at the single dock level)

docking.js
```diff
@@ -1157,7 +1152,8 @@ const DockedDash = new Lang.Class({
      */
     _adjustPanelCorners: function() {
         let extendHeight = this._settings.get_boolean('extend-height');
-        if (!this._isHorizontal && this._isPrimaryMonitor() && extendHeight && this._fixedIsEnabled) {
+        if (!this._isHorizontal && (this._isPrimaryMonitor() || this._settings.get_boolean('multi-monitor'))
+            && extendHeight && this._fixedIsEnabled) {
```
micheleg: Is this executed on all docks? Another candidate to be moved at the manager level?

docking.js
```diff
@@ -1652,32 +1687,163 @@ const DockedDash = new Lang.Class({
...
     _showOverlay: function() {
-        if (this._settings.get_boolean('hotkeys-overlay'))
-            this.dash.toggleNumberOverlay(true);
-
-        // Restart the counting if the shortcut is pressed again
-        if (this._numberOverlayTimeoutId) {
-            Mainloop.source_remove(this._numberOverlayTimeoutId);
-            this._numberOverlayTimeoutId = 0;
-        }
+        for (let i = 0; i < allDocks.length; i ++) {
```
micheleg: This is the only part that I really don't like, with each dock being aware and handling other docks. Do we really need this? Can the overlay keyboard shortcut be moved at the dockManager level? Or, can we just have two independent signal listeners both handling the shortcut for their dock?

I couldn't run the code with multiple screens: what happens to the number shortcuts? Does it take into consideration which is the active display, in particular with the isolated workspace option?

micheleg:
I finally had time to review in more details this code. I thinks the patchset is overall very good. I was wondering if this was a good occasion to do some major code restructuring, as te code is getting a bit convoluted. But this is because of the existing historical codebase, and I don't think I will have the motivation to to put effort into it in the near future. 

Therefore I intend to proceed with the merging of this branch.

Beside the specific comments I left on the codes (take into account I haven't run the code with multiple screens):
* How is the ShowAppsButton handled? E.g., where is the apps animation from (the main dock, the activated one?)
* What's the behaviour  with isolatedWorkspace, in particular in respect to the number shortcuts?

appIcons.js
franglais125: Sure thing, I'll rename the variables!

As for the "bigger thing", it's not clear to me in the end. Would you rather I use `findMonitorForActor` or not? I can test if it works for instance.

appIcons.js
franglais125: **Settings per dock**

I hadn't thought about it. I can check if it's simpler or not, I can't really tell quickly. I'll get back to you on this point.

**Completely reloading the extension**
When toggling multi-monitor, when a monitor changes or when the position is changed, I'm removing all docks and creating new ones within the dockManager class. In particular I found this helps a bit with the feeling of 'jerkiness' when toggling the multi-monitor option, or connecting a new monitor.

For instance, when disabling/enabling the whole extension we briefly reintroduce the stock dash, and then remove it again.

intellihide.js
franglais125: This is an example of 'minimal chnage', where I tried to keep the implementation minimal. I guess we can simply remove lines 205-215 and use `this._monitorIndex` everywhere.

prefs.js
franglais125: Oops, unwanted change! I think it comes from all the rebasing I was doing.

docking.js
franglais125: Agreed, this part is simply horrible, I'll try to rethink this part. Moving it to the manager level seems like a good idea, let's see how the implementation actually looks.

As for the multiple screens: yes, it's actually quite nice, applications will be launched depending on what monitor the mouse is.

franglais125:
Thanks for the thorough review, it showed I missed a few things here and there. I'll try to address the issues you pointed out promptly. As for the major code restructuring, I'd love to help on that front too, but I don't think I will have time to work on it, at least not before the end of the summer.

**ShowAppsButton**
I don't like its current state for a number of reasons:
- The animation starts from the correct monitor, but on overview->hide the animation is directed to the primary monitor
- The overlay with all the apps shows up on the primary monitor, always. I'd have to modify the (I think) overview.js behavior to make the apps appear on the appropriate monitor. I don't think we want to do that, at least for now.

This makes the apps button actually quite useless on secondary monitors, as you'd need to move the mouse to the primary screen anyway. Although it sort of works now, I'd rather remove it from secondary monitors at least until we (if ever) introduce the app overlay on secondary monitors too.

**Workspace Isolation**
Shortcuts are introduced only once: if you try to add more with the same shortcut more than once, Mutter will complain saying that the shortcut already exists.
Workspace isolation doesn't affect this. If you treat different monitors as different workspaces, then the shortcuts will act on the dash on the given monitor. It won't try to trigger `.activate(button)` on more than one dash object. Does this answer your question?

**Merging**
I know there are a few quirks here and there (like showAppsButton), but I think it would be better to merge something like what we have now (I'll include a few more commits to fix the things in your review), and then sort out remaining bits.

Perhaps in the settings we can put something like: "Show on all monitors (experimental)", so that people can use it (lots of requests already), but at they same time they know they can expect minor quirks (and I do mean minor, not crashes).

franglais125:
I just performed a rebase to avoid some conflicts with your master branch.

New commits:
- [22e5f50d469699c0dea0edbb33e81ea7d8b7aacc] to rename monitorNumber
- [6d67ff77ae1d611a50d899a6f7b678cd9c758ac7] to use findIndexForActor. This turned out to be much cleaner!
- [af75e9471496a7ec43f63597238c66758d276cba] Remove unnecessary checks in intellihide
- [0c6999ce5d60971d167e329c1b52aa12c29e905f] Remove duplicated `hot-keys` entry.

**To do:**
- Investigate if creating several settings objects yields simpler code.
- Move some functionality to the dockManagerlevel:
     - adjustPanelCorners
     - Overlay functionality (this is the worst part)
     - forcedOverview
- I'll also wait for your thought on the ShowAppsButton.

micheleg:
Thanks for the quick updates. i agree. I will merge this, and then proceed into fixing the remaining glitches. Sorry for the huge delay, but I needed to directly look at the code to become confident into merging it. 

**showappsbutton**
For the showappsbutton behaviour, a temporary sensible option could be indeed to hide it in the secondary dock, but i think we should as well leave it as it is for the moment.  No need for a "experimental" warning in the settings, as there should not be any problem of stability.

**findIndexForActor**
My concern with this is: is the index assigned correctly? The assignment is performed when the appIcon is created. The assignement is based on the coordinates of the actor it seems:
* what if the actor is not on the stage?
* what if the dash is hidden, therefore it slide out and the coordinates of the icon correspond to the nearby monitor? 

This requires maybe some careful testing. If it prove unreliable, we revert to the previous machinery.

**workspaceIsolation**
Where is the test on the active monitor performed when triggering a keyboard shortcut? Can't find it in the code...

franglais125:
**findIndexForActor**
Amazing, you guessed right that this would fail. I just quickly tested with intellihide on, and indeed the coordinates of the actor are not correct (or some other problem behind the scenes). Let's revert that for now.

**Workspace isolation**
Hmm, I think I see what you mean. In docking.js, when using a keyboard shortcut `_activateApp` is called for the dock on the main monitor, while the application gets activated/launched on the correct monitor.
This doesn't cause a problem as one gets the desired effect. What seems off is that the app activation animation happens on the main dock. This will be another glitch to fix.

franglais125:
I won't be able to finish the remaining bits (to-do list above) before the weekend, in case you want to proceed with other PRs.

micheleg:
Thanks. I'm preparing a cleaned up branch https://github.com/micheleg/dash-to-dock/tree/multi_monitor - I just cleaned up commit messages, reordered and merged some commits for the moment. Please rebase on that in case you work on it.

franglais125:
@micheleg I'll be fine to work on this today and tomorrow.
So far, I really like your changes to the multi_monitor branch! This is somewhat what I had in mind, but you polished it really well.

Before I start producing new commits, since you now had time to get a better feeling of the code, anything knew you'd like to see changed?

Here is the list from before. To be moved to the dockManager (or their own class):
- adjustPanelCorners
- forcedOverview
- (Done by you) overlay functionality, workspace isolation

And finally, what do you think about the settings objects?

micheleg:
I had an additional commit that I forgot to push online before travelling, I just uploaded now [3a4c5f3e90]. I haven't actually tested my changes, so I could have broken something. For the settings, I don't have a clear opinion yet. It can stay like it is for the moment, and it can be simplified later. 

I'm kind of happy to go on and merge to master as soon as possible, and also make a new release to get larger feedback. Beside improvements to the code structure, what i still don't understand is how we manage the screen/workspace isolation. I very briefly tested with an external monitor, and it seems to me that the workspace isolation is not applied to the different monitor. is that the case? Is it what we want?

franglais125:
Thanks for all the work! I had to fix a few minor things:

[a17f8ffd1def0391062f549676f1e704d64cf21f] We need to remove extra features (isolation, shortcuts) every time we delete the docks, as this was creating duplication (old instances were not deleted).

[28bcf1fb2af83e7a56cb5f87a5f3506aae5000c2] Three things here:
- I reversed the logic when assigning the value for `preferredMonitor`, as I think that's what you meant to do... If not, when *not* selecting `multi-monitor`, the dock was always on the primary screen. Now, when one *does* select `multi-monitor`, the first dock is always the primary one. Does it make sense like this?
- I changed a few assignments to `this._monitor` within dockedDash (they were repetitive/wrong).
- Also, I'm saving the `preferredMonitor` value to be used in other functions (for the next commit).

[9f2bc84a857fb5faf7dd657e6f2b50087c7142cf] Move `adjustPanelCorners` to the dockManager. Perhaps we can move the `adjustLegacyTray` as well? It would make sense.

franglais125:
With respect to **workspace isolation**. I have sort of mixed feelings...

All the time I was thinking of adding an option like "isolate monitors", hence why I didn't try to "fix" what you are describing. In this way we would show applications from all monitors, but not all workspaces.

But it is true that it feels more natural (and logical) to only show applications from the current monitor, when `multi-monitor` is selected (and isolation is ON obviously). I'd be more inclined to show applications from the relevant monitor only, whenever we have multiple docks.

In any case, as you say, I think it's better to merge what we have, and improve these peripheral issues after the fact.
I'd only request that you delay merging this until tomorrow evening or Friday, as we just introduced quite a few changes, and I would like to make sure we don't break things in a bad way. I will try to test as much as possible tomorrow.

franglais125:
I added two more fixes:

- [a91d6b0ab0bde92d86c1f80224d490afaee66313] To assign the correct monitor number (I was having small inconsistencies under Wayland)
- [6e6bd0ccf7165cf5c1bcdb06fdfebce8dec0bad6] We were writing to settings on every load (my fault actually). This slows down the process needlessly. With multi-monitor, we were doing this on every call to `createDocks`. This was producing lots of errors (not all the time, but sometimes). Example: https://gist.github.com/franglais125/214cc593b808f8ac5a2957a1f3c188ce. I probably didn't notice this due to having an SSD. I guess it's worse with a HDD.

After these two fixes, I'm satisfied so far with the performance/stability.

micheleg:
Thanks for your additions and for spotting my mistakes. I have updated my multi_monitor branch. I have already  merged some of your fixes and tweak the history.  More importantly, I have tried to simplify the monitor handing code (relevant commit: 18504f3fda85f, a6fa4d89c6bc), but I haven't tested it. I hope I haven't messed up again. If everyithing is good, I'll just clean up the history and finally merge it into master.

Regarding commit [6e6bd0c]: I'm not sure I understand fully. Before we were assigning all keyboard shortcuts in the dockManager code, at each extension loading, while now this is moved to the Settings and run only when the settings are changed. Is this the only change, am I right?

franglais125:
Thanks for the cleaning! A quick test seems to show things are fine... I'll report if I find any other problems.

Regarding commit [6e6bd0c]: yes, you got it right.
Before we were assigning the special shortcut (`Super+q`) on every load, meaning that we were calling `settings.set_strv()`, which was writing to disk. Upon loading a session (or disabling other extensions for instance) this caused a bunch of errors/temporary freeze. Hence I moved this to prefs.js, where it makes more sense (and it's only called upon `changed::`, i.e. when relevant).

I think we are approaching the end of this PR.

franglais125:
I just performed a rebase on your branch, just in case.

franglais125:
I just added two small fixes for workspace isolation. Only the main dock was behaving properly when switching workspaces or when toggling the option.

franglais125:
One more thing, I've created this branch [1] to showcase what "**monitor** isolation" could be like. Perhaps this is what you wanted to have when thinking of "**workspace** isolation".

It's only one commit [354ff771e50113952537e8d6749fa1ec37f69e0e] ahead of the multi_monitor work. However, I would wait before including that work.

The logic can be summarized as adding `return w.get_monitor() == monitorIndex;` in the function `getInterestingWindows`.

[1] https://github.com/franglais125/dash-to-dock/tree/monitor_isolation

micheleg:
I finally merged this PR into master [0f640c0a56b2a3..a253f6619a73]. Compare to the multi_monitor branch, I just cleaned up the history, nothing was changed.

franglais125:
That's great! Thanks, I'll close this PR.

franglais125:
@micheleg do you mind picking up [bc920e498a240e60f96eba973886e21f96b2857d]? I included in my master branch.

Panel corners are not being adjusted properly, we introduced this when moving the function to the dockManager. Just found this!

micheleg:
Thanks, I've merged your patch.

PR #463, opened 2017-04-03
Right click menu

franglais125:
I reworked the commits on a new branch to bring it down to two commits, so that we have an easier log history.

Do you think it's ready as it is?


dash.js
```diff
@@ -78,6 +78,28 @@ const MyDashActor = new Lang.Class({
...
 
         this.actor._delegate = this;
+
+        let alignment = 0.25;
+        if (this._isHorizontal)
+            alignment = 0.5;
+        // Popup menu
+        this.menu = new PopupMenu.PopupMenu(this.actor, alignment, this._position);
+        this.menu.actor.hide();
+        this.menu.actor.add_style_class_name('panel-menu');
```
micheleg: Skipping this css class seems to fix the issue with the additional spacing between the menu and the dock when on bottom. Why do we need this class?

franglais125: Ah, thanks for spotting it! I don't think it's strictly needed, I might have added when trying different styles. Let me remove that, check and resubmit.

micheleg:
An additional thing: if the popupmenu is open, one can open simultaneusly another popupmenu of the launchers. This is at odds with the behaviour of the others menus, which can only be open one at a time.

franglais125:
As usual, your reviews are very thorough. Congrats!

I pushed two new commits to fix both issues. Thanks.

micheleg:
Thanks for the quick fixes! For the popupmenu, it's still not the same behaviour of the other launchers (which is fist click close the popupmenu, second click open a new one). I guess this is because the applications launchers are children of the actor to which this new popupmenu is attached, and therefore remain active? If so I would't know how to fix this.

franglais125:
Yes, I'd need to wire children to parents... As usual with a lot of features I propose, I try to be minimal to help with maintainability.

Let me give a try, if it gets too hard I'll leave it as it is.

micheleg:
If it gets too messy, I can live with the current status. I'm mainly concern about it hidings additional weird behaviours.

franglais125:
Ok, I'll leave as it is. We should then perhaps open an issue to keep this in mind. Is it ok?

franglais125:
I just rebased against current master, and added a commit simplyfying the logic in the _rightClickMenu function.

franglais125:
I rebased this branch on master (after the multi-monitor merge), and added a commit to support multiple docks.

franglais125:
>If it gets too messy, I can live with the current status. I'm mainly concern about it hidings additional weird behaviours.

I've tried a few times to fix this remaining issue. I wanted to know if you have perhaps an idea on how to go about this.

When the popup menu is shown, all elements on the screen behave properly. It's only the appIcons that are still clickable.

franglais125:
@micheleg I think I've finally found a solution to the right-click menu behavior in [59aac6ed43dd85e5034562e0678e438c250cc5e3].

I tweak the "reactiveness" of the AppIcons with the `menu-state-changed` signal.

I think this was the only remaining issue here.

franglais125:
@micheleg I just had to perform a rebase due to the reworking of the `showAppsIcon` wrapper to fix a small inconsistency.

franglais125:
Rebased following the code reorganization!

Issue #385, opened 2016-10-03
dash-to-dock seem to ignore dconf global settings


highvoltage:
Use case: install dash-to-dock settings system-wide.

I have an override file in /usr/share/glib-2.0/schemas/ that contains the following:
"""
[org.gnome.shell.extensions.dash-to-dock]
apply-custom-theme=true
dock-fixed=true
show-apps-at-top=true
extend-height=true
"""

When I run dconf-editor as a brand new user, I can see that those settings have been applied in dconf. However, those settings do not seem to take effect on the dash itself.

If I put that same override file in /usr/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/schemas and run:
glib-compile-schemas /usr/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/schemas/
Then it gets applied on my dash.

Would it be possible to make dash-to-dock also read my global dconf settings?

micheleg:
Hi, I haven't played with system wide installation of the extension, therefore I need to investigate how it works. Have you done this sort of system wide configuration with other extensions? 

I guess the extension  is installed system-wide as well in  /usr/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com

highvoltage:
Hey Michele! Yes, that is correct, it gets installed in /usr/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com.

Here's an example of a file called "95_dashtodock.gschema.override":
```
# Dash to Dock extension
[org.gnome.shell.extensions.dash-to-dock]
apply-custom-theme=true
dock-fixed=true
show-apps-at-top=true
extend-height=true
```

If I install it to '/usr/share/glib-2.0/schemas/' and recompile my gschemas, it should work, but doesn't.
If I install it to '/usr/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/schemas' and recompile the gschemas then that works fine.

This works fine with all the other Gnome Shell extensions I'm using. If you're looking for an example, if it might help, https://extensions.gnome.org/extension/750/openweather/ is one of them that works fine.

micheleg:
Thanks a lot for the additional information  You say with  the openweather extension you can tweak the settings system-wise. To be sure, how is the openweater extension installed (through the source makefile or a distribution package)?

franglais125:
[...]

PR #466, opened 2017-04-08
Show Apps button at top by default

jbicha:
Reasons for this change
--------------------------
1. Since the Activities button (or Applications button if you enabled *that* extension) is at the top left, it makes sense for the Apps button to be at the top left closest to it.
2. Since Unity is being discontinued (or at least no longer Ubuntu's flagship desktop), maybe it's a good idea to provide a default that's more similar to what Unity had.

Reasons against this change
------------------------------------
1. The current default has been here a long time.
2. The current default is more similar to Windows.

Other Info
-------------
Similar reasoning for extend-height to be enabled by default?
These are 2 of the 4 changes suggested by the Debian maintainer @highvoltage in https://github.com/micheleg/dash-to-dock/issues/385

### DIFF
--- a/schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
@@ -177,7 +177,7 @@
       <description>Show appplications button in the dash</description>
     </key>
     <key type="b" name="show-apps-at-top">
-      <default>false</default>
+      <default>true</default>
       <summary>Show application button at top</summary>
       <description>Show appplication button at top of the dash</description>
     </key>


micheleg:
Thanks for your thoughts and thanks for the argumentations on the pros and cons. 

I disagree on making the panel mode the default. The reasons are a purely personal preference (I even pondered many times about dropping the panel mode... but don't worry, I won't do it) but more importantly sticking with my original motivation to develop the extension: let's move the dash to the desktop to have a standard dock, without trying to completely changes the default gnome-shell experience. For the same reasons, I like to keep the default positioning on the left (although I lately personally switched to a bottom positioned dock). For the same argument I'd keep the button at the bottom, but I might rethink this.

I saw a significantly renewed interest for the extension since the spreading of the news about Ubuntu dropping Unity. And I also took unity as an inspiration. But my first target is not trying to reproduce the Unity behaviour, although I'm happy if other people can use the extension for this end, but rather tweaking the Gnome Shell workflow.

jbicha:
One problem with having the dash autohide is that it means that it's not as useful for touch screens. (On the other hand, intellihide is a cool feature and I believe Unity once had a similar feature enabled by default for a while but I think it was disabled by default because of usability issues.)

Anyway, this extension is packaged in Debian/Ubuntu and we might change some of the defaults there.

micheleg:
Yes, i really liked Unity intellihide (which was later dropped) when I was trying around Gnome 3.2 and Unity. Ended up implementing intellihide in Gnome Shell before updating my system to Gnome 3.

I think customizing the defaults at the package level would be the easiest choice. I don't know if I can do something to simplify it (I haven't sort out #385 for instance).

jbicha:
@micheleg I can't duplicate #385 after https://gitlab.com/jbicha/gnome-shell-extension-dashtodock-packaging/commit/89ef6ac 

(I think the schemas ought to be automatically installed to /usr/share/glib-2.0/schemas/ when doing a system install but I haven't looked into tweaking your Makefile to do that yet.)

AsciiWolf:
Any news regarding this PR?

AsciiWolf:
@jbicha I think that, instead of changing the default position, it could be a good idea to add option to change the Show Apps button position to GNOME Control Center.

janopae:
Moving the Apps button to the top has one downside regarding usability, as it will not bee in a screen corner any more. Putting things into screen corners (or edges) however is important regarding usability, as it makes things more reachable with a mouse (as you don't have to stop moving the mouse at the right place and therefore basically can't miss it). The activities button already uses this trick in the top-left edge, the bottom-left edge is unused in default Gnome. Therefore, it would be reasonable to put things into screen corners whenever possible.

If "Unity had it on the top" is the only argumentation for putting it there, it seems to be reasonable to provide it as an option for people who want to recreate Unity, but putting it there shouldn't be the default option, as it is disadvantageous from a usability perspective.

PR #474, opened 2017-04-16
Offer minimize or overview click action

bchauvin:
**Goal**
Offer a new click action option that will allow a more Unity-like behavior:
- when there is a single window, either minimize it or restore it
- when there are multiple windows, launch overview to select window. 

**Notes**

_This is my first attempt at contributing to a GNOME shell extension, please bare with my lack of knowledge ;-)_

- I could not find an option in the gnome shell API in order to filter out windows for a specific application in the overview, it could be a cool improvement as it would give a better consistency. If you have any information on this aspect I'd love to know more.
- Tried to follow your formatting. I did not launch a beautify on the source files as I'm not familiar with the conventions you use yet.
- I can translate my contribution in french if you guide me on how to do it.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -7,15 +7,15 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Dash to Dock\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-03-28 13:30-0400\n"
-"PO-Revision-Date: 2015-09-10 19:41+0100\n"
+"POT-Creation-Date: 2017-04-17 10:07+0200\n"
+"PO-Revision-Date: 2017-04-17 10:31+0200\n"
 "Last-Translator: Thomas GONET <th.gonet@gmail.com>\n"
 "Language-Team: Jean-Baptiste Le Cz <jb.lecoz@gmail.com>\n"
 "Language: fr\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"X-Generator: Poedit 1.6.10\n"
+"X-Generator: Poedit 1.8.7.1\n"
 
 #: prefs.js:98
 msgid "Primary monitor"
@@ -78,8 +78,8 @@ msgid ""
 "Temporarily show the application numbers over the icons, corresponding to "
 "the shortcut."
 msgstr ""
-"Afficher temporairement les numéros d'application sur les icônes, correspondant"
-"au raccourci clavier."
+"Afficher temporairement les numéros d'application sur les icônes, "
+"correspondantau raccourci clavier."
 
 #: Settings.ui.h:7
 msgid "Show the dock if it is hidden"
@@ -90,8 +90,8 @@ msgid ""
 "If using autohide, the dock will appear for a short time when triggering the "
 "shortcut."
 msgstr ""
-"Si le masquage automatique est actif, le dock apparaîtra temporairement "
-"lors de l'utilisation du raccourci."
+"Si le masquage automatique est actif, le dock apparaîtra temporairement lors "
+"de l'utilisation du raccourci."
 
 #: Settings.ui.h:9
 msgid "Shortcut for the options above"
@@ -247,8 +247,8 @@ msgid ""
 "Enable Super+(0-9) as shortcuts to activate apps. It can also be used "
 "together with Shift and Ctrl."
 msgstr ""
-"Activer Super+(0-9) comme raccourcis pour activer les applications. On "
-"peut l'utiliser aussi avec Shift et Ctrl."
+"Activer Super+(0-9) comme raccourcis pour activer les applications. On peut "
+"l'utiliser aussi avec Shift et Ctrl."
 
 #: Settings.ui.h:46
 msgid "Use keyboard shortcuts to activate apps"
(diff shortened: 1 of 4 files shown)


micheleg:
Hi Bertrand,

thanks for you contribution which is very appreciated. I very much like the inclusion of such option. As per your comment:

```// TODO: only show current app windows when gnome shell API will allow it```

this would indeed be the ideal behaviour (is it the unity one, I don't remember?). I've wanted to implement it for a long time (#216) but never manage to get anything working.

That said, I think this is a good first step.  I will try to merge it as soon as possible. Beside some testing -- it seems to be working fine for me at the moment, but there might be some hidden corner cases (isolated workspaces?) -- an minor style, I definitely want to  avoid the workaround on triggering the overview.

it seems to work because you are calling a non existent function (```console.log()```). Logging can be done with ```global.log()``` instead.  You should be able to look at this and other errors with this command in a terminal (might require su/sudo):
```journalctl  -f /usr/bin/gnome-shell```

This error then prevent the execution of [this line](https://github.com/micheleg/dash-to-dock/pull/474/commits/7b05d7d4c753a0a1df0735a4cd43dbbd050b4bb9#diff-f2c18b3123458b0b36d6cad2215ed52fL436), which hides the overview. 

Therefore the logic needs some reworking.

For translation some hints are here:
https://micheleg.github.io/dash-to-dock/localization.html

**Comment by bchauvin** [comment]
Hi Michele,

Thanks a lot for this quick review, that's appreciated too!

Unity does indeed filter out windows for a specific application when showing the overview. It should be possible within gnome shell too, but it doesn't seem to be implemented or exposed ( at least not in https://git.gnome.org/browse/gnome-shell/tree/js/ui/overview.js ).

For the record, Unity's way to deal with multiple windows is the following:
- first click shows the first window, 
- second shows the overview,
- third hides all windows 
It could be mimicked, but I preferred to go for something that is more predictable as I think it's easier for everyone that way (easier to understand and easier to maintain).

Following your recommendation, I tested the isolated workspaces option. Only windows from the current workspace are taken into account when clicking, and it makes sense with the number of dots the extension is showing as well. So it seems to be working fine to me, but more testing is always welcome of course! I'll try to test with various situations and will let you know if there's anything to be noticed.

The workaround problem is refactored, sorry for not noticing that line earlier... The project files are not properly indented in my IDE and it confuses me a lot. It should be working ok now, although a more elegant approach may still be possible.

Thanks for the translation hint! I gave it a try but my option does not appear in the pot file (neither do other configuration options). Maybe there's something I overlooked in the process, I'll give it another try later on ;-)

**Comment by franglais125** [comment]
@bchauvin to include new strings in the translation file, you can run `make mergepo`

For instance, on your branch, I get the following new lines in `po/fr.po`

```
#: Settings.ui.h:50
#, fuzzy
msgid "Minimize or overview"
msgstr "Minimiser la fenêtre"
```

**Comment by bchauvin** [comment]
Thank you @franglais125 you saved me a headache! I was about to look everywhere to see where I forgot to allow the string to be translated :) For some reason most of the strings didn't appear until I copied the `settings.ui.h` file in the `po` folder (don't worry, I removed it afterwards). 

Anyway, I was able to add the translation, it's in the latest commit. I'll make a separate pr for some more translations update as it would be out of scope here.

**Comment by micheleg** [comment]
Looks good to me.
I have prepared a branch for testing https://github.com/micheleg/dash-to-dock/tree/minimize_or_overview; I just fixed the indentation.

**Comment by micheleg** [comment]
To get the single app view, i looked into some way of patching this function in the past: https://git.gnome.org/browse/gnome-shell/tree/js/ui/workspace.js#n1819

However, i think the right thing to do is to create a new view, possibly extending and patching the workspace class https://git.gnome.org/browse/gnome-shell/tree/js/ui/workspace.js#n1097

There's this old branch https://github.com/micheleg/dash-to-dock/tree/experimental_appwindowsview where I was starting to tinker around.

**Comment by bchauvin** [comment]
After some more extensive testing, almost everything seems fine.

I only noticed something that could either be considered a design choice or a minor bug depending on the perspective. 

_How to reproduce_
- Open two windows for two different applications (e.g. firefox & terminal)
- Click on firefox icon => overview is triggered
- Click on terminal icon => overview is closed (because it's calling the toggle() method)

_Possible solutions_
Replacing toggle() with show() at https://github.com/micheleg/dash-to-dock/pull/474/files#diff-f2c18b3123458b0b36d6cad2215ed52fR426 : could be a solution, but then when clicking twice on the same icon the overview would remain open instead of hiding it as it is the case currently. In terms of user experience I like the toggling better, but that's still a possibility.

Remember which icon triggered the overview: then we'd need some kind of event we could catch when the overview is hidden or it would have bad side effects when the user closes the overview in another way (for instance using the keyboard). It would be clean, but not sure it's possible.

_My thoughts_
I would keep things as is with toggling. Overview shows all windows, so a user would get no benefit clicking another icon that has multiple windows while the overview is triggered for now. If we manage to implement window filtering (thanks for the info btw, I'll look into it when I can ;-)) then it would definitely need some adaptation. But in this scenario, I guess we'd trigger the overview differently anyway so it might not be a problem anymore.

**Comment by micheleg** [comment]
Great! I agree, we can keep the current behaviour, and fix it if we implement an application restricted overview. i'll merge the code as soon as possible (I have another pull request to merge first).

**Comment by micheleg** [comment]
There would be one thing to think about: a better name for the option, as "Minimize or overview" is not very clear. But I don't have a better idea right now.

**Comment by bchauvin** [comment]
I also feel like the name could be better. Here are some alternatives in no particular order (I put everything I thought of here in case it brings inspiration to someone). My thoughts below each:

1. Adaptive
Too vague.

1. Toggle minimization or overview
Feels ok.

1. Pick a window or toggle minimization
My favorite. I like it put that way because I understand implicitly that behavior adapts to the windows number.

1. Toggle minimization or windows selection
Not sure it's very clear for someone who doesn't already know what the option does.

1. Toggle minimization (single window) or overview (multiple windows)
Feels ok but maybe too long / too detailed. It would fit best in a description or in a tooltip, not in the option title.

1. Unity-like
Absolutely terrible, don't choose that. Not really true, people are not supposed to know unity and its behavior could be subject to change.

1. Window picker
Not so bad but doesn't describe the feature accurately, I expect the overview when reading that.

**Comment by micheleg** [comment]
I've merged the feature into master [2102c77ca1a], and also the French translation [5f928665fd3].

File: po/fr.po [hunk 1]
```diff
@@ -78,8 +78,8 @@ msgid ""
 "Temporarily show the application numbers over the icons, corresponding to "
 "the shortcut."
 msgstr ""
-"Afficher temporairement les numéros d'application sur les icônes, correspondant"
-"au raccourci clavier."
+"Afficher temporairement les numéros d'application sur les icônes, "
+"correspondantau raccourci clavier."
```
zapashcanon: `"correspondant au raccourci clavier."`
missing space

zapashcanon: @micheleg there's one missing space, I don't think it's worth a PR so maybe you could fix it directly ?

franglais125: I'm addressing this in #494

PR #504, opened 2017-05-14
Click window preview

franglais125:
Hi,

Here is the PR for adding a new click option: window previews.

I actually have two branches:
1. https://github.com/franglais125/dash-to-dock/tree/click_window_preview_dtp (this PR)
2. and https://github.com/franglais125/dash-to-dock/tree/click_window_preview

I like 1) better, as it is smoother and looks better. It is based on the work added on windowPreviews.js from dash-to-panel.

Number 2) has the advantage of being more straightforward and includes fewer changes.

Let me know what you think.

Best,
Fran

### DIFF
--- a/windowPreview.js
+++ b/windowPreview.js
@@ -197,7 +197,8 @@ const WindowPreviewList = new Lang.Class({
 
         while (newIndex < newWin.length || oldIndex < oldWin.length) {
             // No change at oldIndex/newIndex
-            if (oldWin[oldIndex] == newWin[newIndex]) {
+            if (oldWin[oldIndex] &&
+                oldWin[oldIndex] == newWin[newIndex]) {
                 oldIndex++;
                 newIndex++;
                 continue;
--- a/windowPreview.js
+++ b/windowPreview.js
@@ -510,6 +510,7 @@ const WindowPreviewMenuItem = new Lang.Class({
 
         Tweener.addTween(this.actor,
                          { height: 0,
+                           width: 0,
                            time: 0.25,
                            delay: 0.25,
                            onCompleteScope: this,
--- a/windowPreview.js
+++ b/windowPreview.js
@@ -278,15 +278,9 @@ const WindowPreviewList = new Lang.Class({
                 separator_index++;
 
         if (separator_index > 0 && separator_index !== newWin.length) {
+            // Include an empty space to separate windows from other workspaces
             let separatorItem = new PopupMenu.PopupSeparatorMenuItem();
-            if (this.isHorizontal) {
-                separatorItem._separator.destroy();
-                separatorItem._separator = new St.Widget({ name: 'dashtodockPreviewSeparator',
-                                                style_class: 'popup-separator-menu-item-horizontal',
-                                                x_expand: true,
-                                                x_align: Clutter.ActorAlign.CENTER });
-                separatorItem.actor.add(separatorItem._separator);
-            }
+            separatorItem._separator.destroy();
             this.addMenuItem(separatorItem, separator_index);
         }
 
--- a/windowPreview.js
+++ b/windowPreview.js
@@ -364,6 +364,10 @@ const WindowPreviewMenuItem = new Lang.Class({
         let labelBin = new St.Bin({ child: label,
                                     x_align: St.Align.MIDDLE});
 
+        this._windowTitleId = this._window.connect('notify::title', Lang.bind(this, function() {
+                                  label.set_text(this._window.get_title());
+                              }));
+
         let box = new St.BoxLayout({ vertical: true,
                                      reactive:true,
                                      x_expand:true });
@@ -574,9 +578,15 @@ const WindowPreviewMenuItem = new Lang.Class({
             this._windowAddedId = 0;
         }
 
-        if (this._destroyId > 0)
+        if (this._destroyId > 0) {
             this._mutterWindow.disconnect(this._destroyId);
             this._destroyId = 0;
+        }
+
+        if (this._windowTitleId > 0) {
+            this._window.disconnect(this._windowTitleId);
+            this._windowTitleId = 0;
+        }
     }
 
 });
--- a/windowPreview.js
+++ b/windowPreview.js
@@ -1,3 +1,9 @@
+/*
+ * Credits:
+ * This file is based on code from the Dash to Panel extension by Jason DeRose
+ * and code from the Taskbar extension by Zorin OS
+ * Some code was also adapted from the upstream Gnome Shell source code.
+ */
 const Clutter = imports.gi.Clutter;
 const GLib = imports.gi.GLib;
 const Lang = imports.lang;
--- a/windowPreview.js
+++ b/windowPreview.js
@@ -269,6 +269,16 @@ const WindowPreviewList = new Lang.Class({
                 item.actor.destroy();
         }
 
+        // Separate windows from other workspaces
+        let ws_index = global.screen.get_active_workspace_index();
+        let separator_index = 0;
+        for (let i = 0; i < newWin.length; i++)
+            if (newWin[i].get_workspace().index() == ws_index)
+                separator_index++;
+        if (separator_index > 0 && separator_index !== newWin.length) {
+            this.addMenuItem(new PopupMenu.PopupSeparatorMenuItem(), separator_index);
+        }
+
         // Skip animations on first run when adding the initial set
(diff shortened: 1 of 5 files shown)


franglais125:
screenshot from 2017-05-13 23-10-15

franglais125:
It looks like I'll have to start testing against 3.24 more often to check for new warnings.

- [b5e75f6f977414aa3731b79bd0190d5d55a9e88d] to fix warnings I introduced in windowsPreviews.js
- [5bbccbe60b023dca43768558d87596746e992a69] to fix an old warning in dash.js

windowPreview.js
```diff
@@ -3,14 +3,280 @@ const GLib = imports.gi.GLib;
...
+    Name: 'WindowPreviewMenu',
+    Extends: PopupMenu.PopupMenu,
+
+    _init: function(source, settings, monitorIndex) {
+        this._dtdSettings = settings;
+
+        let side = Convenience.getPosition(settings);
+
+        this.parent(source.actor, 0.5, side);
+
```
micheleg: I think we need this for intellihide:
```
        // We want to keep the item hovered while the menu is up
        this.blockSourceEvents = true;
```

micheleg:
I'm definitely for the first option (this PR). It seems to work reasonably well once the above fix is applied for intellihide.  I think this overall has already been tested in dash to panel so it is a very good start.

A couple of minor and very minor issues:
* the tooltip with the app name is not removed when the popupmenu is shown
* Would be nice to have a transition when a windows is closed

To think (non necessary for the first implementation but maybe for further future refinement): 
* what to do in overview? 
* scrolling or windows thumbnails resize? Or what about a multiline grid?

I will look at the code more carefully.

micheleg:
Also, as I guess this code comes from dash-to-panel, we should add the relevant credits to the code.

I'm also wondering what should be the windows ordering. For consistency it should probably be the same of <super>+tab, and what we do in thumbnails integrated in  the right-click menu. It seems that now the thumbnails are not reordeded so that windows of the current workspace appears first. But then maybe one could prefer to have a static order so it's easier to jump from one window to another remembering its position?

After using it for a while, I like this feature a lot. I still think it showing the previews in the overviews would be more of a GNOME Shell way, but this is a very functional alternative.

franglais125:
Hi, thanks for testing and for the quick review!

I also had some more things in mind, but kept it simple at first to see what the feedback was.

1) I'll try to fix the tooltips
2) What do you mean by a transition? Currently there is a small animation when a window is closed.
3) Overview: I agree, it'd be better to have the list of thumbnails when in overview as well. I had it close the overview as it was more consistent with the rest of the click actions.
4) I like more the look of scrolling (and perhaps add a resize). But let me know what your preference is (feedback anyone else?). What do you think about having a more responsive size that depends on, for instance
 - number of windows
 - screen size
5) Window ordering: I hadn't thought about this! I'll see how I can improve it, or at least have a reasonable default.

I don't think I can work on this before Tuesday or Wednesday. Feel free to push some commits, I'll rebase as usual if you do so. I'll add credit where credit is due (@jderose9 :))!

micheleg:
2. It seems the transition only works  for me when the previews are displayed vertically (with dock on left)
3. Not sure yet.
4. For the moment scrolling is good. Eventually, maybe the spacing and preview size can be optimized a bit. Very good idea to think about different screen size ( I have completely different feeling of the shell now with a fullHD display now, compared to when I had a an HD screen).

jderose9:
Very nice! I think this behavior is close to matching windows and KDE so it seems like a great option for people transitioning over.

Regarding credit, I did rework the event handling to fix some stuck focus bugs, but the initial code for the previews came from zorin taskbar.

franglais125:
Added:
- [98bb3bab2f49e44847ec9956e2597bb7eb0634a5] fix for intellihide
- [3d17c9b5edc0f6679833102e0b75092c5610af4a] added credit
- [d7387705b167230000f077cca2e86cfe02206d67] fix horizontal transitions

franglais125:
@micheleg I just committed some more improvements.

Missing/to fix:
1. The separator looks too big when the dock is horizontal.
2. Responsive size of the thumbnails based on number of windows/screen size. We can leave this work for later.

micheleg:
Thanks for your work. Yes I just noticed 1. 2. can definitely wait.

franglais125:
I added a fix by replacing the separator with a new one with a different style.

micheleg:
> windowPreview.js: add signal to update the window title.

Wonderful! I was just noticing this "issue" this morning.

franglais125:
Heh, awesome! Let me know if you a problem with that, as it seems to cause some problems in dash-to-panel. Not sure, though, I opened a PR over there to hopefully address that.

franglais125:
I just rebased to make use of utils.js instead of convenience.js. The branch is now more Gnome 3.24-friendly.

franglais125:
New rebase, due to two commits being cherry-picked from here.

micheleg:
Looks quite good to me. Any known issue?

One minor thing: I find that the separator between windows of different workspaces is very subtle (almost invisible) but it introduces (in horizontal mode) a quite large space. The space is probably the same in vertical mode, but as the thumbnails aspect ration tend to leave quite a bit of horizontal padding the final visual effect in horizontal mode is not the best. Do we even really need such separator (The <Super><tab> for instance doesn't seem to have one, but it also follow a different logic that doesn't differentiate among workspace)? Can we make it visible but avoid wasting space?

franglais125:
I haven't found new issues myself, I guess we'll have to wait for more feedback from different configurations/use cases.

Regarding the separator: I use always use workspace isolation, so your experience with this might be better. If you think no separator is needed, we can simply remove the line, and just keep an empty (thin) space.

I remember spending some time trying to get a nice looking separator, but honestly on Gnome 3.22 (with Adwaita) they are essentially invisible. They were much more prominent on 3.14 as I recall.

I just added [a03d139bc2f55f6e5f35fe734e8be27c4d949e63], to try to address this, but also to point out how we can modify the style. If you want to take a look, it's at the bottom of stylesheet.css.

- Thinner margin, so that the separation is not so wide:
```
-    margin: 32px 6px;
+    margin: 32px 0px;
```

- Color in rgba, so that we can tune it a bit. Although this will be very theme-dependent. Probably not much we can do here.
```
-    border-color: #343a3a;
+    border-color: rgba(52,58,58,1);
```

- Made the line itself thicker, so that it's more visible:
```
-    border-right-width: 1px;
+    border-right-width: 2px;
```

franglais125:
The last commit removes the line altogether (and the styling). It's an optional commit, so you can choose to pick or not. IMO it looks ok without the line, it's just a small empty space now.

micheleg:
I've uploaded an update branch https://github.com/micheleg/dash-to-dock/tree/click_window_preview.
Beside consolidating few commits, I've done some minor changes (hopefully improvements :))

[416b33360620b083337c1aeed2b508c2301e30ab]: I follow the logic of used for the default popupmenu of checking for the open status. Avoid defining additional status variable.

[898c880c9ddf2a55c3b9cb86e87bc96618e58c2a]: I tweak the style of the existing separator instead of creating a new one. I add the additional style_class instead of replacing it with a new one. In this way the style is inherited from the default popupmenu, and only the relevant properties need to be tweaked in our css. In this way it should pick up the custom themes styling like colors etc.

I went for keeping the line for now.

Unless I introduced bugs with this, I will merge into master.

franglais125:
Great, I'll test it a bit to see if I find any problems. The differences are minor, so I don't expect that will be the case! If I don't comment again, you can assume I didn't find anything worth reporting.

franglais125:
Just to confirm: the line separator is only visible when the dash is vertical.

Perhaps this is what you intended, but even if it's not, I'd count it as a minor thing.

micheleg:
I've lost something during a rebase. This should fix it [d385f492d].

micheleg:
This was merged into master [07728149cc88bd434a717710c2ce4fe2c179c2c1]..[3c9f7f3be676d1a781fabc7f711a58416cbaead0].

micheleg:
@franglais125 what do you think of this [848c47abb3b4a29bc8ace630b74dd770b0b08920]?

franglais125:
That sounds very reasonable. It took me a while to understand what changed, but once I understood the logic, I think it actually works very nicely!

micheleg:
Merged in master [5003841bd005d71737fc192afe7fbe8a76512f17].

PR #509, opened 2017-05-28
Move some functions and classes from convenience.js to utils.js.

franglais125:
Perhaps we can also reference the issues it fixes to the changelog.

I tested this on Gnome 3.22 and Gnome 3.24

Also, on 3.24, I purged:
```
gir1.2-clutter-1.0
gir1.2-clutter-gst-3.0
gir1.2-gtkclutter-1.0
```
and the settings dialog loaded fine. So, as you mentioned, it seems that the gir dependencies are no longer a problem. This is a nice plus!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 UUID = dash-to-dock@micxgx.gmail.com
 BASE_MODULES = extension.js stylesheet.css metadata.json COPYING README.md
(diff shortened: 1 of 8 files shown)


micheleg:
I've merged the patch into master [9350417f090ac2a3a253]. Thank you.

PR #510, opened 2017-05-29
Make the drag placeholder smaller than other icons

jackstanek:
This matches the native shell more closely, and also reduces the
chances that the drag placeholder will look blurry or pixelated. I'm
not sure if there's a better factor to divide by, in order to more closely
match the native shell, but it seemed to look fine to me.

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -1016,8 +1016,8 @@ const MyDash = new Lang.Class({
                 fadeIn = true;
 
             this._dragPlaceholder = new Dash.DragPlaceholderItem();
-            this._dragPlaceholder.child.set_width (this.iconSize);
-            this._dragPlaceholder.child.set_height (this.iconSize / 2);
+            this._dragPlaceholder.child.set_width (this.iconSize / 2);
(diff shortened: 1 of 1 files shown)


micheleg:
Hi Jack, 

thanks for your contribution. I'm not sure I understand what you mean by matching the native shell. That part of code comes straight away from upstream

https://git.gnome.org/browse/gnome-shell/tree/js/ui/dash.js#n912

therefore I would expect it to produce the same results. Could you explain further (maybe a screenshot)?

jackstanek:
Hi,
After some further inspection, it seems that the problem actually stems from the Gnome Shell theme I was using, where natively the drag placeholder looked fine, but in the dock, it looked way larger than it should have: you can see the [dock](https://i.imgur.com/ODKl8vh.jpg) versus [the native dash](https://i.imgur.com/defo3xV.jpg)... I suspected it had to do with the icon sizing code, but perhaps it has to do with the icon selection (that is, the dock is choosing the wrong size of the drag placeholder icon?)

Issue #434, opened 2017-01-14
Active Task not easily distinguishable


fastrizwaan:
Hi,

Both Active and running applications show same color of Dots. Also the lighter box around the icon for current active window, is too subtle and not easily noticeable. Please see the screenshot, the dash-to-dock settings in Dash-to-Dock (lighter background) and in Dash-to-Panel (highlighted underline).

We should have something more prominent so that active application can be noticed easily visually. Thanks.

![screenshot from 2017-01-14 16-06-14](https://cloud.githubusercontent.com/assets/1161682/21954350/2252f5c2-da74-11e6-8c99-f4f1ce13f718.png)

fastrizwaan:
If we enable custom themes, then the highlighter is almost invisible. Please see that nautilus is currently active, and if we look at dash it is difficult to distinguish and on panel it is very clear with the underline.
![screenshot from 2017-01-14 16-18-20](https://cloud.githubusercontent.com/assets/1161682/21954392/4eebc158-da75-11e6-8a29-c66b81b0de09.png)

fastrizwaan:
Could we use Underlines like gtk3 tabs on left side?

micheleg:
If we use underline, don't we lose the windows counter indicators? I agree someone might like a more prominent indicator for the active application, but this doesn't look like a good solution. 

What do you mean by "If we enable custom themes" ?

fastrizwaan:
I use zuki theme, which is a custom theme, and the highlight is even less prominent.

A dash followed by dots

We could have dash+dots together in this manner
_...  = first window active rest are minimized
._.. = second window active; 1st and others are minimized/inactive
_._ = 1st and 3rd window active 2nd is inactive/minimized

or if we don't want to go too deep, then active dash with dots in the center, could show both multiple windows and active window.

__.__ = 1 active 1 inactive
__..__ = 1 active 2 inactive
__...__ = 1 active 3 or more inactive

Thanks.

Baltix:
@fastrizwaan it seems you are suggesting Morse code ;)

micheleg:
The first idea seems a bit overkill to me, while I need to see the second visually.

From an implementation point of view, it could be done changing the function drawing the circles  [https://github.com/micheleg/dash-to-dock/blob/master/appIcons.js#L374-#L447]. I think CSS is too limited for this (unless we just want a line in the background).

redvimo:
I have an idea, not much thought went into it though..
[...]

PR #512, opened 2017-05-30
Unity7-esk backlight items support

SavageTiger:
Hi,

Since Unity7 is sort-of discontinued I switched to Gnome3, dash to dock provides most of the missing 
functionality for me.

However I always liked the look of the colorful side-bar that clearly indicated what applications are active. So I tried to replicate this functionality in dash-to-dock and I  think I succeeded.

example

I added a toggle to the Appearance tab to enable / disable the functionality.

I hope you like it.

* There is a small issue where you can only go up to 48px max icon size, since there seems to be a SVG scaling issue in Gtk itself, but other than that it works flawlessly.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -62,6 +62,9 @@ let recentlyClickedAppIndex = 0;
 // pixel buffers that can be used for calculating the backlight color
 let themeLoader = null;
 
+// Global icon cache. Used for Unity7 styling.
+let iconCacheMap = new Map();
+
 /**
  * Extend AppIcon
  *
@@ -87,7 +90,6 @@ const MyAppIcon = new Lang.Class({
         this._monitorIndex = monitorIndex;
         this._signalsHandler = new Utils.GlobalSignalsHandler();
         this._nWindows = 0;
-        this._backgroundColor = null;
 
         this.parent(app, iconParams);
 
@@ -712,9 +714,9 @@ const MyAppIcon = new Lang.Class({
      * Unity7 to javascript, so it more or less works the same way.
      */
     _calculateColorPalette: function (pixBuf) {
-        if (this._backgroundColor !== null) {
+        if (iconCacheMap.get(this.app.get_id())) {
             // We already know the answer
-            return this._backgroundColor;
+            return iconCacheMap.get(this.app.get_id());
         }
     
         let colorUtils = new Utils.ColorUtils();       
@@ -761,13 +763,15 @@ const MyAppIcon = new Lang.Class({
         let rgb = colorUtils.HSVtoRGB(hsv.h, hsv.s, hsv.v);
 
         // Cache the result.
-        this._backgroundColors = {
+        let backgroundColor = {
             lighter: colorUtils.ColorLuminance(rgb.r, rgb.g, rgb.b, 0.2),
             original: colorUtils.ColorLuminance(rgb.r, rgb.g, rgb.b, 0),
             darker: colorUtils.ColorLuminance(rgb.r, rgb.g, rgb.b, -0.5)
         };
 
-        return this._backgroundColors
+        iconCacheMap.set(this.app.get_id(), backgroundColor);
+
+        return backgroundColor;
     },
 
     _drawCircles: function(area, side) {
--- a/appIcons.js
+++ b/appIcons.js
@@ -626,10 +626,10 @@ const MyAppIcon = new Lang.Class({
     },
 
     _enableBacklight: function() {
-        let pixBuf = this._getIconPixBuf();
+        let colorPallete = this._calculateColorPalette();
 
         // Fallback
-        if (pixBuf === null) {
+        if (colorPallete === null) {
             this._iconContainer.set_style(
                 'border-radius: 5px;' +
                 'background-gradient-direction: vertical;' +
@@ -645,8 +645,6 @@ const MyAppIcon = new Lang.Class({
 
             return;
         }
-        
-        let colorPallete = this._calculateColorPalette(pixBuf);
 
         this._iconContainer.set_style(
             'border-radius: 5px;' +
@@ -713,11 +711,15 @@ const MyAppIcon = new Lang.Class({
      * The backlight color choosing algorithm was mostly ported from the C++ source of Canonicals
      * Unity7 to javascript, so it more or less works the same way.
      */
-    _calculateColorPalette: function (pixBuf) {
+    _calculateColorPalette: function () {
         if (iconCacheMap.get(this.app.get_id())) {
             // We already know the answer
             return iconCacheMap.get(this.app.get_id());
         }
+
+        let pixBuf = this._getIconPixBuf();
+        if (pixBuf == null)
+            return null;
     
         let colorUtils = new Utils.ColorUtils();       
         
--- a/appIcons.js
+++ b/appIcons.js
@@ -805,10 +805,11 @@ const MyAppIcon = new Lang.Class({
      */
     _resamplePixels: function (pixBuf, pixels, resampleX, resampleY) {
         let resampledPixels = [];
+        let limit = pixBuf.get_height() * pixBuf.get_width() / (resampleX * resampleY);
 
-        for (let i = 0; i < pixBuf.get_height() * pixBuf.get_width() / (resampleX * resampleY); i++) {
+        for (let i = 0; i < limit; i++) {
             let pixel = i * resampleX * resampleY;
-            
+
             resampledPixels.push(pixels[pixel * 4]);
             resampledPixels.push(pixels[pixel * 4 + 1]);
             resampledPixels.push(pixels[pixel * 4 + 2]);
--- a/appIcons.js
+++ b/appIcons.js
@@ -706,7 +706,7 @@ const MyAppIcon = new Lang.Class({
         }
         
         // Get the pixel buffer from the icon theme
-        return themeLoader.load_icon(iconTexture.get_names()[0], 24, 0);
+        return themeLoader.load_icon(iconTexture.get_names()[0], 64, 0);
     },
 
     /**
--- a/appIcons.js
+++ b/appIcons.js
@@ -87,6 +87,7 @@ const MyAppIcon = new Lang.Class({
         this._monitorIndex = monitorIndex;
         this._signalsHandler = new Utils.GlobalSignalsHandler();
         this._nWindows = 0;
+        this._backgroundColor = null;
 
         this.parent(app, iconParams);
 
@@ -703,7 +704,12 @@ const MyAppIcon = new Lang.Class({
      * Unity7 to javascript, so it more or less works the same way.
      */
     _calculateColorPalette: function (pixBuf) {
-        let colorUtils = new Utils.ColorUtils();
+        if (this._backgroundColor !== null) {
+            // We already know the answer
+            return this._backgroundColor;
+        }
+    
+        let colorUtils = new Utils.ColorUtils();       
         
         let pixels = pixBuf.get_pixels(),
             offset = 0;
@@ -746,11 +752,14 @@ const MyAppIcon = new Lang.Class({
 
         let rgb = colorUtils.HSVtoRGB(hsv.h, hsv.s, hsv.v);
 
-        return {
+        // Cache the result.
+        this._backgroundColors = {
             lighter: colorUtils.ColorLuminance(rgb.r, rgb.g, rgb.b, 0.2),
             original: colorUtils.ColorLuminance(rgb.r, rgb.g, rgb.b, 0),
             darker: colorUtils.ColorLuminance(rgb.r, rgb.g, rgb.b, -0.5)
         };
+
+        return this._backgroundColors
     },
 
     _drawCircles: function(area, side) {
--- a/appIcons.js
+++ b/appIcons.js
@@ -728,33 +728,27 @@ const MyAppIcon = new Lang.Class({
             rTotal = 0,
             gTotal = 0,
             bTotal = 0;
+            
+        let resample_y = 1,
+            resample_x = 1;
 
-        // Re sampling if necessary
-        let resample_x = 1;
-        let resample_y = 1;
-        if (pixBuf.get_height() == 512)
-            resample_y = 8;
-        else if (pixBuf.get_height() == 256)
-            resample_y = 4;
-        if (pixBuf.get_width() == 512)
-            resample_x = 8;
-        else if (pixBuf.get_width() == 256)
-            resample_x = 4;
+        // Resampling of large icons
+        if (pixBuf.get_height() === 512 || pixBuf.get_height() === 256 || pixBuf.get_width() === 256 || pixBuf.get_width() === 512) {
+            if (pixBuf.get_height() === 512)
+                resample_y = 8;
+            else if (pixBuf.get_height() === 256)
+                resample_y = 4;
 
-        if (resample_x > 1 || resample_y > 1) {
-            let resampledPixels = [];
-            for (let i = 0; i < pixBuf.get_height()*pixBuf.get_width()/(resample_x * resample_y); i++) {
-                let pixel = i * resample_x * resample_y;
-                resampledPixels.push(pixels[pixel*4]);
-                resampledPixels.push(pixels[pixel*4 + 1]);
-                resampledPixels.push(pixels[pixel*4 + 2]);
-                resampledPixels.push(pixels[pixel*4 + 3]);
-            }
-            pixels = resampledPixels;
+            if (pixBuf.get_width() === 512)
+                resample_x = 8;
+            else if (pixBuf.get_width() === 256)
+                resample_x = 4;
+
+            pixels = this._resamplePixels(pixBuf, pixels, resample_x, resample_y);
         }
 
-        for (let i = 0; i < pixBuf.get_height()/resample_y; i++) {
-            for (let x = 0; x < pixBuf.get_width()/resample_x; x++) {
+        for (let i = 0; i < pixBuf.get_height() / resample_y; i++) {
+            for (let x = 0; x < pixBuf.get_width() / resample_x; x++) {
                 let r = pixels[offset],
                     g = pixels[offset + 1],
                     b = pixels[offset + 2], 
@@ -796,6 +790,32 @@ const MyAppIcon = new Lang.Class({
         iconCacheMap.set(this.app.get_id(), backgroundColor);
 
         return backgroundColor;
+    },
+
+    /**
+     * Downsample large icons before scanning for the backlight color to
+     * improve performance.
+     *
+     * @param pixBuf
+     * @param pixels
+     * @param resampleX
+     * @param resampleY
+     *
+     * @return [];
+     */
+    _resamplePixels: function (pixels, pixBuf, resampleX, resampleY) {
+        let resampledPixels = [];
+
+        for (let i = 0; i < pixBuf.get_height() * pixBuf.get_width() / (resampleX * resampleY); i++) {
+            let pixel = i * resampleX * resampleY;
+            
+            resampledPixels.push(pixels[pixel * 4]);
+            resampledPixels.push(pixels[pixel * 4 + 1]);
+            resampledPixels.push(pixels[pixel * 4 + 2]);
+            resampledPixels.push(pixels[pixel * 4 + 3]);
+        }
+
+        return resampledPixels;
     },
 
     _drawCircles: function(area, side) {
--- a/appIcons.js
+++ b/appIcons.js
@@ -803,7 +803,7 @@ const MyAppIcon = new Lang.Class({
      *
      * @return [];
      */
-    _resamplePixels: function (pixBuf, pixels, resampleX, resampleY)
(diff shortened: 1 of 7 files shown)


micheleg:
Hi Sven,

thanks a lot for your contribution. I quickly tested it your code and it seems to run smoothly. It's a neat effect which adds quite a lot of colour to the shell! Not sure I personally prefer it over a more simple default theme, but I understand the appealing for people coming from Ubuntu.

Regarding the  styling I have always being a bit careful to not trying to implement every possible customization, preferring instead to leave as much as possible to the theme writers. Not everything can however be controlled by css, and you matching colouring of  the icons background is a good example.

I am considering enabling additional styles for the running/focused applications (See #434), in this would reasonably fit in that direction. I'll try to review with a bit of care your code (it seems reasonable at a first glance).

utils.js
```diff
@@ -83,6 +83,81 @@ const GlobalSignalsHandler = new Lang.Class({
...
+
+        return {
+            h: h,
+            s: s,
+            v: v
+        };
+    }
+});
+
+/**
```
micheleg: I'm just wondering if Clutter color functions could be used instead?

https://developer.gnome.org/clutter/1.26/clutter-Colors.html#clutter-color-to-hls
https://developer.gnome.org/clutter/1.26/clutter-Colors.html#clutter-color-from-hls

SavageTiger: Hi Michele,
Thank you for your quick and enthusiastic response! I am happy that there is interest in my changes, I will investigate if I can use the HLS native code (since its not the same as HSV)

micheleg: It's not a problem to have those functions,  unless there are simple equivalent already available.

micheleg:
I've noticed some performance issue with your branch, in particular when changing workspace, with the shell hanging for ~1s. The same occurs at the first activation of the option from the settings.

SavageTiger:
I think that has to do with having to calculate the color of all items every time they are created, maybe i can fix that by adding named caching. What hardware are you using? I am testing on a pretty new system (ryzen 7) so I dont really notice any slowdown.

micheleg:
I'm on a quite new laptop (i5/skylake). Now, restarting the shell I noticed the the slowdown is gone, but it was very evident before and clearly dependent on the settings for this feature being on or off. I'll see if the problem reappears.

SavageTiger:
Update; I have managed to replicate the laggy behaviour on my local machine, i'll investigate further, see if i can fix it.

micheleg:
I noticed that the glossy svg is missing in the makefile.

micheleg:
I found out that the slow down occurs with specific applications. I noticed it with slack. I noticed that it use a large icon (512x512px), even when used at lower scale. The issue could therefore be the loop over the image pixels? The colour could maybe be cached, but probably its calculation should be optimized. the simplest thing that pass through my mind is sampling only a subset (random/grid?) of the pixels.

micheleg:
Also,

the glossy background works correctly only for icons size of 48 or less. Here for instance the result at icon size 64:
screenshot from 2017-06-02 22-59-26

SavageTiger:
Maybe the glossy background should be removed, as I said in my original pull request, if you scale your images higher then 48pixels, Gtk stops properly scaling the svg's I did find some old bug reports that do not clearly seem to indicate if the problem has been fixed or even related. What do you think?

https://bugzilla.gnome.org/show_bug.cgi?id=316825
https://bugzilla.gnome.org/show_bug.cgi?id=590663

Also i ran into a different issue that my electron apps (jira and whatsapp) seem to always use a fallback, the problem here is that in those (and most likely also other cases) the shell_app_create_icon_texture() in https://github.com/GNOME/gnome-shell/blob/master/src/shell-app.c also has a fallback method that will return a ClutterTexture. I have been trying for hours to read pixels from a ClutterTexture but so far I failed, I did manage to pass it to a clutter stage and call read_pixels on that, but that made Gtk crash doe to this bug: https://bugzilla.gnome.org/show_bug.cgi?id=653590 My electron apps are made using natifier so I am not sure of Slack also has this problem since I am not a slack user. Maybe someone else knows a solution for this, since I am at the point going crazy, but since it rarely happens maybe its not even worth the affort, since there is a sane fallback option, that could be improved by adding a fallback color picker to the settings in the future.

As for the performance problem, I have tested using a local cache and had not had problems as of yet, I am at the office tomorrow, and will publish the updated source code then.

micheleg:
Those bugs are for gnome-panel(gtk), while the shell uses Clutter/St so I'm not sure they aply here.

Adding this css rule seems to fix the background scaling:
 ```background-size: contain;```

Should we keep it or not. I'm not sure. It's very unity like, which I guess was your goal. Without the glossy background, the contrast between the running apps with the colourful background and the non running apps seems a bit ugly to me.

SavageTiger:
Added caching, and fixed the svg resizing issue

appIcons.js
```diff
@@ -297,6 +312,13 @@ const MyAppIcon = new Lang.Class({
             this.parent();
         this._onFocusAppChanged();
         this._updateCounterClass();
+        
+        // Enable / Disable the backlight of running apps
+        if (this.app.state !== Shell.AppState.STOPPED && this._dtdSettings.get_boolean('unity-backlit-items') === true) {
```
franglais125: Could we also check for the number of windows? This would make the feature compatible with workspace isolation!

`&& getInterestingWindows(this.app, this._dtdSettings).length > 0`

SavageTiger:
Done and done :D

appIcons.js
```diff
@@ -689,11 +695,14 @@ const MyAppIcon = new Lang.Class({
 
         let rgb = colorUtils.HSVtoRGB(hsv.h, hsv.s, hsv.v);
 
-        return {
+        // Cache the result.
+        this._backgroundColors = {
```
franglais125: I was trying to figure out why the caching was not working... It turns out the variable here ends with an "s". So the color is actually never cached!

```
this._backgroundColor = null;
this._backgroundColors = {
```

franglais125:
I don't know how reasonable this is, but here [659982c012e7acffe36bfb39653c0805341d03ef] is a global approach to the caching mechanics. I've made it global, so even if an application is closed (or if switching workspaces with isolation enabled) the cache is kept.

For applications like Slack, the shell will only freeze the first time it's opened. A sub-sampling would probably still be needed for weak cpus.

franglais125:
Here is a patch for resampling [71e438ade18bcef2679a17b21d0faf4ba842449a].
Below you can find screenshots, where the only affected icon is Slack, as the resampling is done only for 256 and 512px.

No resampling:
screenshot from 2017-06-10 16-04-54

Moderate resampling (**patch above**):
screenshot from 2017-06-10 16-04-54

Aggressive resampling (resampling twice as hard, this seems too much):
screenshot from 2017-06-10 16-07-29


I'm getting used to this, and I am sort of liking it. Awesome work @SavageTiger !

SavageTiger:
Wow nice work fixing that performance issue!, I merged it into my branch, and did some readability refactoring. My dots still have the cool two tone effect that is missing from your screenshots, do you know why this is?

appIcons.js
```diff
@@ -747,6 +741,32 @@ const MyAppIcon = new Lang.Class({
...
+     * improve performance.
+     *
+     * @param pixBuf
+     * @param pixels
+     * @param resampleX
+     * @param resampleY
+     *
+     * @return [];
+     */
+    _resamplePixels: function (pixels, pixBuf, resampleX, resampleY) {
```
franglais125: Thanks for the refactoring!

Small bug: the `pixels` and `piBuf` inputs are swapped.
```
_resamplePixels: function (pixels, pixBuf, resampleX, resampleY)
_resamplePixels: function (pixBuf, pixels, resampleX, resampleY)
```

SavageTiger: Ow oops, good catch, I dont have any large icons so it was hard for me to test :D

franglais125:
@micheleg @SavageTiger I tried to profile the `_calculateColorPalette` function, to see how much the resampling is helping.

Using the Slack icon (only one I could find with 512px):
- 512 px: takes 3665ms
- 256 px: takes 938ms
- 128 px: takes 240ms
- 64 px: takes 60ms

The result is very linear, as expected. The resampling included in the code now downsamples 512px and 256px icons to 64px.

I got these results on a Sandybridge dual core laptop (~2011), so not exactly something new nor too powerful. At 512, it took over 3s to calculate the icon. So yes, that was too much. The resampling plus the global cache should help a lot in this respect.

Here is the code I used (line 598):
```
        global.log('Calculating colorPalette for: ' + this.app.get_id());
        let now = new Date();
        let ms = now.getTime();
        let colorPallete = this._calculateColorPalette(pixBuf);
        now = new Date();
        ms = now.getTime() - ms;
        global.log('Done in ' + ms+ 'ms');
```

SavageTiger:
My results using the Slack icon, On a Ryzen7 1700x (clocked at 3.9)

- Without re-sampling: 1249ms
- With re-sampling: 37ms
- With cache (after the first call draw-call off course): 0ms

appIcons.js
```diff
@@ -542,6 +574,199 @@ const MyAppIcon = new Lang.Class({
...
+            return null;
+        }
+
+        if (iconTexture instanceof Gio.FileIcon) {
+            // Use GdkPixBuf to load the pixel buffer from the provided file path
+            return GdkPixbuf.Pixbuf.new_from_file(iconTexture.get_file().get_path());
+        }
+        
+        // Get the pixel buffer from the icon theme
+        return themeLoader.load_icon(iconTexture.get_names()[0], 24, 0);
```
franglais125: Small suggestion regarding the default size loaded. If we increase it from 24 to 64, there is quite an impact in the end result. This is of course just a suggestion! Cheers.

24px:
Small suggestion regarding the default size requested. If we increase it from 24 to 64, there is quite an impact in the end result.

24px:
screenshot from 2017-06-12 12-19-00

48px:
screenshot from 2017-06-12 12-19-40

64px:
screenshot from 2017-06-12 12-20-05

SavageTiger: I agree, after adding cache I think that would have minimal impact, and gives us better results.

franglais125:
To help with the (eventual I hope) merge of this branch, I performed a rebase against Michele's master.
https://github.com/franglais125/dash-to-dock/commits/unity7_rebased

I certainly don't want to takeover this PR. Just trying to get it to a merge-able state!

Commits dropped:
- [31489d9] Messy prototype of unity-esk functionality and [e6818ff] Revert "Messy prototype of unity-esk functionality" , as they both cancel out.
- [84d47d6] Simplify Settings.ui changeset. This was squashed so that the changeset on Settings.ui is overall small.
- [b00c6ea] Merge branch 'master' into master. This is no longer necessary.
- [d3970c0] Merged with upstream, fixed mergeconflicts. This is no longer necessary

franglais125:
When resampling, half the time of the whole calculation is actually doing the resampling.

I managed to speed up the resampling by a factor of ~5. On my slow laptop, I can now render the (resampled) 64px in ~35ms, down from 60ms.

Here is the commit: [a282660].

franglais125:
And finally, a new optimization, this time on the main loop for the color calculation. It is now 3 times faster! [ae7d3fa91f3a6024a5ff9d8c43e0b905937fcc94].

64px icon background calculated in 11ms (14ms if resampled). Very reasonable IMO.

As a reminder, this is only the first time the color is computed, then it takes 0ms as it's cached.

SavageTiger:
Wow good work man, I did not realize js recalculates the upperbound every iteration, makes sense I suppose.
And thank you for all the help making this better. Dont worry I understand this is not a hostile takeover.  I'd say its more like teamwork :+1:

SavageTiger:
I just forcepushed your rebased branch into my forked master (I just realized without context this is a weird sentence). Hope the PR is still valid

micheleg:
Nice work!

One note: at the moment this option  doesn't get disabled when "Use built-in theme" is enabled.

micheleg:
I'm thinking of restructuring the code handling the application status style, in particular with the prospect of enabling multiples style options, including this and possible importing some from dash-to-panel (#434). To do this I tried to separate the logic to a dedicated class. I've started working on it on this branch:

https://github.com/micheleg/dash-to-dock/tree/appIconIndicators.

At the moment I'm still playing around to see if it actually makes sense, to be sure to avoid over-engineering the thing, so don't rebase on it.  I just wanted to let you know. Happy to get suggestions.

franglais125:
@micheleg That looks very promising. Just like the docking.js case when introducing multi-monitor, this is a welcome structuring of the code!

SavageTiger:
@micheleg I dont hhink that is a bad plan, stricly speaking they are there own entity.
While i was working on this I was experimenting with drawing the indicators in different shapes (arrow in this case, same Unity), but after changing the color and slightly moving them I decided that it was Ok for now.

franglais125:
I pushed two commits to https://github.com/franglais125/dash-to-dock/commits/unity7_rebased:

- https://github.com/franglais125/dash-to-dock/commit/9ace4325aad24c722d8caee69aa3e055a60de1ae: small optimization
- https://github.com/franglais125/dash-to-dock/commit/3ea0a6d3df670e5108787a58b2e23344cff21cb0 fixes the incompatibility with "Use built-in theme".

appIcons.js
```diff
@@ -623,6 +783,12 @@ const MyAppIcon = new Lang.Class({
         let spacing = radius + borderWidth; // separation between the dots
         let n = this._nWindows;
 
+        // Lightly adjust the styling in case of unity7 backlit mode
+        if (this._dtdSettings.get_boolean('unity-backlit-items') === true) {
```
franglais125: Maybe here we can check if `custom-theme-customize-running-dots` is true, and not override the padding and width from the settings. So something like:
```
if (this._dtdSettings.get_boolean('unity-backlit-items') &&
    !this._dtdSettings.get_boolean('custom-theme-customize-running-dots')) {
        padding = 1.45;
        borderWidth = 2;
}
```

nsrosenqvist:
Sorry for just jumping in here, but is it possible to have the glossiness as a separate option? I'd love to have the backlight feature but as a solid color and square, looks more modern. Or is it customizable with CSS?

franglais125:
Here is how it looks without the glossy effect:
screenshot from 2017-06-20 18-17-45


Maybe we could have a drop-down menu, with the options `[off, glossy, simple]` ?

SavageTiger:
@franglais125 Sorry it took so long for me to get back to you, there is allot going on at the moment. I just merged your changes, thanks for that.

franglais125:
@SavageTiger no worries, I'm just pushing ideas/suggestions. You certainly don't need to pull everything from that branch :)

tliron:
Many thanks to @SavageTiger for this implementation! I have borrowed your code (with credit, of course) for the extension I am working on: https://github.com/tliron/em-dash

SavageTiger:
@tliron You are very welcome. If you ever find out how to get the pixBuf if ClutterTextures (currently this PR uses a fallback), please let me know

micheleg:
@tliron very nice project, I'll definitely dig into it for inspiration!

tliron:
@micheleg It's the opposite, I was mostly inspired by you. :) If you or @SavageTiger would like to be credited with your real names, please let me know. For now I'm just using the names I have.

I will look into that problem, @SavageTiger. It's indeed not so easy because Clutter abstracts that away. With `GIcon`, too, there are a few different types that are available, but really because of GObject anyone can create their own implementation. (For example, someone might create a GIcon that renders dynamically!). By the way, you support `FileIcon`, but I don't you are checking specifically for `ThemedIcon`. Also, there is `BytesIcon`. I never saw one used, but who knows?

tliron:
Unfortunately, it seems impossible to access the ClutterTexture data directly, at least in JavaScript. Internally it uses Cogl for everything, but CoglHandle is not exposed via GObject introspection. It would be possible to write something in C to do this, but that's obviously not in our scope.

tliron:
OK! So I found a better way by looking in the `St.TextureCache` source code. You don't want to use `Clutter.Texture` at all, because actually St hands over the pixbuf data to it. Instead you want to use the `Gtk.IconTheme` to load it directly, essentially duplicating what St does. See my [solution](https://github.com/tliron/em-dash/blob/master/utils/icon.js).

In a perfect world, it might seem nice if `St.TextureCache` would expose what it already loaded so that we can reuse it. But actually, it might not be best for us. For example, if the dash shows very tiny icons, we won't have enough pixels to do the correct backlight calculation. So I think you were smart (or the Unity devs were smart?) to use a 64-sized pixbuf.

SavageTiger:
@tliron Yeah after hours of digging through the documentation and C source I kinda came to the same conclusion as you. I just hoped that there was something I overlooked. There is only one case on my test machines where I see this happen, and thats with apps created using nativefier. I am not sure if its really that big of a deal.

I actually tried to read the pixels of the render stage, but that made Gnome hard-crash due to this bug: https://bugzilla.gnome.org/show_bug.cgi?id=653590

But like I said, I dont think its that big of a deal :)

micheleg:
I'm kind of ready to merge (finally!) this feature. From my experiience it seemed to work very well. I've prepared a branch https://github.com/micheleg/dash-to-dock/tree/glossy-luanchers. Beside simple reorganizations and cleanups: 

1. I reimplemented the colorutils conversion functions. The reason is that, even though those code snipppets seem to be a bit everywhere on the internet, i could't find a clearly licensed source. I realized the math used is on Wikipedia so I just wuicly rewrote the functions. From my test the conversion is exactly the same. They are probably not as optimized as the originals, but we are not abusing of them. I hope I haven't introduced any bug, in partucular for some corner case.  I also moved the colorutils to a "namescape" rather than a class, as it doesn't make sense to instantiate an object all the time we want to use these utility functions.

2. I slightly changed the code computing the resampling factor, to be more generic. I aim at resampling down to 64 px as before. If for any reason the icon size is not specifically ~ 2^n (128, 256, etc), it will resample to something between 64 px and 128 px. I start resampling for size > 128 px, as - although I haven't tested it - I guess that the resampling plus dominance color calculations is not more efficient than calculating it directly (If someone wants to test this we can change thre details)

3. I slightly changed the updating functions, merging it with the dots one. Proper refactoring will hopefully follow.

I'm going to merge al the fixing/refactoring/performance commits and then proceed to merge into master if no problem or bugs is found.

franglais125:
@micheleg Thanks for all the refactoring of the code!

At the moment, the feature is not really working for me. It seems that the color is just not there. Here are two screenshots:

Before refactoring:
screenshot from 2017-08-20 11-15-17

After refactoring:
screenshot from 2017-08-20 11-15-59


I'll try to see if I can find the issue. Did you find the same thing?

micheleg:
Thanks a lot for the testing. It's working for me  (on 3.25.90). Maybe it depends on some specific settings.

franglais125:
Ok, same as you... Work on 3.25.90 (Ubuntu), doesn't work on 3.22 (Debian 9). I'll keep investigating.

franglais125:
Found the problem: I am using `isolate-monitors` on my Debian setup. I'll check the code for this.

franglais125:
@micheleg Ok, following your fix it's all good on my side. I included a small fix here: https://github.com/franglais125/dash-to-dock/tree/glossy-luanchers, removing some trailing white spaces.

micheleg:
Ok, added your fixes and reordered the commits. Thanks.

SavageTiger:
Hurray, exciting! I have been running on my branch for weeks without any issues, I merged your latest changes, and as far as I can tell everything looks and feels the same on my machine. Been running it for a couple of hours now, did not detect any issues.

micheleg:
I merged the feature in master [fe6dde50f0f5..5e2fd4c0c].

Thanks again to everybody.

SavageTiger:
@micheleg I am not sure how much memory the iconCache actually uses. However, if someone for example has 205 active items (although admittedly that nobody should have that many active items), it will always recalculate the colorspace of atleast 5 items when they switch workspace. Although since we added resampling that is not a heavy burden anymore. But its something to keep in mind.

micheleg:
Yes, given the limited memory footprint we can put the limit higher (for me 200 was a large number already). I just want to be sure it will never happen to grow to an uncontrolled size. My concern originates from the possibility that  some applications might get assigned inconsistent ids (e.g., I know that sometimes matlab figures do not get properly grouped under the same icon, which i guess implies different app id are used).   I'll put a higher limit then, with the expectation that we will never reach it.

franglais125:
Perfect, I thought of the same thing with regards to "random" ids, but then couldn't come up with an example... Once you merge this patch (or a version of it), I'll try it over some long sessions to track its size (with some `global.log()` calls), and report back.

micheleg:
Ok, merged in master as [b3f86f74be90a]. I put the limit to 1000.

tliron:
@micheleg Thanks for rewriting the HSV/RGB functions! I adapted them for my project and credited you.

franglais125:
@micheleg, reporting back: I know 5 days is not exactly long, but from my logs I haven't seen any "weird" app-id. I tried on purpose to open more apps and perform more actions than I would normally do, and I have the cache size at 25. Having the cache size limit at < 1000 seems to be more than ok.

micheleg:
Good. Let's keep the check in any case, it just costs an if.

PR #513, opened 2017-06-01
Automatic Transparency

franglais125:
Here is a proposal to imitate the work done upstream on the top bar. The idea is that the dash blends in more uniformly, by behaving according to the top bar's new transparency when 'free-floating'.

**Default behavior**
As of Gnome 3.26, the top bar will be only 20% opaque by default. Hence, as a proof-of-concept, I have modified the default behavior for the dash. I have lowered the default opacity from 80% to 20% (as upstream), and the transparency logic is now a bit different. So far we assumed that the dash was opaque, unless the opacity setting was used.

I have now inverted this logic: the dash is 20% opaque, unless the user decides to set the opacity (permanently) to something else.

**Option**
I did it this way for simplicity, but we can of course change how this is done. The setting could be a combo-box with a few options like "Default, Automatic, Permanent" or something like that.

**Regarding the code**
I tried to keep the `_updateSolidStyle()` function as close as possible to upstream, so that it's easier to update if something changes. There are of course some differences stemming from the dash's complexity:
 - The dash is not only on the primary monitor
 - The dash has a border (that can have a different opacity)
 - The dash can have a custom color

Also, I tried to limit the number of operations on this function, as it gets called very often, hence having an impact on CPU time and responsiveness.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -33,7 +33,7 @@ const LauncherAPI = Me.imports.launcherAPI;
 
 const DOCK_DWELL_CHECK_INTERVAL = 100;
 
-const State = {
+var State = {
     HIDDEN:  0,
     SHOWING: 1,
     SHOWN:   2,


franglais125:
@micheleg did you get a chance to try this?

I'm travelling at the moment, but I can look at this again later this week.

micheleg:
Haven't tried it yet. I have been trying Gnome 3.25.x and I have to say I don't seem to particularly enjoy that feature on the panel, however I think it makes sense to try to be consistent. I'll try to have a look at it. Before this I'd like to try to merge the previews on click branch

franglais125:
Sounds good. Let me know if there is any remaing to-do on the previews PR.

franglais125:
Branch updated. It now includes a checkbutton, and the dynamic transparency uses the alpha value selected by the user. Very similar to #62 .

screenshot from 2017-06-14 12-13-52

didrocks:
I did test your branch and quickly gloss over your changes. They look good to me, but I'm not that familiar with the code itself :)

However, I have a suggestion: with this dynamic transparency, it removes any possiblity to have the launcher (especially in intellihide mode) not fully opaque once a window touches it.
Can't it be better to have 2 transparency values? Like, one when no window touches the dock, and another one when a window touches the dock (as you can modify in the upstream css). Or maybe compute one value from the other, what do you think?

didrocks:
The second idea (I'm happy to contribute btw, as I think we'll need to cherry-pick this soon as well for ubuntu-dock) is to only apply this settings in non intellihide mode.
Intellihide would just set the launcher with the selected transparency level, as by definition, no window can really "touch" it and it will then only appear above, or being far from any window.

franglais125:
@didrocks thanks a lot for the feedback! I keep forgetting to test intellihide when I submit PRs, as it's not something I use regularly...

Before trying to implement one of your ideas, let me try to see if I can fix the issue at its root: reacting to intellihide animations in the first place.

I included a commit to address this, I'm not claiming it is fixed, but it is perhaps better. What do you think? [ce76087].

franglais125:
@didrocks Sorry for the change, I actually found away of doing this that is cleaner and more self-contained. Take a look here [792340a0608ac304fe328d0da6146ec10859630a].

Is this good enough?

micheleg:
I'm still a bit confused by how this should behave. I'm personally distracted by having the top panel and the dash background changing transparency, expecially independently, while moving windows around....

If I have undertood the issue and correctly interpreted @didrocks suggestion, I think that:

* The two transparency status should probably be defined by css classes.  The setting to enable/disable the dynamic transparency should be independent from the the opacity settings.
* The opacity settngs should overwrite the opaque state (this is consistent with the origin of such settings, which was meant to ensure the dock is not transparent when using custom themes).  For the transparent state I would use the top panel one in the default theme (or somethig fitting the defualt theme)
* in intellihide mode the transparency could  probably not be dynamic (constantly opaque). However, latest @franglais125 implementation seems to work quite well for me.

franglais125:
> in intellihide mode the transparency could probably not be dynamic (constantly opaque). However, latest @franglais125 implementation seems to work quite well for me.

I was thinking a bit about this too. The problem I saw with having the dock being opaque when in intellihide mode is from the user's perspective: if intellihide is on, toggling dyamic transparency in the GUI would have no effect, which would be weird/seen as a bug. Alternatively, this setting could be greyed out if intellihide is on, but then it is again weird to guess why it is like that. In the end, I was happy with the latest fix I included, as you said.

In any case, I'm waiting for @didrocks 's feedback again, as I'm not sure I fixed what he pointed out. I certainly don't intend to be dismissive of the suggestions.

didrocks:
Thanks to both of you! I tried latest branch and indeed, the Intellihide glitch is now fixed. I'm quite in favor for michele's suggestions, being:

1. Fixed dock:
* Toggle between min and max opacity values. Those values could be defined by css class when toggling dynamic transparency or adding a "min transparency" settings in the UI (widget is only active when dynamic transparency is set to true).
* Opacity settings override the max opacity value set in css (or as told by Michele "opaque state"). Note that you can't use the top panel transparency value as it will be a non linear gradient (currently discussed on #gnome-design). I thus think that the css class approach for the min value is fine as told above (as I can understand you don't want to introduce yet another settings for this min opacity).
2. Intellihide mode:
* the dynamic transparency settings isn't active (and the widget can't be enabled/disabled) as it doesn't have any impact: only the "opaque" state (which can be slightly transparent, based on the existing opacity settings). Indeed, I see a lot of value of having a non fully opaque dock in intellihide mode, but I think it should always stay at the same opacity value.

Does it sum up and align with your understanding? I'm happy to give any help if needed :)

franglais125:
1) **Fixed dock:** (This is w.r.t. css)
To give some context: I initially tried to go the css route (as is done upstream for the top panel). However, I tried to keep in mind that:
- There is a built-in theme (be it Adwaita, Ambiance, etc.)
- The user can manually set the color of the dock in the GUI
- The user can manually set an opacity value in the GUI

The first problem I found was that we need to adjust the property `background-color`, which unfortunately takes as input `rgba`, meaning that I need to specify both color and alpha.
Assuming we specify a css class for dynamic transparency, we would then be forcing the `(r,g,b,a)` combination we chose on all users, whatever theme/color they chose (i.e. dynamic transparency would be incompatible with any non-default theme).

Specifically, I guess this is not a problem for "Ubuntu Dock", as I'm assuming it will be packaged with a specific theme anyway. But this wouldn't be the case for anyone installing d-t-d from e.g.o/git.

In any case, I'm not exactly experienced with css, so please let me know how to work around this. I feel like I'm missing something!

CSS reference: https://developer.gnome.org/gtk3/stable/chap-css-properties.html

2) **Intellihide mode:** I'm fine with turning it off completely. As I don't use it often I don't know what feels more natural. I just thought it didn't look nice when in panel mode (à la Unity) when there are no windows, next to the transparent top panel.

didrocks:
Those are valid points. I'll let Michele answer if the css or the extra settings path is the better with what he has in his mind. I'm starting to think (seeing how the code is structured) that another opacity settings for the start of the transparency transition makes sense.

On intellihide, indeed, it's true that it can look weird when no window touching it or the top panel. So, you may be right by keeping the min and max transparency to accomodate. However, I think that the "max" value (when hidden) shouldn't be opaque, but respecting the same opacity settings (which would be a little bit transparent in ubuntu dock when on top of a window pushing the dock away).

micheleg:
Quick comments:

> The first problem I found was that we need to adjust the property background-color, which unfortunately takes as input rgba, meaning that I need to specify both color and alpha.

I didn't realise that. I don't think we can easily workaround it. I would still offer two css classes  for our "embedded theme" and third parties themes (if they want to support dash-to-dock). 

My idea is 1) we support  the defualt theme and if possible adapt to other themes. 2) We  provide great flexibility for to custom themes to define their own style. 3) The user can still manually tweak some parameter

> I'm starting to think (seeing how the code is structured) that another opacity settings for the start of the transparency transition makes sense.

I still think all Ubuntu-dock needs should be satisfiable with changes at the theme level (with us adding the above mentioned css classes). On the dash-to-dock side, it's not too much of a deal to add another opacity slider if that is the most logical sulution.

For the intellihide case, I'm not sure what's the best visually speaking (I haven't used 3.26 enough to judge). Everything looks a bit unpleasnt to me with a dock touching the top panel (I remember the gradient mockups, but it seems to  still be just transparent as of now). I'm even wondering if the Ubuntu folks have even considered disabling the dynamic transparency...

 As we seems to have a working implementation, and if we a re going to add the second opacity settings, I would go for the current behaviour but with the opacity values corrected (no full opacity).

docking.js
```diff
@@ -350,7 +350,7 @@ const DockedDash = new Lang.Class({
         ]);
 
         this._injectionsHandler = new Utils.InjectionsHandler();
-        this._themeManager = new Theming.ThemeManager(this._settings, this.actor, this.dash);
+        this._themeManager = new Theming.ThemeManager(this._settings, this.actor, this.dash, this);
```
micheleg: Could be just: 
```this._themeManager = new Theming.ThemeManager(this._settings, this);```

**Comment by didrocks** [comment]
>I'm even wondering if the Ubuntu folks have even considered disabling the dynamic transparency...

We did, but for obvious reasons, we don't really want to deviate from upstream's design on that one. I think we can achieve something matching with a very transparent dock when no window is touching them.

> As we seems to have a working implementation, and if we a re going to add the second opacity settings, I would go for the current behaviour but with the opacity values corrected (no full opacity).

Yeah, so basically: current behavior, 2 opacity settings (min/max), intellihide mode respecting as well those 2 opacity settings. That makes sense to me as well. Tell me if you need any help.

**Comment by franglais125** [comment]
Thank you both again for the feedback :)! I was taking some time before tackling the code, as I wanted to make sure I understood what we want.

I'll try to come up with a few patches in the next few days. Further feedback is of course appreciated!

File: docking.js (hunk 1 above)
Comment (franglais125) [reply docking.js]: Thanks. Included this here [e0fc16a2f1dae821d2f91e1695dd27beea0512c0].

**Comment by franglais125** [comment]
Introduced two commits:
- [f6b0ea1] move the styling to `stylesheet.css`. I took the colors for background and border from the upstream default styles.
- [90cba5e] set the maximum opacity to the user-selected value, and the minimum opacity to `0.2` (as the top panel).

If you look at the new style classes added, you'll see that the `opaque` one has a maximum of `0.8` (**suggestions?**). I put this value tentatively, but I want to point out that this is not used anyway, as the maximum is the one selected from the slider.

I still need to rework the GUI+settings to add the second slider.

**To do**:
1. Add setting for minimum opacity
2. Respect the min/max values set in `stylesheet.css` if they were not set by the user.

**Comment by didrocks** [comment]
One last thing while thinking about it and playing with it a little bit more: I think, for full height panel (like ubuntu-dock default and your special settings for it), I think it will make sense to just link the transparency detection with the top panel one. You can consider it as just one UI and not 2 independant part, which could look "flaky".

That could be one option (that I'll toggle in ubuntu dock IMHO), which is "link panel and dock opacity state" or something along those lines. Only the Shell (via the top panel transparency decision) would in that option decides if toggling something transparent or not.
Does it make sense? If you don't feel like doing it, I'm happy providing a PR for this.

**Comment by franglais125** [comment]
I had the same idea actually, while I was setting the minimum to 20%, so yes, I'd find that to be the most consistent look ootb.

I will definitely keep it in mind! I'll try to see how I can implement the various elements (new slider, etc.), while not cluttering the GUI with too many options.

**Comment by didrocks** [comment]
Excellent! :)

Basically, I guess the idea is to check for this settings in `_updateSolidStyle` and check if the "solid" css class is set on the panel element via `has_style_class_name`. But then, there is the question of the race (which _updateSolidStyle() is called first.

So, the easiest and more robust IMHO is to unwire this function when using that settings and rely on, the "style-class" signal set by the panel clutter actor each time the class has changed.

**Comment by didrocks** [comment]
FYI, Shell 3.25.92 disable this dynamic transparency for now: https://bugzilla.gnome.org/show_bug.cgi?id=787446.

There is thus a little bit of time to polish this PR and get the exact desired experience :)

**Comment by micheleg** [comment]
It's still not clear what it's going to happen to the upstream transparency, I have the feeling that that patch is ready just in case that's the final decision.

If we want full syncronization, I think that the best route is to have a single '_updateSolidStyle' function  handling both the panel and dock. As we can't easily disconnect the panel signals, we should replace ```Main.panel._updateSolidStyle``` with an extended one testing on both the dock and panel and then applying wathever logic we like. Connecting to the style-changed signal would work only one way (making the dock opaque when the panel is, but not the opposite)

@franglais125 : beside the UI and settings, is the logic functioning properly now?

**Comment by franglais125** [comment]
> we should replace Main.panel._updateSolidStyle with an extended one 

Good, this was one of the things I was considering, glad to know it would be an acceptable path to take. But I'd wait a bit more to see what happens in the end for 3.26. If they don't ship the feature, we can always reconsider this for 3.28, if it comes then.

> @franglais125 : beside the UI and settings, is the logic functioning properly now?

The logic is functioning properly now, with one caveat: the **minimum** transparency is currently hard-coded to `0.2`, it doesn't respect whatever is included in stylesheet.css. The thing is, I haven't found a reliable way of accessing this information from the stylesheet. Is there a proper way of doing it? A hacky workaround was to apply the `transparent` style to a dummy `St`object, to then retrieve the alpha value from it.

I would like to point out that the **maximum** transparency is set by the opacity slider. The `rgba` included in the `opaque` style is simply to follow the theme's color.

**Comment by micheleg** [comment]
>Good, this was one of the things I was considering, glad to know it would be an acceptable path to take. 

Not too clean but not dirtier then many other parts of the extension :) There's no much choise if we want to change the panel opacity logic.  

> The thing is, I haven't found a reliable way of accessing this information from the stylesheet. Is there a proper way of doing it? A hacky workaround was to apply the transparent style to a dummy Stobject, to then retrieve the alpha value from it.

Can't think of a better way than what you are suggesting. I was a bit naive in thinking it would have been straightforward. If the issue is that we want to specify the opacity but not the color in the css, we could define a custom css property (something like ```-dashtodock-opacity```). The we could apply the class to the dock directly and read the value. But I'm not sure if it works so easily like that, i never tried.

**Comment by franglais125** [comment]
@didrocks It seems Shell 3.26 was tagged today with the top bar transparency enabled. Do you have any other info on this? Just to make sure it will stay like that for now.

**Comment by franglais125** [comment]
Following the limitations of the css approach that we discussed, here is a mock GUI configuration.

It will be easier for me to tackle the code once we agree on the better way to present the options (and what those options should be).

screenshot from 2017-09-12 14-32-16

- **Fixed**: the dock has the same opacity all the time. In the GUI, the opacity slider is active, and the user can set the opacity to any value.
- **Adaptative**: the dock and panel are 'locked', and share the same opaque/transparent state. The dock's maximum opacity is set via a `css` style.
- **Dynamic**: the user can choose the min and max values for opacity, that is triggered according to the windows' positions. In the GUI, we set the `GtkButton` to active, and an advance menu presents slider for min and max.

This is what the "maximum" approach would be, i.e. basically all the options are exposed. I'm not advocating for this model, but this is at least a first sketch of how we could organize things.

Thoughts, suggestions? Options to be dropped?

**Comment by didrocks** [comment]
@franglais125: it seems that it will stay finally. They did some mutter tweaks for the gradient itself, but the design team wasn't fully happy with this. In the end, it's shipped as is (will be probably refined later in 3.27 branch).

I like your sketch and suggestions, very clear! I just have one question in adaptative case, you told that the max opacity is set via a css style. I guess you meant max and min opacity, correct?

Any reason you don't allow in the **adaptative** case as the **dynamic** one? Letting in both case the user changing min and max opacityin via the advanced menu? The difference between the 2 being just a behavioral one, when/how they change from the min to max value?

The rest looks good to me!

**Comment by franglais125** [comment]
> I guess you meant max and min opacity, correct?

Yes!

> Any reason you don't allow in the adaptative case as the dynamic one? Letting in both case the user changing min and max opacityin via the advanced menu? The difference between the 2 being just a **behavioral** one, when/how they change from the min to max value?

Indeed, I don't see any reason why not, under the **behavioral** perspective :). Thanks for the feedback!

**Comment by franglais125** [comment]
Commit [dba6458] is the first part of the work, adding support to extract alpha values from css styles.

To do:
- Add min/max values to override this from the GUI (for both adaptative and dynamic options)
- Implement the 'Adaptative' mode.

**Comment by didrocks** [comment]
Do you need any help around this work?

I want to be able to upload a new ubuntu dock version on Monday, so that we have enough time (even if we will be after UI Freeze for us) to get that tested in the wild before release.

**Comment by franglais125** [comment]
For the Ubuntu dock, I guess the pressing feature is the "adaptative" one, right? I'm assuming since the min/max settings probably won't be exposed.

**Comment by didrocks** [comment]
That's correct. Adaptative and we'll just use some gsettings overrides for min/max

**Comment by franglais125** [comment]
> we'll just use some gsettings overrides for min/max

Is this necessary? You can override it with the css styles. Is there another constraint?

**Comment by didrocks** [comment]
Not necessary, it's just a little bit less instrusive on the long term than a gsettings override (doesn't really patch directly in the source code), when you want to reconcile. But if that's an interim approach, that's fine :)

**Comment by franglais125** [comment]
@didrocks Ok, I think I will be ok to implement this, I wanted to make sure before answering.

Here is a first (sloppy) try: https://github.com/franglais125/dash-to-dock/tree/transparency_adaptative (it's only one commit ahead of this PR so far: [8dc1e759dfff769c9d7c0b279e04db313be1d59e]).

You can toggle the "dynamic" setting, and it will be forcibly "adaptative". This is a proof-of-concept, so you can see how things work, and if anything needs fixing!

Again, feedback appreciated if you get the chance to test it!

**Comment by micheleg** [comment]
Sorry, haven't had too much time to follow this. The mock-up looks good to
me, but haven't thought much about it. I'll try to have a look maybe
tonight. If you are rushing for this feature, maybe we can skip the UI for
the moment (you are not exposing it in Ubuntu anyhow), and also avoid
trying to be flexible with CSS rules (if that's still not sorted out).



On 14 Sep 2017 08:41, "Fran Glais" <notifications@github.com> wrote:

@didrocks <https://github.com/didrocks> Ok, I think I will be ok to
implement this, I wanted to make sure before answering.

Here is a first (sloppy) try: https://github.com/franglais125/dash-to-dock/
tree/transparency_adaptative (it's only one commit ahead of this PR so far:
[8dc1e75
<https://github.com/micheleg/dash-to-dock/commit/8dc1e759dfff769c9d7c0b279e04db313be1d59e>
]).

You can toggle the "dynamic" setting, and it will be forcibly "adaptative".
This is a proof-of-concept, so you can see how things work, and if anything
needs fixing!

Again, feedback appreciated if you get the chance to test it!

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/micheleg/dash-to-dock/pull/513#issuecomment-329400811>,
or mute the thread

**Comment by franglais125** [comment]
@micheleg Thanks!

- The CSS rules are working now.
- [0cf649f5c34545618ffeac3e9912d2c5799de1aa] add adaptive mode
- [cd02e34caab59872ce9c30a4136923107e0e2a77] polish the UI (and add all modes)

This is working quite well in my (limited) testing.

To-do:
1. Min/max values: Need to add the GUI and implement the options
2. Reorganize this PR, as it got messy (I'll clean up the commits)

**Comment by didrocks** [comment]
Nice work! I only had a look and testing your branch, didn't look at the code much, but behavior-wise, it seems consistent to me.

The only little gotcha I noticed is that the fade out timing is slower on the dock than the top panel. This looks weird IMHO. Fade in doesn't have much of a difference. Otherwise, it looks quite good :)

**Comment by franglais125** [comment]
Thanks for testing it!

> the fade out timing is slower on the dock than the top panel

Weird, ai took the timings from upstream (500 and 300ms). Was this tweaked in Ubuntu?

In any case, itp robably makes sensw to get the `transition-duration` from the panel, since we are coordinating the style.

https://git.gnome.org/browse/gnome-shell/commit/?id=447bf55e45b00426ed908b1b1035f472c2466956

**Comment by didrocks** [comment]
> Weird, ai took the timings from upstream (500 and 300ms). Was this tweaked in Ubuntu?
No, this is still the defaults. Maybe a different easing? Or it seems that the dock starts to fade out always before the panel does, in my testing. Start time out of sync, probably?

I agree anyway that getting this property for the panel makes sense.

**Comment by franglais125** [comment]
I just performed a rebase and reorganization. So far at least things are working nicely.

Still to do:
1. Min/max values
2. Get `transition-duration` from top panel.

**Comment by franglais125** [comment]
@didrocks finally found the problem with the transition, it was a `+` missing in the code that made the transition 0 for the dock (hence the panel looked delay). All fixed now (plus we are getting the `transition-duration` from the top panel as well now).

As for the to do list, everything is done now:
- Min/max values: [95416e018b4379f002d6d901ce57649dc514a903]
- Get `transition-duration` from panel [567b66b837e35880f0da0c46331370e1fcceb754]

**Comment by didrocks** [comment]
Excellent work, thanks @franglais125 :)
I'll give a round of testing today and report back later.

Tell me if you want me to look at the code more deeply than a quick review, but I guess both you and Michele are way more acquainted to the code :)

**Comment by franglais125** [comment]
@didrocks Thanks, as usual, feedback is very appreciated :)

@micheleg When you get a chance, two questions:
1. what do you think about the current defaults?
2. are the options reasonable in your opinion?

File: theming.js [hunk 2]
```diff
@@ -291,3 +313,303 @@ var ThemeManager = new Lang.Class({
...
+        });
+        this._trackedWindows.set(metaWindowActor, signalIds);
+    },
+
+    _onWindowActorAdded: function(container, metaWindowActor) {
+        this._addWindowSignals(metaWindowActor);
+    },
+
+    _onWindowActorRemoved: function(container, metaWindowActor) {
+        this._trackedWindows.get(metaWindowActor).forEach(id => {
```
micheleg: We need to check that this._trackedWindows.get() is not undefined.  Why this can happen is not clear to me, but I've seen errors in the log and we do check for it also in the equivalent functoin of intellihide.js.

theming.js
```diff
@@ -17,30 +17,40 @@ const Dash = imports.ui.dash;
...
+const Dock = Me.imports.docking;
 const Utils = Me.imports.utils;
 
+const TransparencyMode = {
+    default:  0,
+    fixed:    1,
+    adaptive: 2,
+    dynamic:  3
+};
+
```
micheleg: Purely style: we have other "ENUMs" with capitalized values.

micheleg: Also, we need some comment defining what those means.

theming.js
```diff
@@ -17,30 +17,40 @@ const Dash = imports.ui.dash;
 const DND = imports.ui.dnd;
 const IconGrid = imports.ui.iconGrid;
 const Main = imports.ui.main;
+const Panel = Main.panel;
```
micheleg: We usually don't do this, but just use Main.panel. I guess it was to keep the code copied from upstram the same. Not a big deal.

franglais125:
@micheleg Thanks a lot for the reviews so far. I'll try to address the comments later today (and anything else you comment on).

I had a rather big issue with this branch, and it's the fact that I was essentially copying '_updateSolidStyle' from the upstream `panel.js`. I don't like this as we would need to port into dash-to-dock every change from upstream, if there are updates. I did it this way as the first thing I tried to do was to see if this 'adaptive' mode would work at all.

Since, I have reworked this in another branch (https://github.com/franglais125/dash-to-dock/tree/transparency_usePanelFunc), where I make use of the upstream function instead (for the 'adaptive' mode) of copying it.

Do you think this is a better approach?

theming.js
```diff
@@ -139,25 +150,26 @@ var ThemeManager = new Lang.Class({
     },
 
     _updateDashColor: function() {
-        if (this._settings.get_boolean('custom-background-color')) {
-            let [backgroundColor, borderColor] = this._getDefaultColors();
+        let [backgroundColor, borderColor] = this._getDefaultColors();
```
micheleg: Are we sure we get the right alpha of the opaque state from getDefaultColors()? Similar to how we remove custom styles in there, maube we should also temporary remove/add the right css classes.

micheleg:
>Since, I have reworked this in another branch (https://github.com/franglais125/dash-to-dock/tree/transparency_usePanelFunc), where I make use of the upstream function instead (for the 'adaptive' mode) of copying it.

At the moment it doesn't seem to make a big difference as the copied code is quite simple. Generally speaking yes, I prefer this second approach, as we don't need to track minor changes upstream, also helping maintaining mutlti-version support.  That said, this feature will be surely reworked upstream in the next cycle, and we will have to readapt it again in any case. We can think about it later maybe.

theming.js
micheleg: A comment here would be very helpful, as the way we tweak the colour is a bit messy :)
1. We get the default dash background (inherithed from the theme)
2. If customized we use the cutsom background color, but maintain the opacityof the default one
3. If the opacity is customized, we also get the custom opacity from the settings
4. The opacity is eventually overwritten by the transparency machinery

micheleg:
@franglais125 Something to bear in mind. I'd like to keep the master branch compatible with recent older GNOME Shell  versions. At the moment we support down to 3.18, but I think it is reasonable to drop the oldest versions. Hoerver,  it is quite important for me to keep the same codebase for the version I run daily on my systems, that is 3.22. I think that this should also include most stable versions of major distributions, so we don't need to backports minor bugfixes. 

This include dealing with the shell code and js syntax. For this branch we need to decide what to do in 3.24 and before. I'm fine in just disabling it for the moment. I was concern about the arrow functions syntax but they seem to work in 3.22 too.

franglais125:
@micheleg Good point. I run Debian 9 too, so whenever I test some new code, the first thing I use is 3.22. This at least ensures we are compatible with Gnome => 3.22. I never test something older though... Perhaps I should, every once in a while.

As for this branch, while it is compatible with Gnome <= 3.24, what happens is that `adaptive` will just work as `dynamic` (I need to add more descriptions to the GUI to warn people). Is this acceptable?

(This is why I test for `!Panel._updateSolidStyle` before enabling the 'adaptive' part.)

If you are ok with this, to-do (feel free to extend it):
1. Address new comments
2. Add more descriptions to the new options in the GUI

micheleg:
Great! For now not crashing or thowing errors in 3.22+ is ok for me, which I plan on support from now on. Let's finalise the bahaviour for 3.26 (which is also what ubuntu dock will support), and than before releasing a new version of dash-to-dock we can clean up the settings, in case hiding the unnecessary ones.

theming.js
franglais125: Addresses here [b52f68f50cb68befc4c258093590ee74ff47192b].

theming.js
franglais125: Thanks, I saw these errors too and tried too hard to find the cause. It's perhaps better to simply check for this, as you say. Addresses here: [64d7a5d6ceb0bf931b303baf5c3282e92f78072c].

theming.js
franglais125: Fair point, we certainly don't do this anywhere, plus there is no need for `Panel` to be global.

Addressed here [be41bd1f7db367e4f1907106842a8da7b06af352]

theming.js
franglais125: > Are we sure we get the right alpha of the opaque state from getDefaultColors()?

This is perhaps something we need to double-check, but just to be clear, this is not something I have changed in this PR. While the changeset for this function looks relatively big, actually all I'm doing is moving `backgroundColor` out of the `if{}` to have it accessible to pass on to `transparency.setColor()`.

In any case, the best way to check this would be with a theme that has some transparency by default. I'll try to get back to you on this.

franglais125: I added some comments here [5e1790435553318c29d6b705a25bb16699d72694]. Not sure if you find this is enough though.

franglais125:
@micheleg I hope I addresses all/most of your comments. In the case of an eventual merge, the new commits can certainly be squashed where appropriate. Let me know if you want me to do this, just o save you some time. I didn't do so to make clear what the new changes are.

didrocks:
Thanks for working on this over this week-end! (And late, from what I see). I'm going to give another round of testing of the new code. The one from Friday was working very well with 3.26, I didn't notice any issue. I'll try to the updated code today.

I guess it's going to be too late for uploading today, I'm trying to aim for tomorrow/wednesday thus (which should be ok given the amount of testing). How should we proceed? I see you are ok maybe to merge this and refine later for earlier G-S release.

Note that I will only cherry-pick the commits + some other fixes from master for 17.10: I can't introduce the Unity backlight supports as we are past Feature Freeze (this is why I didn't rebase the branch recenly) and will get the FF exception for this dynamic opacity only. I'll wait for next cycle thus to do a full rebase, so if you prefer me to cherry-pick directly from this PR branch and not merge it directly, it's fine as well.

didrocks:
Ok, just gave it an advanced round of testing and didn't spot anything weird or incoherent in the dock behavior. Even with intellihide, it's nice to see the transition starting when you push it on side (as it will reappear with max opacity)!

A little gotcha, which doesn't really concern the ubuntu dock case: when you change the advanced settings for min and max opacity in the adaptative or dynamic case like the switch on/off the customize switch or play with the switchers:
1. if you change the max opacity and you are not in the max opacity case with a window next to it, you won't see the change (same with the min opacity in case you have a window next to the launcher).
2. everytime you change a value, a transition is triggered from opaque switcher to max or min opacity (even if you just play with the switch).

Excellent work :)

franglais125:
@didrocks Thanks a lot for the testing!

> 1. if you change the max opacity and you are not in the max opacity case with a window next to it, you won't see the change (same with the min opacity in case you have a window next to the launcher).
> 2. everytime you change a value, a transition is triggered from opaque switcher to max or min opacity (even if you just play with the switch).

I'll try to see if I can work around these two annoyances. Lower priority though!

@micheleg I checked the compatibility of arrow functions. They were apparently introduced in libmozjs = 24, which was included as far back as Jessie. I tested the new code on 3.18 and it worked fine. I actually thought 3.18 was old, but as an example Ubuntu 16.04 is using it. Anyway, I'll try to be more careful with coding style and JS functionality!

micheleg:
Good! Let me see see If I can clean up the branch and merge it in master.

franglais125:
@micheleg thanks! Let me know if you need some help.

theming.js
```diff
@@ -466,28 +467,47 @@ const Transparency = new Lang.Class({
...
         this._updateStyles();
     },
 
-    setAlpha: function(alpha, borderAlpha, originalBorderAlpha) {
-        this._alpha = alpha.toString();
-        borderAlpha = Math.round(borderAlpha*100)/100;
-        this._borderAlpha = borderAlpha.toString();
-        this._originalBorderAlpha = originalBorderAlpha.toString();
-        this._updateStyles();
+    _getAlphas: function() {
```
micheleg: Do we need this only for the styles defined in our stylesheet? Or is is necessary also to support third party themes?

stylesheet.css
```diff
@@ -95,6 +95,20 @@
     background: #2e3436;
 }
 
+/* Only alpha value is used */
+#dashtodockContainer.opaque {
```
micheleg: I know that I suggested to define these styles here, but I don't see the point of this any more.

theming.js
franglais125: We need this for both:

- to retrieve our own `opaque` and `transparent` styles
- if a (3rd party) theme author includes his own `opaque` and `transparent` styles

Does it answer your question?

theming.js
```diff
@@ -360,7 +360,8 @@ const Transparency = new Lang.Class({
 
         // Window signals
         global.get_window_actors().forEach(function(win) {
-            this._addWindowSignals(win);
+            if (win.get_meta_window().get_wm_class() !== 'Gnome-shell')
```
micheleg: This works upstream because they never disconnect the signal? How do you reproduce the bug?

franglais125: Exactly, it's a bit annoying, but sometimes I can trigger this. I tried to pinpoint the root cause of this, but I was unable to. There are some actors that don't get properly connected, and then fail disconnection. As you guessed, the `panel` signals are never disconnected and hence no warning appears.

stylesheet.css
franglais125: Instead you would simply set min/max alpha values in the settings for instance?

I think the idea was to better integrate to specific themes, in the case a theme author wants to set a specific opacity level for the panel and the dock (both opaque and transparent).

I don't feel strongly about this though.

theming.js
franglais125: The css classes might come with specific colors, and we might be overriding this already.

This was the whole problem since the beginning: is there a property e.g. `alpha: 0.8` that we can use? If not, we need the full `rgba()`.

stylesheet.css
micheleg: This is connected with the idea of appplying classes directly. The issue originates from my idea of defining these classes in our theme in the first place. If we don't do it (or better we do it only for the "our custom theme" (dashtodock class), we do not overwrite anymore the style and we can just apply the classes. I'm mainly trying to avoid this weird workaround of adding an actor to retrieve the style

theming.js
micheleg: If the color comes from the theme it's fine, that's what the auhor wants. Doesn't it makes sense? I can try to preapare a patch and see if it work, as I might be missing something :)

theming.js
micheleg: Ok: we have the same problem in the intellihide class https://github.com/micheleg/dash-to-dock/blob/master/intellihide.js#L107-L109

theming.js
franglais125: What I meant is, if the user sets a specific color in the settings, this color will be overridden by the theme.

Normally, even if the theme proposes a default color, the GUI lets the user set whatever color he wants.

franglais125: In any case, as you suggested elsewhere, we could just drop the whole `opaque`/`transparent` styles, and remove this machinery. I'm fine with that!

theming.js
franglais125: I don't use intellihide, so perhaps this is why I didn'd encounter the error before. Did you see it anywhere else? Like so:

```
gnome-shell[1313]: /build/glib2.0-B1uXKV/glib2.0-2.50.3/./gobject/gsignal.c:2641: instance '0x55fc2d1d47f0' has no handler with id '52060'
```

where `0x55fc2d1d47f0` is a window instance (I did check this in the past).

micheleg:
@didrocks: I have prepared a slightly cleaned-up branch in the development branch (https://github.com/micheleg/dash-to-dock/commits/development). It is substantially the same of this PR, beside very minor cosmetic changes and history cleanup. This is what I'm going to be merge in master, although I still want to improve the behaviour on dash-to-dock  side. I think the best strategy for you is to cherry pick from there.

stylesheet.css
micheleg: This https://github.com/micheleg/dash-to-dock/tree/experimental_tweak_transparency_behaviour is wip towards what I was trying to describe:

1. With custom theme enabled, integrate as well as possible with the default theme
   * Enable adaptive transparency if the dock is in panel mode and vertically placed, matching the background of the panel. 
   * Keep the dock opaque in the other cases (personal preference)
2. For custom themes, in the DEFAULT behavior, the style provided by the theme will be used. Using the .solid class, themes writers can directly support dash-to-dock.
3. If this still does not match, the user can play with the advanced settings.

franglais125: Ah, I see! Indeed, I was not understanding what you meant. I'll give it some testing and report back. So far the only problem I found is that some styles are completely black (I guess due to my using `(0,0,0,alpha)` for rgba in the stylesheet).

I love the changes so far, I didn't know you wanted to change the default behavior as well :).

franglais125: @micheleg I included a few fixes here: https://github.com/franglais125/dash-to-dock/tree/experimental_tweak_transparency_behaviour

didrocks:
Here we go! I just based the ubuntu-dock branch, cherry-picking the needed commits. Thanks a lot @micheleg and @franglais125  for working on this. After the currrent finale round of testing, everything looks good to me.
I'm waiting for our Feature Freeze exception to be acked and will upload that to ubuntu itself.

I'm planning as well blogging about your excellent work there :) Do not hesitate if you need anything!

franglais125:
@didrocks thanks a lot for the feedback and ideas! Good luck with the FF exception!

micheleg:
So I've merged this in master as of [225b51dd71a].

franglais125:
@micheleg Thanks a lot! Let me know if you find remaining issues or things to polish. One quickly comes to mind: hide the "Adaptive" option on Gnome < 3.26, or at least offer an explanation on the UI.

micheleg:
That's exactly what is missing on our side, before reconsidering our default behaviour. I would possibly hide the adaptive case. For adding explanationas, I don't know where to put it. An idea would be to update the "subtitle" of the option, the one that now "Tune the dash backgrond opacity".

franglais125:
@micheleg I added two commits to see how this would work. In general, I'm not really satisfied with the result. I needed to work around a few things which I didn't expect:

- [c2b7b24] I needed to add a compare function. I'd have liked to simply check for `Main._panel._updateSolidStyle`, but importing `imports.ui.main` introduces a bug we have hit before (missing `Meta` library, etc.).
- [93826df] Changing the label can change the space it takes up in the GUI, hence I manually set the height request. Also, the description itself might need some work.

franglais125:
Added one commit to remove JS warning.

micheleg:
Last mentioned commit was merged as ae7b86ec.

franglais125:
Finally closing this PR. Thanks everyone for the effort!

didrocks:
Thanks again for your work! :)

joca-bt:
Hi all, sorry to barge in. I can't find the options mocked here in the version available at https://extensions.gnome.org/extension/307/dash-to-dock/. How can I get the adaptative behaviour?

micheleg:
v62 is still awaiting review (#679). You have to  manually install the latest release if you want access to those features.

PR #518, opened 2017-06-09
Number Overlay: set color.

franglais125:
With some themes, the font is transparent and the overlay cannot be seen, reported here: https://github.com/jderose9/dash-to-panel/issues/182.

We can put the style in stylesheet.css, or in the code directly. I thought that have it in stylesheet.css was enough for theming, but since it's not, I guess it makes no difference. Let me know if you'd rather have it in stylesheet.css.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -624,7 +624,7 @@ const MyAppIcon = new Lang.Class({
             x_align: St.Align.START, y_align: St.Align.START,
             x_expand: true, y_expand: true
         });
-        this._numberOverlayStyle = 'background-color: rgba(0,0,0,0.8);'
+        this._numberOverlayLabel.add_style_class_name('number-overlay');
         this._numberOverlayOrder = -1;
         this._numberOverlayBin.hide();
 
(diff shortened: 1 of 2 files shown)


micheleg:
What does it means that putting it in "stylesheet.css" is it not enough for theming?

micheleg:
What does it means that putting it in "stylesheet.css" is it not enough for theming? I would move it to the stylesheet, it makes more sense there.

franglais125:
I just moved it to stylesheet.css.

I meant that if the style is taken from stylesheet.css, it would override whatever global theme is being used. But my understanding/experience of this is limited, I actually never use other themes than the default.

micheleg:
Ok. I was thinking in terms of lettings themes author customizing the style. With the class in the stylesheet this should be possible. The default white on black is a good fallback in any case In my opinion.

franglais125:
Sounds good to me!

micheleg:
Merged [91fb537b7]. Thanks.

PR #519, opened 2017-06-10
Change minimize or overview with selected app windows only

ggcode:
Hi,

i have seen in your code, that you wanted the overview to only show the windows of the app icon clicked in the dash when in "minimize or overview" mode.
As i wanted to try gnome extension development anyway for some time, i found it to be a good practice.
I changed it to the settings i had in Unity for this for some time, so now:
* If the app is not focused (other app or minmized) -> dash click-> just show app window.
* If the app is focused -> dash click -> show overview with selected app windows only.

For me it works quite nice.
Hope, you like it.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -47,7 +47,8 @@ const clickAction = {
     MINIMIZE_OR_OVERVIEW_APP: 5,
     PREVIEWS: 6,
     MINIMIZE_OR_PREVIEWS: 7,
-    QUIT: 8
+    FOCUS_OR_PREVIEWS: 8,
+    QUIT: 9,
 };
 
 const scrollAction = {
@@ -471,6 +472,17 @@ var MyAppIcon = new Lang.Class({
                 }
                 else
                     this.app.activate();
+                break;
+
+            case clickAction.FOCUS_OR_PREVIEWS:
+                if (this.app == focusedApp &&
+                    (windows.length > 1 || modifiers || button != 1)) {
+                    this._windowPreviews();
+                } else {
+                    // Activate the first window
+                    let w = windows[0];
+                    Main.activateWindow(w);
+                }
                 break;
 
             case clickAction.LAUNCH:
--- a/appIcons.js
+++ b/appIcons.js
@@ -1,6 +1,7 @@
 // -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
 
 const Clutter = imports.gi.Clutter;
+const GdkPixbuf = imports.gi.GdkPixbuf
 const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const Gtk = imports.gi.Gtk;
@@ -31,6 +32,7 @@ const Me = imports.misc.extensionUtils.getCurrentExtension();
 const Utils = Me.imports.utils;
 const WindowPreview = Me.imports.windowPreview;
 const AppExposeOverview = Me.imports.appExposeOverview;
+const AppIconIndicators = Me.imports.appIconIndicators;
 
 let tracker = Shell.WindowTracker.get_default();
 
@@ -43,7 +45,8 @@ const clickAction = {
     CYCLE_WINDOWS: 3,
     MINIMIZE_OR_OVERVIEW: 4,
     PREVIEWS: 5,
-    QUIT: 6
+    MINIMIZE_OR_PREVIEWS: 6,
+    QUIT: 7
 };
 
 const scrollAction = {
@@ -63,7 +66,7 @@ let recentlyClickedAppMonitor = -1;
  *
  * - Pass settings to the constructor and bind settings changes
  * - Apply a css class based on the number of windows of each application (#N);
- * - Draw a dot for each window of the application based on the default "dot" style which is hidden (#N);
+ * - Customized indicators for running applications in place of the default "dot" style which is hidden (#N);
  *   a class of the form "running#N" is applied to the AppWellIcon actor.
  *   like the original .running one.
  * - Add a .focused style to the focused app
@@ -71,20 +74,24 @@ let recentlyClickedAppMonitor = -1;
  * - Update minimization animation target
  * - Update menu if open on windows change
  */
-const MyAppIcon = new Lang.Class({
+var MyAppIcon = new Lang.Class({
     Name: 'DashToDock.AppIcon',
     Extends: AppDisplay.AppIcon,
 
     // settings are required inside.
-    _init: function(settings, app, monitorIndex, iconParams) {
+    _init: function(settings, remoteModel, app, monitorIndex, iconParams) {
         // a prefix is required to avoid conflicting with the parent class variable
         this._dtdSettings = settings;
         this.monitorIndex = monitorIndex;
         this._signalsHandler = new Utils.GlobalSignalsHandler();
         this._nWindows = 0;
         this.appExposeOverview = new AppExposeOverview.AppExposeOverview();
+        this.remoteModel = remoteModel;
+        this._indicator = null;
 
         this.parent(app, iconParams);
+
+        this._updateIndicatorStyle();
 
         // Monitor windows-changes instead of app state.
         // Keep using the same Id and function callback (that is extended)
@@ -93,30 +100,42 @@ const MyAppIcon = new Lang.Class({
             this._stateChangedId = 0;
         }
 
-        this._stateChangedId = this.app.connect('windows-changed',
+        this._windowsChangedId = this.app.connect('windows-changed',
                                                 Lang.bind(this,
                                                           this.onWindowsChanged));
         this._focusAppChangeId = tracker.connect('notify::focus-app',
                                                  Lang.bind(this,
                                                            this._onFocusAppChanged));
-        this._dots = null;
+
+        // In Wayland sessions, this signal is needed to track the state of windows dragged
+        // from one monitor to another. As this is triggered quite often (whenever a new winow
+        // of any application opened or moved to a different desktop),
+        // we restrict this signal to  the case when 'isolate-monitors' is true,
+        // and if there are at least 2 monitors.
+        if (this._dtdSettings.get_boolean('isolate-monitors') &&
+            Main.layoutManager.monitors.length > 1) {
+            this._signalsHandler.removeWithLabel('isolate-monitors');
+            this._signalsHandler.addWithLabel('isolate-monitors', [
+                global.screen,
+                'window-entered-monitor',
+                Lang.bind(this, this._onWindowEntered)
+            ]);
+        }
+
+        this._progressOverlayArea = null;
+        this._progress = 0;
 
         let keys = ['apply-custom-theme',
-                   'custom-theme-running-dots',
-                   'custom-theme-customize-running-dots',
-                   'custom-theme-running-dots-color',
-                   'custom-theme-running-dots-border-color',
-                   'custom-theme-running-dots-border-width'];
+                   'running-indicator-style',
+                    ];
 
         keys.forEach(function(key) {
             this._signalsHandler.add([
                 this._dtdSettings,
                 'changed::' + key,
-                Lang.bind(this, this._toggleDots)
+                Lang.bind(this, this._updateIndicatorStyle)
             ]);
         }, this);
-
-        this._toggleDots();
 
         this._dtdSettings.connect('changed::scroll-action', Lang.bind(this, function() {
             this._optionalScrollCycleWindows();
@@ -139,7 +158,11 @@ const MyAppIcon = new Lang.Class({
             this._menu.close(false);
 
         // Disconect global signals
-        // stateChangedId is already handled by parent)
+
+        if (this._windowsChangedId > 0)
+            this.app.disconnect(this._windowsChangedId);
+        this._windowsChangedId = 0;
+
         if (this._focusAppChangeId > 0) {
             tracker.disconnect(this._focusAppChangeId);
             this._focusAppChangeId = 0;
@@ -153,6 +176,23 @@ const MyAppIcon = new Lang.Class({
         this.appExposeOverview.disconnect();
     },
 
+    // TOOD Rename this function
+    _updateIndicatorStyle: function() {
+
+        if (this._indicator !== null) {
+            this._indicator.destroy();
+            this._indicator = null;
+        }
+        this._indicator = new AppIconIndicators.AppIconIndicator(this, this._dtdSettings);
+        this._indicator.update();
+    },
+
+    _onWindowEntered: function(metaScreen, monitorIndex, metaWin) {
+        let app = Shell.WindowTracker.get_default().get_window_app(metaWin);
+        if (app && app.get_id() == this.app.get_id())
+            this.onWindowsChanged();
+    },
+
     _optionalScrollCycleWindows: function() {
         if (this._scrollEventHandler) {
             this.actor.disconnect(this._scrollEventHandler);
@@ -161,7 +201,6 @@ const MyAppIcon = new Lang.Class({
 
         let isEnabled = this._dtdSettings.get_enum('scroll-action') === scrollAction.CYCLE_WINDOWS;
         if (!isEnabled) return;
-
         this._scrollEventHandler = this.actor.connect('scroll-event', Lang.bind(this,
                                                           this.onScrollEvent));
     },
@@ -210,7 +249,7 @@ const MyAppIcon = new Lang.Class({
         if (this._menu && this._menu.isOpen)
             this._menu.update();
 
-        this._updateRunningStyle();
+        this._indicator.update();
         this.updateIconGeometry();
     },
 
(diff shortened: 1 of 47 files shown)


micheleg:
Very nice! Good to have a working prototype for this  feature. Beside the implementation and possible technical limitations, would you mind describing what's the ideal behaviour you have in mind? In particular regarding:
1 showing application windows on the different workspace or all together
2 showing not (related to point 1.) the workspace thumbnails. 
3 Hiding other applications windows from the workspace thumbnails too
4. What happen when you click on another application from the overview: go to the desktop or switch to that application windows
5. Provide a button to revert to the global overview (without having to exit/reenter)
6. Should this "application overview" be a clearly different status, that cannot be easily mistaken for the standard overview.

Just thoughts. Also curious to know about other implementations (Unity,MacOS/Windows?)

ggcode:
Thank you very much for your fast feedback and those good questions, which gave me a lot to think about.

To be honest, I dont use workspaces that much and just overlooked them in this context. Taken them into account changes the whole thing. In Unity Tweak Tool i always had the minimze option enabled: http://www.omgubuntu.co.uk/2014/03/minimize-click-launcher-option-ubuntu-14-04. 

So this was kind of my inspiration and i was glad to see it got started in version 59 ( 'TODO: only show current app windows when gnome shell API will allow it'). So, you are right, its more of a prototype for this functionality.

I now made the implementation to show only the selected application windows in the thumbnails view  and hide the search bar. Also at least moved the code into its own function.

ggcode:
Now with the workspaces in mind, i think there are basically two options here:

1. -> show application windows overview - with thumbnails, remove other apps from thumbnails +
 hide search (done with new commit), hide workspaces without application windows, mouse scroll on app icon also switches workspaces.
2. -> show application windows overview - remove thumbnails, no workspace switching and show all application windows from all workspaces (that might get confusing if you select one window and that selection switches back to the desktop on a different workspace without having noticed ..).

Could be changed via dash-to-dock settings. I would prefer to make the first option the default. While playing around with your suggestions, i also found the 'mouse scroll through application windows' option to be a very similar idea. So this can be seen as a more visual version/alternative to that.

For the questions:
1. Currently it will show the application windows on their workspaces in the overview. So its kind of an application specific overview.
2. This would be option two from above.
3. Done.
4. Switch to that applications window (same behaviour as the 'original' overview).
5. Would be good if the 'top left activity corner'-action could be made aware if in app overview mode and then switch to the original overview. Dont know right now, how difficult that might get..
6. Maybe, but as this is my first time having a look at Gnome extensions and Gnome code in general, it felt more natural to me to just replace one tiny little function in a certain state for the overview (and one of gnomes tutorials also suggested it). But you are right, might be a good idea if the code quantity increases even more. Im a big fan of refactorings like these.

Currently i dont know when i have the time next but I will definitely give it a try.

Really thank you for your feedback so far.

appIcons.js
```diff
@@ -1203,3 +1183,90 @@ function itemShowLabel()  {
...
+    _init: function() {
+        this.isInAppOverview = false;
+        this.originalTriggerSearchFunction = ViewSelector.ViewSelector.prototype._shouldTriggerSearch;
+        this.originalOverviewFunction = Workspace.Workspace.prototype._isOverviewWindow;
+        this.originalThumbnailFunction = WorkspaceThumbnail.WorkspaceThumbnail.prototype._isOverviewWindow;
+        this.hidingId = Main.overview.connect('hidden', Lang.bind(this, this._onOverviewHidden));
+    },
+
+    dissconect: function() {
+        Main.overview.dissconect(this.hidingId);
```
franglais125: I'm getting errors due to this line.

`dissconect` is misspelled in a coupe of places!

It should be `disconnect` instead :)

ggcode: You are absolutely right. Always looked strange to me to ;) Fixed it with new commit.  Thank you!
Where did you see the errors? Im currently alway using 
`tail -f /var/log/syslog | grep Gjs-Message`
Is there a better way?

franglais125: On Debian Stretch, I use: `sudo journalctl /usr/bin/gnome-shell -f`.

Thanks a lot for the quick fix!

ggcode: Thank you for the hint and the new command in my toolbox. Just tried it, looks good.

ggcode:
@micheleg

I was testing the application overview mode and while closing windows in this view, i found it a little bit odd, that if you close the last window of the current application you are stuck in the overview without any windows at all.

So i tried how it 'feels', switching back to the desktop after closing the last window. This felt kind of weird to. Because, if you are in an application specific view of all windows and close the last one.. you are suddendly switched to the desktop without any windows of that context. So i tried to switch back to the desktop with one window left...

It felt a little bit weird to me to, but less than before and as the behaivor is currently not set to stone and still experimental i just left it as it is and commited 'at least something' ;)

With only one workspace the whole thing works quite nice now. With multiple workspaces i think its also quite ok. (I never noticed this before, but gnome has some strange behaivor with closing windows in the standard overview and workspaces to) This whole workspace stuff made it way more complex regarding the user experience...

But i get your point.. so what do you think the behaivor should be, if you close the last window in the application specific overview?

One question though. As this is also my first pull request on github: Is it ok, to work on this and leave it as pull request and test stuff like this.. or should we close this and just raise a new one after having a version im ok with? And btw.. Im ok with the current version. At least its closer to the original intent of this, than v59..

As always thank you for your thoughtfull feedback and testing this thing.
I really appreciate it.

micheleg:
I don't know yet what's the behaviour that we want :) It's good to experiment because before trying things up it's difficult to realise what makes sense. The best is keeping the experiments around. If there are unclear things like this I would say the best is to get the simple behaviour in a state that can be merged, and then think about refining it. I haven't look in details to your code but I understand the logic. it's a bit of a "dirty" approach but it's also the quickest way to  achieve this and I'm happy to go on with this route. I will need to look to your code more in details, but it seems quite uncoupled from the rest, which is very good. 

Regarding:
>Would be good if the 'top left activity corner'-action could be made aware if in app overview mode and then switch to the original overview. Dont know right now, how difficult that might get..

In principle you could apply the same principle of intercepting the click and then retrigger the overview update (I'm not sure which function etc etc but I'm confident is feasible... what I don't know is how it would look visually)/

micheleg:
Regarding this:

>4. What happen when you click on another application from the overview: go to the desktop or switch to that application windows

>Switch to that applications window (same behaviour as the 'original' overview).

It is a bit questionable if this should be the behaviour when this action is associated with an augmented click (middle-click or click with a key modifier), for motivations similar to  this [5003841bd005d71737fc192afe7fbe8a76512f17]. I don't know if it's easy to change the overview without exiting and reentering it though.

ggcode:
Just rebased everything to your master branch and into one commit.
Regarding your last comment: You mean, you want e.g. the preview windows to be visible on middle click in app overview mode compared to just switch back to the desktop and open a window of the application (middle) clicked?

micheleg:
I was imagining this situation: let's say I have mapped the minimize or overview action to the middle-click 
1.) enter the overview looking for my window
2.) Ops, too many windows
3.) middle-click usually lead me to the restricted overview (it's clearly intentional that I want to trigger that action), therefore it would makes sense to do so as well

It doesn't make sense instead for the default click. It was just a thought. I don't know what we want yet!

ggcode:
Well maybe im 'to used' using the (app) minimize or overview now, but if want to select a specific window of the app in the first place i would not open the normal overview first, as at least i have always way to many windows there ;) And thats the whole point of the action, to narrow that down.

Moved the code into its own file now and added the switch from app overview to normal overview via hotcorner.
Found one odd behaivor though, when clicking the "Applications Icon" (Top Icon in Dash), the search stays hidden..
Maybe i also rename this to something like ApplicationExpose or WindowSpread to distinguish it from that application view of the standard gnome shell..
(as it is called here: http://www.omgubuntu.co.uk/2017/06/dash-to-dock-unity-features-coming)

franglais125:
I think what Michele was trying to convey is not that a user should open the overview to use this feature, but rather: what should we do if the overview is *already* opened?

Suppose (which is true in my case) a user has lots of windows, but spread across many workspaces. In some workspaces I only have a few windows, so using the overview makes sense. Now imagine:

1) I move to another workspace where I have a lot of windows, and I open the overview (mistakenly or not).
2) So I decide to use this feature (mapped to the middle-click).

What should happen now? Since the user is issuing a **middle-click**, it's pretty clear that s/he intentionally wants to trigger the ApplicationExpose (to narrow down the relevant windows), and not activate the app, which is what happens here: https://github.com/micheleg/dash-to-dock/pull/519/files#diff-f2c18b3123458b0b36d6cad2215ed52fR453 (link to specific line).

Instead, we could checks what kind of click it is. If it's not the default click, then we perform the specific feature, instead of activating the app.

Something like here: https://github.com/micheleg/dash-to-dock/pull/519/files#diff-f2c18b3123458b0b36d6cad2215ed52fR447

```
if (this.app == focusedApp ||
   (windows.length > 1 && (modifiers || button !== 1))) {
```

Cheers, love this feature btw! To be honest, I tried a few times to get it working and was frustrated more than once, so congrats.

ggcode:
Thank you and im happy you like it that much. This keeps me going..

For the middle-click: This is how i understood it to. Maybe i revert just showing the application specific Windows in the thumbnails box in ApplicationExpose (for now lets call the narrow down action like this) or just for the thumbnail previews on workspaces without app specific windows at least as a first step to better handle workspaces.
In additon to that (and now that people actually using workspaces telling me to ;) ) i will give switching back from overview to ApplicationExpose as described by you and Michele and you another try this weeken. Just need to be a little carefull here, as switching the 'overviews' without hiding them in between changes a lot of state in the event handlers, as i ve seen while implementing the new hot corner-action.

As always, thank you for your precious feedback

ggcode:
Hmn.. now i really get the point. You seem to testing this preferably as a agumented click option (middle button or something to do this). In contrast i am testing this most of the time as primary click option with the first click showing the last window on dash and the second click showing the app expose (as im used to).
I'll try to combine both worlds from an user expirence point of view in the next days. 
Might be a lot of try and error though, as there are so many combinations of settings regarding this in dash to dock (which is good)..

franglais125:
Unless I'm missing something, the logic should simply be (greatly simplified):
```
if (not overview) {
    AppExpose();
}
else {
    if (default click && no modifiers)
        app.activate();
    else
        AppExpose();
}
```

I'm not saying that this is what we should do though, as Michele said, it often takes some testing to see what is really better.
@micheleg in the same way, it'd probably make sense to show the window previews even on overview if mapped to middle-click/modifiers.

ggcode:
For the window previews: I agree. Was exactly what i was thinking, beeing asked to provide the AppExpose on middle-click.. 
For the other stuff: Im currently in trouble switching between overview types/states without calling Main.overview.hide() in between. e.g. if i switch from appExpose to overview right now, it works fine (had trouble with resetting the search providers before) apart from: the thumbnails-box is 'hidden somewhere in event translation'. Try to figure that out currently and reshow the search box if going to standard gnome application view.. 
Will try to test it more with beeing an secondary click option though.

Thanks for your feedback. Really have fun trying to get this in a usable state for everybody.

franglais125:
IMO it is very usable already. Having the alternative click options is just a nice plus!

ggcode:
I think so to. But you know, its always the last 10% or so, that gives you headaches ;)

micheleg:
> Might be a lot of try and error though, as there are so many combinations of settings regarding this in dash to dock (which is good)..

I know...

Thanks for keeping working on this feature. I haven't had time to test and look at your latest changes. If we get the ability to move between different overview states right, then it won't be difficult to test different behaviours with the activating clicks.

> @micheleg in the same way, it'd probably make sense to show the window previews even on overview if mapped to middle-click/modifiers

Meh... do we want windows previews inside the overview showing the same windows previews? That was my first thought. Then we already do this for the previews in the default popupmenu. I don't know :)

ggcode:
Hi, just wanted to give a short update on this...

I have some kind of prototype for the alternative click function here https://github.com/ggcode/dash-to-dock/tree/minimize_or_overview_appExpose_experimental2. This 'feels' better for the alternative click function (always shows overview on alternative click for apps with more then one window, else standard behavoir).

BUT: state handling gets really difficult. As its not just switching workspace on/off in between like now any more. But also switching between pages of the overview and finding the places certain 'state' cleanups happen. Try to switch a litte bit between different apps in AppExpose and/or classic overview via left top hot corner. Then try search or application page... oh man :)

franglais125:
Just a small comment: `appExposeOverview.js` is missing in the `Makefile`.

ggcode:
@franglais125 Ok like this? (Makefile)

franglais125:
Yes, thank you!

ggcode:
Hey, sorry for not giving an update on this lately. But i use it on a day to day basis without thinking about it to much. As a left click option it works quite nice. Not sure, what it feels like as secondary option as i seldomly try this. But its an enhancement to the original intention of this stuff for sure. Im a little bit worried about older/different versions of gnome though. Dont have a test setup for this here. So what do you think about merging it anyway?

franglais125:
@ggcode fwiw, I've been using your code for weeks without any problems. Cheers!

micheleg:
Suggestion from @gunwald in #559:
> To make it transparent in the overview that not all windows of one workspace, but all windows of a certain kind of all workspaces are shown, the workspaces which contains an instance of this certain application could be marked. To make clear, which application is shown, the icon in the dock should be also marked.

ggcode:
Regarding the suggestion of @gunwald. I liked his suggestion to mark workspaces that contain non 'appExpose' Applications and his mockup for it. I never liked the fact in the current prototype, that workspaces without app windows in the workspace thumbnails are just shown 'empty'. Maybe grey out the thumbnail and ignore/jump to next with app windows while switching workspaces is a good solution.
Showing the app icons is a nice little reminder that you are in an overview of an sepcific app to, though not critical as a user should know he has triggering the special action for that behaivor already.
Apart from that, im thinking about moving this thing from overriding workspace methods into a complete reimplementation with its own code for 'painting' all of it and hooking/overriding as less gnome3 events/methods as possible and also abstracting the gnome version support for that thing. like @dmo60 did in a very beautiful way with his coverflow tab extension.
Currently dont know, if i will have the time to do..

micheleg:
> Apart from that, im thinking about moving this thing from overriding workspace methods into a complete reimplementation with its own code for 'painting' [...]

That would be great indeed, as the current solution  is a bit hacky (even though the implementation is quite self-contained so if it works reliably we could go with it to start ). However, I don't know what's the effort required. It's good to know there's some code to take inspiration from.

marcelopedras:
Is this resource ready? 
Ubuntu Unity has this behavior. It would be nice to have this option to maintain this behavior.

Congratulations for the initiative.

micheleg:
The feature is stil in an hacky status. It hasn't been finalised yet.

jmaris:
Any progress on this ?

micheleg:
Nope, unfortunately no work has been don on this.

jasjuang:
I just updated to Ubuntu 18.04 and this is the main feature I missed from Ubuntu 16.04. This would be such a great feature to have. Hope someone will be able to complete this. I will be happy to do the testing.

ahmadhasan2k8:
It would be a great option to have back. The new way is hard to use.

ggcode:
@micheleg  Hi again,
after to much time i tried the code yesterday again on Gnome Shell 3.28, merged your master, resolved two small conflicts. Also added appExpose as additional option in settings, so its easier to compare with the old minimize or (full) overview. And people can always switch back if they dont like it.
Regarding

> even though the implementation is quite self-contained so if it works reliably we could go with it to start

Didnt read/see that last time. Worked for me for half a year on 3.26 without problems. Did a short test yesterday on 3.28 comparing "appExpose" with "minimize or overview". Memory didnt rise and CPU was even lower than normal overview while extensively open and closing it. Also tried it in a VM against Gnome 3.2 (oldest shell version supported as mentionend on extensions.gnome), worked just fine. franglais125 wrote in a comment it also worked for him.

(now that users have it as 'just another additonal option') Maybe just try it out and decide if you like it  ;)

(btw.: 'paint' stuff myself instead of hooking in.. Tried that, but gave up after 2 days or so. Because essentially i was recreating/copying all js-files needed for overview, thumbnails, viewSelector etc. And in the end also had to hook into the original workspace code to disable stuff for state handling.. So, wont try that any further)

paroj:
> Found a way that also works nice with maximized windows not in expose.
Doesnt need screen width/height for calculation.

the calculations do not correctly consider the screen: non-app windows always fade in from the left screen in a multi-monitor setup, instead of the origin screen.

ggcode:
@paroj Your right, didnt test it after implemeting with multiple screens. will fix, if i find time. Thanks for your review.👍

paroj:
@micheleg I have been using this on several machines for a while now without negative effects. Is there anything holding this back from being merged?

micheleg:
@paroj I'm "holding this back", In the sense that I haven't had time to review this (and I forgot,so thanks for pingeng me on this). 

I had a quick look at the code, and I see that it is not very invasive (most of the implementation is self contained), and with the option not selected nothing should be running. A couple of events are connected:
```
this.hiddenId = Main.overview.connect('hidden', Lang.bind(this, this._onOverviewHidden));
this.pageChangeId = Main.overview.viewSelector.connect('page-changed', Lang.bind(this, this._onPageChanged));
```
but they do nothing with the option enabled.

I haven't look in details to the implentation which I expect to be  a bit hacky, but that's probably the best that can be done in an extension without rewriting substantial parts of the upstream code :) 

In the interest of not loosing this constribution I'm therefore very temped to merge as it is as it seems to be working well. Anybody wants to help review @ggcode's code?

ggcode:
@paroj  just approve the code review once again?

paroj:
:+1: but note that I mainly judge this by that it "just works" and the implementation indeed looks minimal. I cannot judge the API usage.

ggcode:
OK ;)

paroj:
@micheleg is there anything else needed here?

ggcode:
Sorry for the git push 'attack'. I just didnt recognize, when i tried to merge the upstream master branch last time, that it is only compatible with gnome 3.32...
The PR at least works from gnome 3.2 up to 3.30 again ;)
So changed the PR target to gnome-3.30 for now.
I am not using 3.32,  nevertheless, im gonna have a look at the error logs for that in the near future ..

ggcode:
And the "trouble" i had with 3.32 was just something like this:
https://github.com/micheleg/dash-to-dock/issues/889
So.. nothing concerning the functions changed for the PR, but missing functions from 3.32 ..
For the last three gnome updates, i tested the stuff with all previous versions in a container/chroot/vm. This time i just merged upstream, because it worked every time. Today i tested it. Well.. my guess is, it will work on a 3.32 installation.. but it didnt work out with master of dtd on a 3.28 system. Obviously ;)

ggcode:
Ok.. just in case.. Checked out master, applied the changes, tested it on fedora 30 with gnome 3.32. Still works as expected. Branch for dash-to-dock master | gnome-3.32: 
https://github.com/ggcode/dash-to-dock/tree/minimize_or_overview_current_app_windows-3.32

Changed files are still the same..
https://github.com/micheleg/dash-to-dock/compare/master...ggcode:minimize_or_overview_current_app_windows-3.32

pyroxar:
@ggcode do you have any screenshot of this feature?
Did you can see this?
https://github.com/linuxmint/cinnamon/issues/8539

ggcode:
Finally closing this PR. Because of.. well you know: https://github.com/micheleg/dash-to-dock/pull/1253 ;)
@paroj thanks for continuing. :sunglasses: :thumbsup:

PR #522, opened 2017-06-14
Emit 'toggled' signal to facilitate reconnecting to dock events

passingthru67:
With the merging of the multi-monitor branch, the dock is no longer a single object but an array of objects. It would be nice to have a 'toggled' signal from the dockManager to let other extensions know when the dock(s) have been destroyed and recreated. This would facilitate reconnecting to the dock(s) events such as showing, hiding, and mouse hover.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -1607,6 +1607,7 @@ const DockManager = new Lang.Class({
     _toggle: function() {
         this._deleteDocks();
         this._createDocks();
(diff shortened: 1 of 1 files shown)


micheleg:
Looks perfectly reasonable to me. I've merged the commit into master [4c9480a4b2c]. Thank you.

PR #531, opened 2017-06-17
Implement monitor-isolation.

franglais125:
This is the monitor isolation work.

There are a few extra changes in appIcons.js and windowPreview.js that fix inconsistencies introduced when moving the window-handling function to methods.

Other than that, I haven't found problems... Since I had to rebase this, I'll keep testing though, just in case.

### DIFF
--- a/windowPreview.js
+++ b/windowPreview.js
@@ -19,7 +19,6 @@ const Workspace = imports.ui.workspace;
 
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 const Utils = Me.imports.utils;
-const AppIcons = Me.imports.appIcons;
 
 const PREVIEW_MAX_WIDTH = 250;
 const PREVIEW_MAX_HEIGHT = 150;
@@ -28,7 +27,7 @@ const WindowPreviewMenu = new Lang.Class({
     Name: 'WindowPreviewMenu',
     Extends: PopupMenu.PopupMenu,
 
-    _init: function(source, settings, monitorIndex) {
+    _init: function(source, settings) {
         this._dtdSettings = settings;
 
         let side = Utils.getPosition(settings);
@@ -40,6 +39,7 @@ const WindowPreviewMenu = new Lang.Class({
 
         this._source = source;
         this._app = this._source.app;
+        let monitorIndex = this._source._monitorIndex;
 
         this.actor.add_style_class_name('app-well-menu');
         this.actor.set_style('max-width: '  + (Main.layoutManager.monitors[monitorIndex].width  - 22) + 'px; ' +
@@ -60,7 +60,7 @@ const WindowPreviewMenu = new Lang.Class({
         this._boxPointer._arrowSide = side;
         this._boxPointer._userArrowSide = side;
 
-        this._previewBox = new WindowPreviewList(this._app, this._dtdSettings);
+        this._previewBox = new WindowPreviewList(this._source, this._dtdSettings);
         this.addMenuItem(this._previewBox);
     },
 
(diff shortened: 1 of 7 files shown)


micheleg:
Thanks, looks good. I have now realised that I pushed my changes to master [ee55b90c969d14c53307f762e842302262b0c325], which is not what I intended to do, and now understand your confusion yesterday.

windowPreview.js
```diff
@@ -95,7 +95,7 @@ const WindowPreviewList = new Lang.Class({
     Name: 'WindowPreviewMenuSection',
     Extends: PopupMenu.PopupMenuSection,
 
-    _init: function(app, settings) {
```
micheleg: Reason of this? Consistency? Future extension?

franglais125: The reason is we can use the `source`  in line 209 for 
```
let newWin = this._source.getInterestingWindows().sort(this.sortWindowsCompareFunction);
```
At the moment `this._source` is undefined.

The alternative is passing the necessary arguments to call the global function `getInterestingWindows`.

windowPreview.js
```diff
@@ -60,7 +60,7 @@ const WindowPreviewMenu = new Lang.Class({
         this._boxPointer._arrowSide = side;
         this._boxPointer._userArrowSide = side;
 
-        this._previewBox = new WindowPreviewList(this._app, this._dtdSettings);
+        this._previewBox = new WindowPreviewList(this._source, this._dtdSettings);
```
franglais125: Regarding your previous comment, I've changed here the input type as well.

windowPreview.js
micheleg: I see, thanks. Was looking at your changeset.

micheleg:
I've cleaned up the history and then applied your patch here: https://github.com/micheleg/dash-to-dock/tree/development. I should not have changed anything.

franglais125:
`git diff upstream/development` Gives me nothing :)

franglais125:
@micheleg I rebased this based on your development branch.

I was testing the monitor isolation work to make sure it was working, and realized the click actions were  a bit inconsistent... So I added a few commits. I think it's better like this. Let me know what you think:

- [efbcde1c610e5259aeeb005682e5d81eca7a6370], [d2f853df686cff9c434fe7b83e3b6dc51418a23e] and [844832df9315c753a25136b294552fa0d6f711e8] essentially replace `app.activate()` with `Main.activateWindow(w)`.
- [137655fd893c2708beb6d7551428fc9795721a38] is a small simplification.

micheleg:
Great!
app.activate() and Main.activateWindows() seem to do the same thing when the app is running (beside our ability to select the window to be activated), so it should be ok.

 I have combined your commits and reworded the messages, and finally merged into master [93c5a7a9462f03b9c9e825ced056ddacf3d8c350..5c6208ddc1adcc0c03d6a0a976d540d68a50de0a]

Can't really test extensively on multimonitor still (but I might update my work machine to Stretch at a certain point). We'll fix any additional corner case in due course with the help of users feedback. Closing this for now.

franglais125:
I just pushed a small fix to the monitor_isolation branch. https://github.com/franglais125/dash-to-dock/commit/b4e2148380ad5c39e8e69cf8b6cf822eaa4b5156.

I moved a signal from appIcons.js to docking.js to improve Wayland compatibility.

https://github.com/franglais125/dash-to-dock/tree/monitor_isolation

micheleg:
Merged [058e97d57], thanks.

franglais125:
Another small fix. It's a partial revert of the previous one. We actually need the signal in both appIcons.js and docking.js.

https://github.com/franglais125/dash-to-dock/commit/b3d03c413a52249fc7c98da6e06b12721beaeb41

in https://github.com/franglais125/dash-to-dock/tree/monitor_isolation

micheleg:
Thanks, I've merged it [bccbc4d54dd9974cc66e5965d7731b1359a09967].

PR #558, opened 2017-08-07
Don't use markups in a normal string

seb128:
The original strings doesn't use markups and the code seems to expect a normal label, the current translation results in <i></i> to be displayed in the text.


micheleg:
Thank you! Merged [184b150ecd8227bc929fe6e24ce3b14298352345].

PR #561, opened 2017-08-10
Support system-wide installation

didrocks:
System-wide installation should install locale/ and gsettings schema in system
directories. This enables override supports, locale stripping and not shipping
pre-compiled schema file.


micheleg:
Great! Committed as [4dd565dad920ce44303635b550a39bbdbe34fb70].

PR #570, opened 2017-08-20
Fixed a translation error in the italian locale

bertof:
Just a quick fix to the locale. "Scroll action" has been named as "Azione clic" which is the one for "Clic action".


bertof:
Sorry for the double post

PR #571, opened 2017-08-20
Fixed a translation error in the italian locale

bertof:
Just a quick fix to the locale. "Scroll action" has been named as "Azione clic" which is the one for "Clic action".


micheleg:
Merged! [04bb0a901ff37542f74d69a80e761dd145f39833]. Grazie.

bertof:
Di nulla. Grazie dell'estensione, anzi!

-------- Original Message --------

PR #573, opened 2017-08-21
Monitor isolation: limit the impact of 'window-entered-monitor'.

franglais125:
In Wayland sessions, this signal is needed to track the state of windows dragged from one monitor to another. However, the performance impact is questionable as it triggers a redraw of all icons' style quite often.

To limit this, we only really connect to this signal if 'isolate-monitors' is true, and if there are at least 2 monitors.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -128,7 +128,7 @@ var MyAppIcon = new Lang.Class({
             this._signalsHandler.addWithLabel('isolate-monitors', [
                 global.screen,
                 'window-entered-monitor',
-                Lang.bind(this, this.onWindowsChanged)
+                Lang.bind(this, this._onWindowEntered)
             ]);
         }
 
@@ -183,6 +183,16 @@ var MyAppIcon = new Lang.Class({
 
         if (this._scrollEventHandler)
             this.actor.disconnect(this._scrollEventHandler);
+    },
+
+    _onWindowEntered: function(metaScreen, monitorIndex, metaWin) {
+        let updateStyle = false;
+        let app = tracker.get_window_app(metaWin);
+        if (app.get_id() == this.app.get_id())
+            updateStyle = true;
+
+        if (updateStyle)
+            this.onWindowsChanged();
     },
 
     _optionalScrollCycleWindows: function() {
--- a/appIcons.js
+++ b/appIcons.js
@@ -116,7 +116,21 @@ var MyAppIcon = new Lang.Class({
                                                  Lang.bind(this,
                                                            this._onFocusAppChanged));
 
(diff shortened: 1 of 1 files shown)


micheleg:
Aren't we missing a connection to the change of the 'isolated-monitors' settings?

franglais125:
We already have a connection to this in `docking.js`, which reset the icons: https://github.com/micheleg/dash-to-dock/blob/master/docking.js#L1556

We need this there because some icons might disappear, depending on what apps are running on each monitor (same as with workspace isolation).

micheleg:
Ups, I was confused by that because it is hidden in the WorkspaceIsolation. 

I'm going to go forward with this, but I was thinking: we could get the application corresponind to the window that triggered the signal,  and only carry on the update of the icon style if it match the app of the icon. Does it make sense?

micheleg:
Started the merging  in the development branch.

franglais125:
Yes, sounds good. I was actually taken aback as well... I thought I had forgotten the signal and didn't understand why it was working in the first place!

>  we could get the application corresponind to the window that triggered the signal, and only carry on the update of the icon style if it match the app of the icon

I would love this. I actually didn't have enough time to look into it, sorry. I've been busier than usual. Once you merge this into master, I'd close this PR and open an issue as a reminder of this further optimization.

micheleg:
Yes, there's no hurry. Just dropping the idea here for future reference :)

franglais125:
@micheleg I included a tentative improvement [1dc3537a36bb725e352b5dba12432818dd7c809d]. I still have to test it. I rebased on your development branch as well.

franglais125:
@micheleg I added the better optimization now in [bf18ad1], and it's tested this time.

I had to slightly modify `getInterestingWIndows()`, the problem was that I couldn't get a proper handle to update icons in the monitor that the window is *leaving*, as `this.getInterestingWindows()` would only give us windows from the current monitor, and the match (`win == metaWIn`) would never work.

franglais125:
@micheleg Great suggestion, thanks for the pointer! [55fe51bfd79dbaea2d7135c31411963788d9f161].

micheleg:
Great, I have added this commit to the development branch. i will cleanup and merge into master soon.

franglais125:
Ugh, sorry, I actually just fixed it again.. For the 3rd time! [ca376eac6fb6f496b4f3b0142e1d086752b928bb]

(As obviously there is no need anymore for `getInterestingWindows()`)

micheleg:
Merged in master as  9ad9e0651af and  c6c56866e61.

PR #589, opened 2017-09-05
Add counters

azzar1:
Initial support to 'com.canonical.Unity.LauncherEntry' APIs. At the moment it only supports the badge notification. The code can be tested using the following script: http://paste.ubuntu.com/25473314/


joaosardinha:
What is the status on this? Would like to test it!

PR #599, opened 2017-09-22
Signal when the dash (and top panel) transparency has changed.

passingthru67:
NOTE:
This will help workspaces-to-dock (and other extensions) to sync with the panel/dash transparency.

### DIFF
--- a/po/id.po
+++ b/po/id.po
@@ -1,0 +1,450 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Dash-to-Dock\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2017-06-04 12:35+0100\n"
+"PO-Revision-Date: 2017-10-02 11:25+0700\n"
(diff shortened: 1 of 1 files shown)


micheleg:
Thanks for the pull request. I'm intentionate to support this. What is yout goal? Which specific transparency change are you trying to track? i'm asking because we are supporting different transparent behaviours and I'm not sure if these signals are enough.

passingthru67:
This is mainly for the adaptive transparency (because it affects the top panel), but should also apply to the dynamic transparency. My goal is to provide dynamic/adaptive transparency for workspaces-to-dock that stays in sync with dash-to-dock and the top panel.

micheleg:
Woudn't inside this https://github.com/micheleg/dash-to-dock/blob/master/theming.js#L437 be the right place to emit the signals, together with the setting of the two styles?

passingthru67:
That's where I thought I had put it. Here's the three functions that I've added emit signals to. I'm pretty sure I synced with upstream changes before I made the mods.

enable .. emit 'transparency-enabled'
disable .. emit 'transparency-disabled'
_updateSolidStyle .. emit 'update-solid-style'

micheleg:
Sorry, I got fooled by the code folding... so you just want a ''update-solid-style', without information on which status (solid/transparent) is set?

passingthru67:
It would be nice to include that information in the emit signal if possible.

micheleg:
some user_data could be added to the signal (this.emit('update-solid-style; true/false)), or two indepedent signals could be emitted. I would do the first.

passingthru67:
That would be nice. Do you want me to rework the pull request with the changes?

micheleg:
I would appreciate it :)

PR #608, opened 2017-09-28
Load gettext dashtodock domain before calling gettext()

didrocks:
appIcons.js is assigning const __ = Gettext.gettext.
This is what enables to have "All windows" and other non Shell strings (dash to dock specific) translated.
However, the domain was set in init(), after this assignement, which was thus loading the GNOME Shell domain. Ensure we bind the domain before (which is a really fast call).


didrocks:
Let me close that one, seems that I can't reproduce reliably the race which justified putting this i18n domain init outside of the `init()`.

Issue #454, opened 2017-03-20
Most maximized applications are resized after unlock


gyeben:
After locking and unlocking my computer, most maximized applications are resized.

It also happens with no other extensions enabled.
I am using Fedora 25, Gnome 3.22 with Wayland and the latest version of dash-to-dock.

![kepernyokep errol 2017-03-20 23-32-49](https://cloud.githubusercontent.com/assets/2202350/24124926/3dfe09d0-0dc6-11e7-9161-a5c044cef516.png)
![kepernyokep errol 2017-03-20 23-32-57](https://cloud.githubusercontent.com/assets/2202350/24124925/3dfcb706-0dc6-11e7-9e02-46ee4fc6130b.png)

franglais125:
I think this is a duplicate of https://github.com/micheleg/dash-to-dock/issues/423

Also, I can confirm this problem after unlocking. Debian Stretch, Wayland, Gnome 3.22.

micheleg:
Even in current master? Only under Wayland?

franglais125:
Yes, I'm always running master :). Let me give it a try on X as well, as I haven't used it in a while (Wayland is quite pleasant!).

As a side note, let me add that this is not 100% reproducible. It just seems to happen sometimes after a screen unlock (maybe 1/10 times). I'll try to pay more attention and see if I can get a pattern... I wasn't too bothered by this.

gyeben:
It only happens under Wayland.

franglais125:
In my case too it seems to happen only under Wayland. And now that I'm paying attention, it happens much more often than I thought: roughly 50% of the times many of the maximized windows will get resized (but not necessarily all).

Also, the problem only occurs with native Wayland applications. Is this the case for you too, @gyeben ? For example, I never see this happen with Firefox, Thunderbird or VLC.

franglais125:
Another way of triggering something similar, is by successively disabling/enabling dash-to-dock from the Tweak tool (with all other extensions turned off). It's not even necessary to do it quickly, just doing it a few times will sometimes yield the wrong window size for native Wayland applications (if maximized).

i guess this is due to the dash spacer that is introduced when enabling the extension.

gyeben:
@franglais125: yeah, it seems like only the native Wayland applications are affected.

franglais125:
I have a proposed fixed here: https://github.com/franglais125/dash-to-dock/tree/fix_maximize_unlock

[...]

PR #609, opened 2017-09-28
Dash: properly initialize the icon size.

franglais125:
It took me some time to identify the culprit. I thought it was due to several calls to `_resetPosition`, but this was in fact due to quickly resizing the icons when initializing.

Hence, here we:
- Immediately set the icon size upon initilizing.
- Just in case we call `resetPosition` in succession, we all an `idle` call.

Results:
- faster and lighter initialization
- prevents flickering when unlocking the screen
- on Wayland: some windows used to not be maximized properly after unlocking

Addresses #454

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -179,12 +179,15 @@ var MyDash = new Lang.Class({
     Name: 'DashToDock.MyDash',
 
     _init: function(settings, remoteModel, monitorIndex) {
+        this._dtdSettings = settings;
+
+        // Initialize icon variables and size
         this._maxHeight = -1;
-        this.iconSize = 64;
+        this.iconSize = this._dtdSettings.get_int('dash-max-icon-size');
         this._availableIconSizes = baseIconSizes;
         this._shownInitially = false;
+        this._initializeIconSize(this.iconSize);
 
-        this._dtdSettings = settings;
         this._remoteModel = remoteModel;
         this._monitorIndex = monitorIndex;
         this._position = Utils.getPosition(settings);
@@ -902,7 +905,7 @@ var MyDash = new Lang.Class({
         });
     },
 
-    setIconSize: function(max_size, doNotAnimate) {
+    _initializeIconSize: function(max_size) {
         let max_allowed = baseIconSizes[baseIconSizes.length-1];
         max_size = Math.min(max_size, max_allowed);
 
@@ -914,6 +917,10 @@ var MyDash = new Lang.Class({
             });
             this._availableIconSizes.push(max_size);
         }
+    },
+
+    setIconSize: function(max_size, doNotAnimate) {
+        this._initializeIconSize(max_size);
 
         if (doNotAnimate)
             this._shownInitially = false;
--- a/docking.js
+++ b/docking.js
@@ -430,8 +430,6 @@ const DockedDash = new Lang.Class({
             this._paintId=0;
         }
 
-        this.dash.setIconSize(this._settings.get_int('dash-max-icon-size'), true);
-
         // Apply custome css class according to the settings
         this._themeManager.updateCustomTheme();
 


dash.js
```diff
@@ -179,12 +179,15 @@ var MyDash = new Lang.Class({
     Name: 'DashToDock.MyDash',
 
     _init: function(settings, remoteModel, monitorIndex) {
+        this._dtdSettings = settings;
+
+        // Initialize icon variables and size
         this._maxHeight = -1;
-        this.iconSize = 64;
+        this.iconSize = settings.get_int('dash-max-icon-size');
```
micheleg: to avoid confusion with the call below, use here this._dtdSettings as well?

dash.js
```diff
@@ -179,12 +179,15 @@ var MyDash = new Lang.Class({
...
         this._availableIconSizes = baseIconSizes;
         this._shownInitially = false;
+        this._initializeIconSize(this._dtdSettings.get_int('dash-max-icon-size'));
```
micheleg: and here pass directly this.iconSize...

micheleg: Is this superfluous now?https://github.com/micheleg/dash-to-dock/blob/master/docking.js#L433.

micheleg:
From your comment it seem that you added the  idle thing on reset_position just in case, but you actually think it is not necessary?

franglais125:
I altered the order of your questions in my explanation, so that the second part
makes more sense.


## Logic
> Did you actually understood (i.e. can you explain to me) what's the code flow
that was causing the flickering?

Sure. The problem was within dash.js, here is the succession of events:
1. Icons get initialized to a size of **64**px (value taken from upstream).
2. Eventually, `_redisplay()` gets called, which calls `_adjustIconSize()`,
which in turns calls `setIconSize()`.
3. Now, in this function we have a problem. `iconSize` was set to 64, and we try
to look for the min here:

```
let max_allowed = baseIconSizes[baseIconSizes.length-1];
max_size = Math.min(max_size, max_allowed);
```

In my case (small screen), the `max_allowed` value is 48px, so now we transition
from **64->48**.

4. Finally, in docking.js, the painId timeout comes in (which calls
`_initialize()`, and finally sets the actual icon size the user requested (in my
 case this was 40px). Hence, again we transition from **48->40px**.

Every icon size transition does not necessarily map to a resetPosition call, as
it may depend on some factors (max_allowed value, iconSize set by the user,
etc). For instance, on a reasonably large screen, and if the user sets the icon
size to 64px, there will be no transition at all and things will look smooth. In
my case (and I guess many other users'), there is up to 2 size transitions,
causing 2 separate `_resetPosition` calls (plus the initialization one, which ofc is always there).

Hence, I essentially changed two things:
1. I `init` MyDash with a more adequate iconSize.
2. I then immediately call `setIconSize`, to properly set the max_size (which is
stored as `this._availableIconSizes`).


## Idle call
> From your comment it seem that you added the idle thing on reset_position just
in case, but you actually think it is not necessary?
I'm not even sure it that idle does nothing? each call will reset the id,
therefore the next will run in any case. Maybe you wanted to not add the idle
again if the id had not been reset yet?

Before adding a new idle call, I am removing the old one. I did check manually
that the call was indeed removed. In general, the calls are valid, as the
workareas are actually changed. However (as explained above), in some cases the
size changes were in quick succession, and two calls would be triggered within a
few milliseconds, resulting in one valid call, and a redundant one. If we fix
the problem above, we then don't need this idle call.

---------------------------
Upon more testing (just now), I actually think it is smoother if we don't do this. I initially added it since I was encountering many `resetPosition` calls, but now I actually find it detrimental.

dash.js
franglais125: Done!

dash.js
franglais125: Done, both!

micheleg:
Ok, it looks good to me.

micheleg:
Merged in master [fd81051ee0ae8320e]. Thanks!

franglais125:
Thanks! This was one of the annoying bits of running Wayland for me :)

PR #614, opened 2017-10-02
Unity7 backlight: optional glossy effect

franglais125:
@micheleg Just a small PR on something I tried. I turned off the glossy effect and IMO it looks cleaner, especially if only a few apps are open.

I added an option for this. Here are a few screenshots:

screenshot from 2017-10-02 19-02-29

**Before:**
screenshot from 2017-10-02 19-01-51

**After:**
screenshot from 2017-10-02 19-04-31

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -308,6 +308,11 @@
                       </object>
                     </child>
                     <child>
+                      <object class="GtkCheckButton" id="apply_gloss_effect_checkbutton">
+                        <property name="label" translatable="yes">Apply glossy effect.</property>
+                      </object>
+                    </child>
+                    <child>
                       <object class="GtkBox">
                         <property name="can_focus">0</property>
                         <child>
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -288,6 +288,7 @@ var RunningIndicatorDots = class DashToDock_RunningIndicatorDots extends Running
                    'custom-theme-running-dots-border-width',
                    'custom-theme-customize-running-dots',
                    'unity-backlit-items',
+                   'apply-glossy-effect',
                    'running-indicator-dominant-color'];
 
         keys.forEach(function(key) {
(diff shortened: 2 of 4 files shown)


SavageTiger:
@franglais125 I think this is a good addition, if someone likes flat-designed docks, but also the backlight colors they might be able to enjoy the best of both worlds.

Maybe the title should be changed, to "Enable unity7 like backlit items", since they might not be glossy anymore :)

Also if somebody has the "Show applications" icon at the top or start of their bar, it looks a bit weird that is has no glossy svg, do you think we need an option to also add glossy to that item?

franglais125:
> @franglais125 I think this is a good addition, if someone likes flat-designed docks, but also the backlight colors they might be able to enjoy the best of both worlds.

 Thanks :), and sorry for the late response!

> Maybe the title should be changed, to "Enable unity7 like backlit items", since they might not be glossy anymore :)

Indeed! I'll change this today/tomorrow with your suggestion.

> Also if somebody has the "Show applications" icon at the top or start of their bar, it looks a bit weird that is has no glossy svg, do you think we need an option to also add glossy to that item?

I'm a bit unsure about this one, after all it's not an application that will be active (or "lit" in this case). But I'll leave this up to you and @micheleg, if it's an option then it doesn't hurt, except with clutter in the UI.

franglais125:
Rebased following the code reorganization!

franglais125:
Rebased against current master.

I still think this can give a cleaner look to those who prefer less gloss:


Screenshot from 2019-10-16 13-18-43

Screenshot from 2019-10-16 13-18-26

3v1n0:
> @franglais125
> do you still plan to integrate this feature?
> 
> We could really use this feature in Ubuntu
> Best regards from the Yaru team heart

@Feichtmeier are you still interested in having this?

I may include this into the ubuntu-dock branch for now, so you guys can play around it till UI freeze maybe?

Feichtmeier:
I'd say yes, but it depends on how much control we have over the tint. A soft transparent tint of the main colour of the app icon might look very nice. Let's wait for the others of our crew
While you are on it, could you remove the hardrcoded running dot colour so we can style them in yaru?

clobrano:
I'm not much in favour actually. I like the colorful effect, but long term seems a bit messy for me

3v1n0:
> I'd say yes, but it depends on how much control we have over the tint. A soft transparent tint of the main colour of the app icon might look very nice. Let's wait for the others of our crew
> While you are on it, could you remove the hardrcoded running dot colour so we can style them in yaru?

@Feichtmeier please open a bug for that (or even a PR :)), and we can handle that

Jhackler:
I would like this to be a thing as well

3v1n0 (approved):
Finally rebased and adapted to current code, this can go in now.

PR #615, opened 2017-10-04
Add signals to indicate when the dash style/transparency have changed.

passingthru67:
NOTE:
This will help workspaces-to-dock (and other extensions) to sync with the panel/dash transparency.


micheleg:
Merged [4ce4464987ddd62]. Thanks.

PR #616, opened 2017-10-04
Update dockManager variable declaration for Gnome 3.26 GJS (mozjs52)

passingthru67:
Dash-to-Dock's dockManager properties are not accessible by the workspaces-to-dock extension after locking/unlocking the Gnome 3.26 desktop or after disabling/enabling the extension (see https://github.com/passingthru67/workspaces-to-dock/issues/119). What's weird is that it seems to work fine on the initial login to the desktop. Changing the dockManager declaration from 'let' to 'var' fixes the issue.


micheleg:
Ok. merged [e2745fde3124e9].

PR #619, opened 2017-10-08
RFC: Trash and Removable Device Icons

philipl:
This set of changes adds Trash and Removable Device/Drive icons similar to what the Unity dock offers. 

I don't consider it a complete change at this point; I think the code organisation isn't correct, if nothing else, but I'm particularly interested in feedback on the way I created the icons using DesktopAppInfo.

Other obvious shortcomings:
- Lack of Translations
- Nautilus dependency in Empty Trash call
- No drag-and-drop for Trash (I don't personally care about this)
- 'Inauthentic' placement of Trash Icon

dash

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 UUID = dash-to-dock@micxgx.gmail.com
 BASE_MODULES = extension.js stylesheet.css metadata.json COPYING README.md
-EXTRA_MODULES = convenience.js dash.js docking.js appIcons.js appIconIndicators.js launcherAPI.js locations.js windowPreview.js intellihide.js prefs.js theming.js utils.js Settings.ui
+EXTRA_MODULES = convenience.js dash.js docking.js appIcons.js appIconIndicators.js fileManager1API.js launcherAPI.js locations.js windowPreview.js intellihide.js prefs.js theming.js utils.js Settings.ui
 EXTRA_MEDIA = logo.svg glossy.svg highlight_stacked_bg.svg highlight_stacked_bg_h.svg
 TOLOCALIZE =  prefs.js appIcons.js locations.js
 MSGSRC = $(wildcard po/*.po)
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,9 @@
 
 UUID = dash-to-dock@micxgx.gmail.com
 BASE_MODULES = extension.js stylesheet.css metadata.json COPYING README.md
-EXTRA_MODULES = convenience.js dash.js docking.js appIcons.js appIconIndicators.js launcherAPI.js windowPreview.js intellihide.js prefs.js theming.js utils.js Settings.ui
+EXTRA_MODULES = convenience.js dash.js docking.js appIcons.js appIconIndicators.js launcherAPI.js locations.js windowPreview.js intellihide.js prefs.js theming.js utils.js Settings.ui
 EXTRA_MEDIA = logo.svg glossy.svg highlight_stacked_bg.svg highlight_stacked_bg_h.svg
-TOLOCALIZE =  prefs.js appIcons.js
+TOLOCALIZE =  prefs.js appIcons.js locations.js
 MSGSRC = $(wildcard po/*.po)
 ifeq ($(strip $(DESTDIR)),)
 	INSTALLTYPE = local
--- a/Settings.ui
+++ b/Settings.ui
@@ -1332,6 +1332,90 @@
                     </child>
                   </object>
                 </child>
+                <child>
+                  <object class="GtkListBoxRow" id="listboxrow18">
+                    <property name="visible">True</property>
+                    <property name="can_focus">True</property>
+                    <child>
+                      <object class="GtkGrid" id="shrink_dash4">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="margin_left">12</property>
+                        <property name="margin_right">12</property>
+                        <property name="margin_top">12</property>
+                        <property name="margin_bottom">12</property>
+                        <property name="column_spacing">32</property>
+                        <child>
+                          <object class="GtkSwitch" id="show_trash_switch">
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                            <property name="halign">end</property>
+                            <property name="valign">center</property>
+                          </object>
+                          <packing>
+                            <property name="left_attach">1</property>
+                            <property name="top_attach">0</property>
+                          </packing>
+                        </child>
+                        <child>
+                          <object class="GtkLabel" id="shrink_dash_label4">
+                            <property name="visible">True</property>
+                            <property name="can_focus">False</property>
+                            <property name="hexpand">True</property>
+                            <property name="label" translatable="yes">Show trash can</property>
+                            <property name="xalign">0</property>
+                          </object>
+                          <packing>
+                            <property name="left_attach">0</property>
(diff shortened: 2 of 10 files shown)


franglais125:
I'd love to have this! It seems non-trivial ti implement though, and not having d-n-d is too bad (although nothing we can do about it).

Quick comments:
1. Clicking on the icon is supposed to open the location? This is not working for me now. WIP?
2. On Debian, external USB drives don't seem to have the correct icon:

screenshot from 2017-10-08 13-42-51

Cheers!

philipl:
My handling of the icon is too simple. It's a themed icon and so has multiple names, and the name that actually yields an icon varies depending on theme, and indirectly that depends on the distro.

I would need to iterate over the names and pick the first one that actually works.

philipl:
I've updated the pr with icon lookup; please try it out.

franglais125:
Nice and quick fix! It's showing the correct icon now, at least for this drive I just tested.

philipl:
Another known problem is that opening a removable drive location creates a window that's mapped to Nautilus in the dock - so if you click on it again, you get a new window each time.

Unity had this exact problem initially as well, so the Canonical folks patched in a dbus mechanism to get a mapping of XIDs to file locations which they used to recognise removable drive windows.

The original bug report is here: https://bugs.launchpad.net/unity/+bug/887449

and there's a patch in the ubuntu sources for Nautilus.

The next step on my side is to work out the best west to override the window -> app mappings that the shell is doing.

philipl:
Just pushed an incremental update to implement Unity's window mapping for the special icons.

philipl:
I've done some refactoring so that all the new logic is properly contained. I feel good about the code organisation now.

philipl:
Adding translation support.

@micheleg: I consider this mergeable now, if you're interested in looking.

franglais125:
Thanks a lot for the continuous work on this! I've been using it since last time with great success (and some minor quirks).

I tested today the latest code from your branch, but I can't get the extension to load properly. Upon logging in, the shell freezes for ~10 seconds, and then loads with the extension disabled and this error in the log:

```
Oct 13 16:10:38 v131 gnome-shell[14761]: Extension "dash-to-dock@micxgx.gmail.com" had error: Gio.IOErrorEnum: Error calling  StartServiceByName for org.freedesktop.FileManager1: Timeout was reached
```

This is on Gnome 3.26, Wayland.

philipl:
Thanks for the feedback.

The special window matching logic that Ubuntu patches into nautilus is X11 specific as it identifies windows by XID. I have no idea what happens in Wayland and cannot easily test yet as my primary machine uses nvidia graphics. I'll be able to try it out in a couple of weeks when I can upgrade another machine to 17.10. I had hoped that it would just fail to report windows and wouldn't do something nasty like timeout.

But one thing - if you log out and back in again, does it still block for 10 seconds? I had a problem where it would get stuck like this and I had to log out to fully restart gnome shell.

philipl:
Pushed a few improvements including what should be a fix for freezing if the DBus proxy decided to be slow to init.

micheleg:
I tried to run the code on 3.22, but it is not compatible due to the use of the operator ```...```. I want  to maintain compatibility with older versions for the moment, so these javascript features should not be used.

philipl:
I've updated it to use push.apply() syntax. Should work now.

micheleg:
```Map.forEach``` is also not supported unfortunately.  The supported syntax is  ```for (let [k,v] of Map){ }```.  So is ```Array.from()```. It's annoying not being able to use these, but I prioritize back compatibility. 

Anyhow, I got the branch running.

philipl:
New update. I've fixed the js compatibility issues you mentioned, and added an improvement to deduplicate when a window is listed twice due to a location being open in multiple tabs.

philipl:
I had a chance to test in Wayland. As expected, window tracking doesn’t work, but it fails gracefully. It looks like the Nautilus patch would need enhancements.

@micheleg, are you interested in merging this? Is there anything you’d like me to address?

didrocks:
@franglais125: On drag and drop, we had "virtual placeholder" in unity if  you are interested. Basically, the default of the gsettings key was:
[<list of favorites app>, 'unity://running-apps',  'unity://expo-icon', 'unity://devices']

The running apps were the additional applications running but not in favorites, expo-icon, was the workspace expo mode (doesn't have any equivalent in GNOME Shell) and devices were those mounted and unmounted removable devices with an option to mount/unmount/unplug them.

So, moving one item was moving the whole group. I don't say that's what you want to do here, just giving a possible solution we explored on Unity some years ago ;)

0matgal0:
Works okaish. I have a 2 issues:
- the trash / removable icons are not movable (even though it might seem like they can be moved)
- opening trash actually opens nautilus not trash itself (as you can see the trash icon is not highlighted):
image

I know that unity had some issues with decoupling nautilus with device icons but afaik they somehow figured it out in 16.04 or something like that.

philipl:
@0matgal0. I'm guessing you're using Wayland. As I said above, window matching doesn't work on Wayland.

philipl:
I've updated the pull request to adjust for the appIconIndicators refactoring. @micheleg can you let me know if you'd ever accept this change? or there's work you want me to do to it before you'd accept it?

franglais125:
@philipl While I can't speak for @micheleg , on my side I ran out of time in the last few months and couldn't test your branch more thoroughly. I hope I'll get back to it some day, as it's pretty neat (and something I remember from Unity as well).

If it's any indication, it took me ~6 months to get the multi-monitor branch merged.  But it was nicely done eventually, with some nice rework by Michele.

franglais125:
Just an idea: would it be possible to add some configurability?

1. I'm sure some users won't want this at all.

2. Perhaps make the trash configurable as well?

Something like:


screenshot from 2018-01-07 20-08-47

philipl:
Done!

franglais125:
Gorgeous, thanks! For anyone reading, here is a screenshot of the prefs:

screenshot from 2018-01-08 23-41-13

micheleg:
Hi,

sorry for the late reply.  Thanks for your work. I quickly tested your code a week ago and I was going to ask for some settings but I see that it has been done now. I still find it a bit incomplete ( without drag and drop support, yet the changes seems decently self-container so I'm keen on trying to merge this. 

Can you please summarize if there's any known corner case or issue? I didn't notice anything wrong on my side.  I'll now  try to look more thoroughly at your code.

philipl:
The issues I'm aware of:
- No drag and drop for trash. I think the best thing to do here is let the gnome shell folks implement this for [their desktop icons extension](https://gitlab.gnome.org/csoriano/org.gnome.desktop-icons) and then copy/use it. :-)
- No support for window tracking with Wayland. This requires the Ubuntu team to extend the window tracking system to work with Wayland. Without it, Nautilus windows for trash/devices cannot be associated with the icons
- Trash icon cannot be pinned to the bottom of the bar as in Unity. This would probably be easier for you, but I really struggled to understand the St layout code.

Thanks for merging!

franglais125:
It seems this was un-merged...

Perhaps a local `push -f` before pulling, or simply wrongly merged?

philipl:
Yeah, there was a force push reported and now it's not on master.

micheleg:
Yes, my fault, I pushed your branch by mistake while merging another patch. Sorry for the confusion.

philipl:
Can you reopen this pull request then? It won't let me do it.

PR #637, opened 2017-11-07
Disable transparency state computation when dock is hidden

### LINKED ISSUE #636: Do not take dock into account for global transparency when hiding
I have read some feedbacks over the net that some people were complaining about the panel turning opaque when there is no reason for this and it was flickery.

I found it weird until I realized that they probably are in hidden mode (either autohide or intellihide), dock hidden and have a window not near the top panel. They can end up in that situation:

Dock hidden

It's even worse in autohide than intellihide, as you don't get the reason why the opacity changes.

I think what makes sense is to not take the dock into account for window placement when it's going to be hidden and only consider top panel one.

Of course, it means that the panel will be shown as minimum transparency when reappearing (on top of any application), but I find it less confusing and more predictable that way. Thoughts?

didrocks:
When dock is either in autohide or intellihide mode, taking the dock vs window
position into account can lead to confusing behavior, hard to understand for
users.
Fixes #636.

### DIFF
--- a/schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
@@ -67,7 +67,7 @@
     <key name="transparency-mode" enum="org.gnome.shell.extensions.dash-to-dock.transparency-mode">
       <default>'DEFAULT'</default>
       <summary>Transparency mode for the dock</summary>
-      <description>FIXED: constant transparency. ADAPTIVE: lock state with the top panel. DYNAMIC: dock takes the opaque style only when windows are close to it.</description>
+      <description>FIXED: constant transparency. ADAPTIVE: lock state with the top panel when not hidden. DYNAMIC: dock takes the opaque style only when windows are close to it.</description>
     </key>
     <key type="b" name="customize-alphas">
       <default>false</default>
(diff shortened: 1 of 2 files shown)


franglais125:
As is often the case with intellihide issues, I don't feel strongly about this, as I don't use it.

One suggestion though: we can perhaps still check for the dock's state in the `'dynamic'` case, which doesn't affect the top panel. The downside to this is that users trying the different options might find this inconsistent... The reason I do like this is that the dock can be opaque when showing, making it easier to see the apps and dots/counters/number-overlays.

didrocks:
Yeah, I have some thought about dynamic and that could be inconsistent.

Or, we can mimick the dynamic case even in adapative mode only for the dock? It means:
- the top bar transparency state controls the dock transparency state.
- however, in intellihide/authoide, the dock transparency state doesn't control the top bar transparency.

That way, when revealing the dock in intellihide, it will be opaque, having or not the window  below the dock touching the top bar.

What do you think?

micheleg:
Sorry for the late reply. The second proposal sounds definitely better to me. Using the dock in autohide with a large transparency value looks quite unpleasant to me.

didrocks:
Just pushed a commit behaving this way:

Enhance adaptive mode avoiding perceived panel flickering

When the Dock is hidden (autohide/intellihide), and a window is next to it
without touching the panel, the panel was going opaque.
Disable the dock computation for panel state when in adaptive and
in autohiding mode. This ensure that the dock itself though has the correct
transparency/opaque state when revealed above a window (independently of
the panel).

Tell me what you think!

micheleg:
It seems good to me. Merged as [87a2357820175b4]. Thanks you!

PR #667, opened 2018-01-06
appIconIndicators: fix back backLight for workspace/monitor isolation.

franglais125:
Hi,

Just a quick fix for semi-broken monitor/ws isolation.

With the code reorganization (nice work btw! I still need to rebase some branches here and there), the backlights remain on when switching ws.


franglais125:
@micheleg At least on my system, this is still a necessary patch. Can you reproduce this?

To reproduce:
- Turn on Unity7 backlight
- Turn on workspace isolation
- Open a favorite app
- Switch workspace

Result: The backlight doesn't turn off, even if the app is not active on a new WS.

micheleg:
Yes, I can reproduce it. Sorry, I forgot about this. Merged in master [f7676108ef]. Thanks.

PR #675, opened 2018-01-18
Avoid repainting an unchanging dock.

vanvugt:
Since we are usually visible but not usually changing, make sure
most repaint requests don't actually require us to repaint anything.
This saves significant CPU when repainting the screen. In my case
gnome-shell's time spent under clutter_actor_paint dropped from
43% to 30% after this change. (LP: #1743976)

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -78,6 +78,11 @@ const MyDashActor = new Lang.Class({
         this.actor.connect('allocate', Lang.bind(this, this._allocate));
 
         this.actor._delegate = this;
+
+        // Since we are usually visible but not usually changing, make sure
+        // most repaint requests don't actually require us to repaint anything.
+        // This saves significant CPU when repainting the screen.
+        this.actor.set_offscreen_redirect(Clutter.OffscreenRedirect.ALWAYS);
     },
 
     _allocate: function(actor, box, flags) {


franglais125:
This is great! I love optimizations :)

Small thing I noticed: with this patch applied I see the dock's opacity flicker whenever a new window is opened. Do you see this too? This is on Gnome Shell 3.26 on Wayland.

vanvugt:
No I can't see any flicker, sorry. I am using Gnome Shell 3.26.2 on Wayland (Ubuntu 18.04).

I wonder; are you missing commit 87a23578?

vanvugt:
P.S. Another fix you will want is: https://bugzilla.gnome.org/show_bug.cgi?id=792633

micheleg:
Thanks for the pull request. I have been following the upstream bug and it seems that this optimization covers some edge cases after all. So is this still relevant?

vanvugt:
Yes this patch is still relevant. In fact more so now...

I've just noticed the bug/inefficiency is triggered by a non-maximized window constantly rendering while touching the edge of the dock (maybe because it still has a shadow that's being redrawn underneath?). So even a common use case triggers the higher CPU.

vanvugt:
Oddly, the panel is more resilient than the dock:

panel: CPU increases only when a legacy app tries to go fullscreen underneath it.
dock: CPU increases even when a small window is touching it and constantly rendering.

I've verified these not just by observing 'top' but particularly by running all use cases under Google CPU Profiler and comparing the subgraph for the CPU usage of clutter_actor_paint. So yeah I think we need this fix even more than the upstream one.

franglais125:
@vanvugt Finally found the time to show you the effect I see. Here is a short screencast: https://streamable.com/polwv. Notice that the dock's opacity flickers twice.

PS: I don't want to hijack this PR with a (perhaps silly) troubleshooting, so feel free to spin off this into a new issue if you want! Or contact me via email.

vanvugt:
That actually looks like your system is painting the dock twice, since the highlights get brighter and the background gets darker. But actually the only way I can imagine this patch somehow triggering that issue is if some subset of the dock is getting repainted during animations, triggering an accumulation of the addition/subtraction shading.

It's basically (almost) the kind of issue mentioned here:
  https://developer.gnome.org/clutter/stable/ClutterActor.html#clutter-actor-set-offscreen-redirect
where the offscreen_redirect setting can affect appearance. And actually you only get correct appearance of blended actors when you enable offscreen_redirect for all of them(!)

According to those docs, the default is mean to be "CLUTTER_OFFSCREEN_REDIRECT_AUTOMATIC_FOR_OPACITY" [https://developer.gnome.org/clutter/stable/ClutterActor.html#ClutterOffscreenRedirect]. However the docs are incorrect for mutter at least, which defaults to 0 ("NEVER"). I think that's either a mistake in mutter or at least a bad choice, for performance and appearance. You can see the same issue with the default set_offscreen_redirect(0) causing appearance problems here:
  https://bugzilla.gnome.org/show_bug.cgi?id=792634

I think the answer will be to apply the more conservative "CLUTTER_OFFSCREEN_REDIRECT_AUTOMATIC_FOR_OPACITY" to all actors, so that such transparent overlapping actors never accidentally change appearance. And that's what the clutter docs say that clutter is meant to do by default anyway:
  https://developer.gnome.org/clutter/stable/ClutterActor.html#ClutterOffscreenRedirect

Defaulting to CLUTTER_OFFSCREEN_REDIRECT_AUTOMATIC_FOR_OPACITY should solve:
  (a) This bug; and
  (b) Your flickering; and
  (c) https://bug792634.bugzilla-attachments.gnome.org/attachment.cgi?id=367198 and
  (d) Other performance issues in mutter/gnome-shell.

So that would be a fix directly to mutter...

franglais125:
Indeed, using `Clutter.OffscreenRedirect.AUTOMATIC_FOR_OPACITY` fixes the flickering issue for me!

vanvugt:
I meant applied to all clutter actors. If it stops the flickering when applied to just one actor then that implies it might not be helping to resolve the CPU issue.

micheleg:
@vanvugt What it the current understanding of this matter? This does not seems to have been merged by your team into the ubuntu branch. Is this still relevant? If empirically it is proven to help I'm happy to merge this.

vanvugt:
I've been too busy to return to this for a while but I think my previous comments are current. I'll look at  a generalised fix to mutter instead of this.

Closed for now.

vanvugt:
Refreshed and reopened as https://github.com/micheleg/dash-to-dock/pull/859

Issue #173, opened 2015-05-16
Trash on the Dash


luigidr:
Hi,

first of all, thanks for your amazing and useful extension. I'm using it daily on Fedora 21 (and soon on Fedora 22).

I was wondering if you plan to add a trash/recycle bin option for showing the trash in the dock, like in Unity or in this [derivative](https://github.com/itprojects/gnome-dock) of your extension (that doesn't work, at the moment).

What do you think? Could it be a useful feature (not only for me)?

micheleg:
This would be a nice optional addition, I agree. I'm aware of that derivative of my extension. However, I find a trash icon not that useful if one can't drag an drop things on it. Unfortunately, this is not possible right now due to lack of upstream support as explained in issue #49.

luigidr:
Thanks for the quick reply!
Personally, I drag things on the trash very rarely, but I get the point. :-)

micheleg:
So what would you do with the trash icon on the dock? Just empty it?

bulletmark:
Having a Trash icon anywhere specially convenient is an archaic idea which desktops blindly mimicked from the original 1980's Mac. Should the other "special" directories like Documents, Downloads, Pictures, etc be added to the dock also? Modern desktops, particularly GNOME, have depreciated the Trash desktop icon metaphor. I scoff whenever I see a Trash icon on anyone's Desktop so if this idea gets implemented, please ensure it can be configured off! :)

luigidr:
@micheleg Basically, yes: see the trash status (empty/not empty), open it directly, and empty it.

@bulletmark I agree on the configuration option, to turn the functionality on or off (if it will be implemented). Then, I see a difference between the trash and other "standard" directories (Documents, Downloads, etc.): they don't have a "binary state" (i.e., empty/non empty) nor any "special" action associated (i.e., "empty the trash"). Moreover, I think that "on the dock" is different from "on the Desktop": for me, it could be convenient to have the trash on the dock, but I don't have the trash on the Desktop. ;)

ainola:
From a design paradigm, the dock is supposed to be an 'Application Switcher' for managing your windows and application workflow. Trash (and other file-related activities) are best left to the file manager itself: Indeed, all the file managers have an easily-found trash icon in which a user can drag if need be. There's no reason to make the two interact from both a design perspective and code-maintenance perspective.

Just my two cents.

rickhall:
[...]

PR #677, opened 2018-01-19
Trash and Removable Device Icons

philipl:
This is a re-submission after the original pull request ( #619 ) was accidentally merged and then unmerged.

This set of changes adds Trash and Removable Device/Drive icons similar to what the Unity dock offers.
- Trash icon that reflects whether trash is empty or not, and has an action to empty the trash
- Icons for removable devices and mounted volumes with unmount and eject actions (as appropriate)
- Preferences to control whether to show icons or not
- Window tracking using either the original extended Ubuntu fileManager1 api added for Unity, or the new upstream Nautilus version (which works with Wayland)

dashtodock

To summarise the old discussion, this is mostly complete with a few issues that I wouldn't consider blockers:

- No drag and drop for trash. I think the best thing to do here is let the gnome shell folks implement this for [their desktop icons extension](https://gitlab.gnome.org/csoriano/org.gnome.desktop-icons) and then copy/use it. :-)
- Trash icon cannot be pinned to the bottom of the bar as in Unity. This could be done by someone with more knowledge than me but I really struggled to understand the St layout code.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -29,6 +29,7 @@ const Util = imports.misc.util;
 const Workspace = imports.ui.workspace;
 
 const Me = imports.misc.extensionUtils.getCurrentExtension();
+const FileManager1API = Me.imports.fileManager1API;
 const Utils = Me.imports.utils;
 const WindowPreview = Me.imports.windowPreview;
 const AppIconIndicators = Me.imports.appIconIndicators;
@@ -87,6 +88,9 @@ var MyAppIcon = class DashToDock_AppIcon extends AppDisplay.AppIcon {
         this.remoteModel = remoteModel;
         this._indicator = null;
 
+        let appInfo = app.get_app_info();
+        this._location = appInfo ? appInfo.get_string('XdtdUri') : null;
+
         this._updateIndicatorStyle();
 
         // Monitor windows-changes instead of app state.
@@ -130,6 +134,11 @@ var MyAppIcon = class DashToDock_AppIcon extends AppDisplay.AppIcon {
                 this._updateIndicatorStyle.bind(this)
             ]);
         }, this);
+        this._signalsHandler.add([
+            FileManager1API.fm1Client,
+            'windows-changed',
+            this.onWindowsChanged.bind(this)
+        ]);
 
         this._dtdSettings.connect('changed::scroll-action', () => {
             this._optionalScrollCycleWindows();
@@ -277,7 +286,7 @@ var MyAppIcon = class DashToDock_AppIcon extends AppDisplay.AppIcon {
         [rect.x, rect.y] = this.actor.get_transformed_position();
         [rect.width, rect.height] = this.actor.get_transformed_size();
 
-        let windows = this.app.get_windows();
+        let windows = this.getWindows();
         if (this._dtdSettings.get_boolean('multi-monitor')){
             let monitorIndex = this.monitorIndex;
             windows = windows.filter(function(w) {
@@ -386,7 +395,7 @@ var MyAppIcon = class DashToDock_AppIcon extends AppDisplay.AppIcon {
         // We check if the app is running, and that the # of windows is > 0 in
         // case we use workspace isolation.
         let windows = this.getInterestingWindows();
-        let appIsRunning = this.app.state == Shell.AppState.RUNNING
+        let appIsRunning = (this.app.state == Shell.AppState.RUNNING || this.isLocation())
             && windows.length > 0;
 
         // Some action modes (e.g. MINIMIZE_OR_OVERVIEW) require overview to remain open
@@ -591,7 +600,7 @@ var MyAppIcon = class DashToDock_AppIcon extends AppDisplay.AppIcon {
             // Try to manually activate the first window. Otherwise, when the app is activated by
             // switching to a different workspace, a launch spinning icon is shown and disappers only
             // after a timeout.
-            let windows = this.app.get_windows();
+            let windows = this.getWindows();
             if (windows.length > 0)
                 Main.activateWindow(windows[0])
             else
@@ -746,10 +755,19 @@ var MyAppIcon = class DashToDock_AppIcon extends AppDisplay.AppIcon {
         return false;
     }
 
+    getWindows() {
+        return getWindows(this.app, this._location);
+    }
+
     // Filter out unnecessary windows, for instance
     // nautilus desktop window.
     getInterestingWindows() {
-        return getInterestingWindows(this.app, this._dtdSettings, this.monitorIndex);
+        return getInterestingWindows(this.app, this._dtdSettings, this.monitorIndex, this._location);
+    }
+
+    // Does the Icon represent a location rather than an App
+    isLocation() {
+        return this._location != null;
     }
 };
 /**
@@ -832,7 +850,8 @@ const MyAppIconMenu = class DashToDock_MyAppIconMenu extends AppDisplay.AppIconM
                     });
                 }
 
-                let canFavorite = global.settings.is_writable('favorite-apps');
+                let canFavorite = global.settings.is_writable('favorite-apps') &&
+                                  !this._source.isLocation();
 
                 if (canFavorite) {
                     this._appendSeparator();
@@ -854,7 +873,8 @@ const MyAppIconMenu = class DashToDock_MyAppIconMenu extends AppDisplay.AppIconM
                     }
                 }
 
-                if (Shell.AppSystem.get_default().lookup_app('org.gnome.Software.desktop')) {
+                if (Shell.AppSystem.get_default().lookup_app('org.gnome.Software.desktop') &&
+                    !this._source.isLocation()) {
                     this._appendSeparator();
                     let item = this._appendMenuItem(_("Show Details"));
                     item.connect('activate', () => {
@@ -976,10 +996,18 @@ const MyAppIconMenu = class DashToDock_MyAppIconMenu extends AppDisplay.AppIconM
 };
 Signals.addSignalMethods(MyAppIconMenu.prototype);
 
+function getWindows(app, location) {
+    if (location != null) {
+        return FileManager1API.fm1Client.getWindows(location);
+    } else {
+        return app.get_windows();
+    }
+}
+
 // Filter out unnecessary windows, for instance
 // nautilus desktop window.
-function getInterestingWindows(app, settings, monitorIndex) {
-    let windows = app.get_windows().filter(function(w) {
+function getInterestingWindows(app, settings, monitorIndex, location) {
+    let windows = getWindows(app, location).filter(function(w) {
         return !w.skip_taskbar;
     });
 
--- a/appIcons.js
+++ b/appIcons.js
@@ -29,7 +29,7 @@ const Util = imports.misc.util;
 const Workspace = imports.ui.workspace;
 
 const Me = imports.misc.extensionUtils.getCurrentExtension();
-const FileManager1API = Me.imports.fileManager1API;
+const Docking = Me.imports.docking;
 const Utils = Me.imports.utils;
 const WindowPreview = Me.imports.windowPreview;
 const AppIconIndicators = Me.imports.appIconIndicators;
@@ -134,11 +134,14 @@ var MyAppIcon = class DashToDock_AppIcon extends AppDisplay.AppIcon {
                 this._updateIndicatorStyle.bind(this)
             ]);
         }, this);
-        this._signalsHandler.add([
-            FileManager1API.fm1Client,
-            'windows-changed',
-            this.onWindowsChanged.bind(this)
-        ]);
+
+        if (this._location) {
+            this._signalsHandler.add([
+                Docking.DockManager.getDefault().fm1Client,
+                'windows-changed',
+                this.onWindowsChanged.bind(this)
+            ]);
+        }
 
         this._dtdSettings.connect('changed::scroll-action', () => {
             this._optionalScrollCycleWindows();
@@ -997,8 +1000,8 @@ const MyAppIconMenu = class DashToDock_MyAppIconMenu extends AppDisplay.AppIconM
 Signals.addSignalMethods(MyAppIconMenu.prototype);
 
 function getWindows(app, location) {
-    if (location != null) {
-        return FileManager1API.fm1Client.getWindows(location);
+    if (location != null && Docking.DockManager.getDefault().fm1Client) {
+        return Docking.DockManager.getDefault().fm1Client.getWindows(location);
     } else {
         return app.get_windows();
     }
--- a/dash.js
+++ b/dash.js
@@ -755,6 +755,7 @@ var MyDash = class DashToDock_MyDash {
             Array.prototype.push.apply(newApps, this._removables.getApps());
         } else if (this._removables) {
             this._signalsHandler.removeWithLabel('show-mounts');
+            this._removables.destroy();
             this._removables = null;
         }
 
@@ -769,6 +770,7 @@ var MyDash = class DashToDock_MyDash {
             newApps.push(this._trash.getApp());
         } else if (this._trash) {
             this._signalsHandler.removeWithLabel('show-trash');
+            this._trash.destroy();
             this._trash = null;
         }
 
--- a/dash.js
+++ b/dash.js
@@ -274,12 +274,6 @@ var MyDash = class DashToDock_MyDash {
 
         this._appSystem = Shell.AppSystem.get_default();
 
-        // Remove Drive Icons
-        this._removables = new Locations.Removables();
-
-        // Trash Icon
-        this._trash = new Locations.Trash();
-
         this._signalsHandler.add([
             this._appSystem,
             'installed-changed',
@@ -307,14 +301,6 @@ var MyDash = class DashToDock_MyDash {
             Main.overview,
             'item-drag-cancelled',
             this._onDragCancelled.bind(this)
-        ], [
-            this._trash,
-            'changed',
-            this._queueRedisplay.bind(this)
-        ], [
-            this._removables,
-            'changed',
-            this._queueRedisplay.bind(this)
         ]);
     }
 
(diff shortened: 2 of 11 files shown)


AsciiWolf:
Nice! I hope that this will be merged soon to fix #173.

corebots:
great! thanks!

beidl:
Regarding those 3 points you make, I have some questions to the first 2 points.
Writing here in the hope of getting the point across to GNOME upstream (especially the designers) sooner rather than too late for GNOME 4.

* Might be interesting if some upstream GNOME developers have different plans, especially since extending Wayland with custom protocols seems to be a thing now.

* Keeping trash/device separation intact seems to be a feasable goal AFAIK (preferrable for me since I initially made the bug report in the Unity days), though I could see Canonical/Ubuntu ditching their patches sooner rather than later. Which almost screams for a general protocol/API to be in place.

Is anybody else interested in getting those ideas back to upstream? I'd like to help in that case.

beidl:
Continuing:

I could see some difficulties (not getting the "full experience" on non-Ubuntu distros) with those options being turned on by default. I don't want to add insult to injury but would it make sense to keep those options turned off by default and be enabled by Ubuntu via gsettings overrides (in case they want to get the feature in quickly)?

dash.js
```diff
@@ -277,6 +278,12 @@ var MyDash = new Lang.Class({
 
         this._appSystem = Shell.AppSystem.get_default();
 
+        // Remove Drive Icons
+        this._removables = new Locations.Removables();
+
+        // Trash Icon
+        this._trash = new Locations.Trash();
```
franglais125: Would it be possible to create the `_trash` object only if enabled in the settings?

I actually disabled it a long time ago, and only today I realized why my system froze for a few seconds every time I deleted files: the call back to trash changes is still active!

philipl:
I've updated it to not create the apps  if the icons are hidden.

philipl:
@didrocks Do you have any interest in putting this functionality into the ubuntu dock?

didrocks:
@philipl: yes, once merged in Dash To Dock properly, I think we'll rebase on it and activate that feature right away. No window matching in Wayland (and I guess no Trash window matching either) would be great to see it fixed though.

didrocks:
Sounds excellent! I would prefer that you MR against nautilus and try to get some feedback from Carlos and other upstreams! You can maybe refer this discussions telling it's going to help dash to dock and others extensions matching windows, like ubuntu-dock?

Ofc, need to ensure that it doesn't regress for GNOME Shell dash in particular when you only have a nautilus desktop icon without any Trash (that it still matches nautilus).

philipl:
@didrocks https://gitlab.gnome.org/GNOME/nautilus/merge_requests/223

philipl:
@didrocks It's now merged into Nautilus. If you want to backport to 3.26, the patch is on my `ubuntu-3-26` branch in gitlab.

@micheleg I'd love to get this merged; is there anything you want me to do? any specific comments?

philipl:
@micheleg, do you have any feedback? As I said before, I'd like to get this merged, and if you can give me any guidance on that, I can work on it. Thanks!

khurshid-alam:
@philipl 
This actually broke the [xid patch for unity](https://git.launchpad.net/ubuntu/+source/nautilus/tree/debian/patches/19_unity_open_location_xid.patch?h=ubuntu/disco-devel). 

Can we still keep the xid patch along with this one ?

philipl:
@khurshid-alam, sorry, I don't know what you mean. The xid patch is for nautilus and nothing I do here can break it. Are you saying that window tracking doesn't work for you when you? I'm still running with the default ubuntu nautilus which only supports xid based tracking (vs the application id based tracking I got merged into upstream nautilus) and window tracking is working correctly for me here.

khurshid-alam:
@philipl 

Sorry I didn't explain myself well.

We plan to add that patch back into nautilus 3.30 in disco for unity dock. So I was concerned if it could interfere with application based tracking. But now that I checked again it doesn't. Another option is to use id based tracking in unity dock as well so that we don't need yet another dbus property. I well let Trevinho (it's his patch) to look into this. 

Thanks.

philipl:
The existing code is aware of both mechanisms and prefers the GTK based one over XID, which seems correct. A nautilus with both mechanisms would work equally well using both but you want to prefer GTK because when running under Wayland, the XID info would be absent or useless.

I agree that you'd more likely want to drop the XID mechanism and update the unity dock instead to avoid ambiguity.

AsciiWolf:
@jbicha Any chance this could be merged and added in Ubuntu (instead of the current desktop icons, just like on Unity)?

3v1n0:
I'm looking forward to do this next cycle

Il gio 18 apr 2019, 17:29 AsciiWolf <notifications@github.com> ha scritto:

philipl:
I’ll rebase onto master in the next few days. I’ve been waiting for 19.04 and gnome 3.32 so I can do that.

philipl:
I've now updated the patch set to work with latest master and squashed a bunch of the changes together.

3v1n0:
Hi @philipl, we're merging this in the ubuntu branch at least, although applying it together with https://github.com/micheleg/dash-to-dock/pull/975 so wondering the merge ordering.

In general I'm in favor for this to get it in master as well if @micheleg is also fine

philipl:
@3v1n0 Thanks. I think it probably makes sense to merge #975 first and then adjust this change accordingly (looks minor but probably a lot of conflicts due all the lines being changed everywhere). I'd also expect @micheleg to merge that one to master before this one (given how this PR has gone over the years) and so ordering that way will make it easier.

3v1n0:
Ok, this looks good to me...

I've applied the changes needed to get that working here, so fine to do this way.

locations.js
```diff
@@ -0,0 +1,252 @@
...
+ * This class maintains a Shell.App representing the Trash and keeps it
+ * up-to-date as the trash fills and is emptied over time.
+ */
+var Trash = class DashToDock_Trash {
+
+    constructor() {
+        this._file = Gio.file_new_for_uri('trash://');
+        this._monitor = this._file.monitor_directory(0, null);
+        this._lastEmpty = null;
+        this._empty = null;
```
3v1n0: I'd use booleans for this.

locations.js
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * This class maintains a Shell.App representing the Trash and keeps it
+ * up-to-date as the trash fills and is emptied over time.
+ */
+var Trash = class DashToDock_Trash {
+
+    constructor() {
+        this._file = Gio.file_new_for_uri('trash://');
+        this._monitor = this._file.monitor_directory(0, null);
```
3v1n0: File operations can fail, causing the whole dash stop working when some permissions aren't set, so should use more `try/catch` here

fileManager1API.js
```diff
@@ -0,0 +1,220 @@
...
+        this._locationMap = locationToWindow;
+        this.emit('windows-changed');
+    }
+});
+Signals.addSignalMethods(FileManager1Client.prototype);
+
+/**
+ * Construct a map of gtk application window object paths to MetaWindows.
+ */
+function getPathToWindow() {
```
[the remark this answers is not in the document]
philipl: I did this thinking about what would happen if you had multiple locations open. If I made getPathToWindow() take a path and search just that one window, then for multiple locations, it would have to repeat the search, and that looked more heavy-weight than searching through the map. So I figured, collect all the windows once and use that result to check each location.

locations.js
philipl: Fixed with your suggestion.

locations.js
philipl: Fixed with your suggestion.

dash.js
```diff
@@ -743,6 +744,34 @@ var MyDash = class DashToDock_MyDash {
...
 
+        if (this._dtdSettings.get_boolean('show-mounts')) {
+            if (!this._removables) {
+                this._removables = new Locations.Removables();
+                this._signalsHandler.addWithLabel('show-mounts',
+                    [ this._removables,
+                      'changed',
+                      this._queueRedisplay.bind(this) ]);
+            }
+            Array.prototype.push.apply(newApps, this._removables.getApps());
```
3v1n0: Why not just calling `newApps.push(this._removables.getApps())` ?

dash.js
```diff
@@ -743,6 +744,34 @@ var MyDash = class DashToDock_MyDash {
...
+        } else if (this._removables) {
+            this._signalsHandler.removeWithLabel('show-mounts');
+            this._removables = null;
```
3v1n0: You should also call `destroy()` before to ensure we disconnect from signals

dash.js
```diff
@@ -743,6 +744,34 @@ var MyDash = class DashToDock_MyDash {
...
+                this._trash = new Locations.Trash();
+                this._signalsHandler.addWithLabel('show-trash',
+                    [ this._trash,
+                      'changed',
+                      this._queueRedisplay.bind(this) ]);
+            }
+            newApps.push(this._trash.getApp());
+        } else if (this._trash) {
+            this._signalsHandler.removeWithLabel('show-trash');
+            this._trash = null;
```
3v1n0: Same here

fileManager1API.js
```diff
@@ -0,0 +1,218 @@
...
+            this._updateLocationMap.bind(this)
+        ], [
+            global.display,
+            'window-left-monitor',
+            this._updateLocationMap.bind(this)
+        ]);
+    }
+
+    destroy() {
+        this._signalsHandler.destroy();
```
3v1n0: Also add here a `this._monitor.run_dispose();` so that we are more explicit on destroying the underlying GObject.

fileManager1API.js
```diff
@@ -0,0 +1,218 @@
...
+ */
+function guessWindowXID(win) {
+    try {
+        return win.get_description().match(/0x[0-9a-f]+/)[0];
+    } catch (err) {
+        return null;
+    }
+}
+
+var fm1Client = new FileManager1Client();
```
3v1n0: Not in love with this, also because it's not destroyed when the extension is disabled, and when not needed.

So in general would be nice to make this part of `DockManager` and then destroy/recreate depending on the locations settings are updated.

I'll prepare something, although a better implementation might be done following what done in https://github.com/micheleg/dash-to-dock/pull/975/commits/4dcaa49ae51f6486b6fad2f9a170f867065d5c1e when that branch will land

locations.js
```diff
@@ -0,0 +1,264 @@
...
+
+    constructor() {
+        this._file = Gio.file_new_for_uri('trash://');
+        try {
+            this._monitor = this._file.monitor_directory(0, null);
+            this._signalId = this._monitor.connect(
+                'changed',
+                this._onTrashChange.bind(this));
+        } catch (e) {
+            log('Unable to monitor trash: ${e}');
```
3v1n0: Need to use '`' (backtick) or the `${e}` won't be expanded

locations.js
```diff
@@ -0,0 +1,264 @@
...
+        this._file.run_dispose();
+    }
+
+    _onTrashChange() {
+        try {
+            let children = this._file.enumerate_children('*', 0, null);
+            this._empty = children.next_file(null) == null;
+            children.close(null);
+        } catch (e) {
+            log('Unable to enumerate trash children: ${e}');
```
3v1n0: Same as before

appIcons.js
```diff
@@ -130,6 +134,11 @@ var MyAppIcon = class DashToDock_AppIcon extends AppDisplay.AppIcon {
                 this._updateIndicatorStyle.bind(this)
             ]);
         }, this);
+        this._signalsHandler.add([
```
3v1n0: I expect we need this connection only if `this._location` is true, isn't it?

philipl:
@3v1n0  Thanks for the additional comments. Any objections if I just roll your changes from the `ubuntu-dock` branch into this PR?

3v1n0:
@philipl sure that's fine!

Something I'd like to do for a next iteration is to make real deviceicons instead of re-using appIcons in all so that we can change things like DnD and right-click actions easier.

dash.js
philipl: At the time I wrote the original patch, the js version in ghome-shell didn't directly support calling push and this was the way you had to do it. I'm not surprised that it's directly callable now, but I wouldn't change it until it was clear what the minimum supported gnome-shell version is.

philipl:
> @philipl sure that's fine!
> 
> Something I'd like to do for a next iteration is to make real deviceicons instead of re-using appIcons in all so that we can change things like DnD and right-click actions easier.

Thanks. I agree on making proper device icons as a different class from appIcons. I wanted to get the base implementation in before working on more elaborate functionality that would be harder to rebase without the initial merge.

3v1n0:
Hey, you should probably rebase on https://github.com/3v1n0/dash-to-dock/tree/gnome-3-34 (or well you need this commit https://github.com/3v1n0/dash-to-dock/commit/3ff622dedc9193fba0e933a58b3765761d894de1) and then you should cherry-pick this commit https://github.com/micheleg/dash-to-dock/commit/cbb7ce75737ccd226c16ad0c18b6cedcd134b865 as or otherwise there might be errors enabling/disabling things which need the fm1 proxy

philipl:
@3v1n0 Thanks for the heads up, but I'm now somewhat struggling to decide how to maintain this PR. If it depends on the gnome-3-34 work, which isn't merged, then it becomes difficult to actually have a PR against master at all. Can I just cherry-pick the two commits into PR? It's not really mergeable in that form, but at least it would work.

3v1n0:
You can probably rebase on gnome-3-34 and that's it, not sure if @micheleg is able to merge that one and this, or I should I.

In any case, this is already part of [ubuntu dock branch](https://github.com/micheleg/dash-to-dock/commits/ubuntu-dock) that if you want you can also target to.

philipl:
i've added the two changes you recommended. Honestly, the main reason I don't want to rebase onto 3.34 is I want my primary branch to be one I can use and until I switch to 19.10, I won't have 3.34 :-)

mohegan:
Great job !

It tested it and it works great for me. But, can you add an option to only hide google drive, please ?
I have too many icons in the dock because I have several gmail accounts.

Can I create a fork of dash-to-dock with this request ? Maybe some people can test it.

mohegan:
Maybe a very little bug in **Settings.ui**  : I think 
`<property name="label" translatable="yes">Show trash can</property>`
must be :
`<property name="label" translatable="yes">Show trash icon</property>`

mohegan:
For French translation, just add this lines in the file **po/fr.po** :
`#: Settings.ui.h
msgid "Show trash can"
msgstr "Afficher la corbeille"

#: Settings.ui.h
msgid "Show mounted volumes and devices"
msgstr "Afficher les volumes et périphériques montés"`

mohegan:
Can you add the possibility to translate the trash icon name and volume actions (_Trash_ and _Unmount_ for examples) ?

For now, it isn't easily translatable (in the locations.js file, I think).

3v1n0:
> It tested it and it works great for me. But, can you add an option to only hide google drive, please ?

Maybe having a setting allow to filter drivers by protocol/regex could be nice

mohegan:
With bad code, I modified the file **locations.js** to :
```
    _onVolumeAdded(monitor, volume) {
        if (!volume.can_mount()) {
            return;
        }

        let activationRoot = volume.get_activation_root();
        if (!activationRoot) {
            // Can't offer to mount a device if we don't know
            // where to mount it.
            // These devices are usually ejectable so you
            // don't normally unmount them anyway.
            return;
        }
        let uri = GLib.uri_unescape_string(activationRoot.get_uri(), null);

        let volumeKeys = new GLib.KeyFile();
        volumeKeys.set_string('Desktop Entry', 'Name', volume.get_name());
        volumeKeys.set_string('Desktop Entry', 'Icon', this._getWorkingIconName(volume.get_icon()));
        volumeKeys.set_string('Desktop Entry', 'Type', 'Application');
        volumeKeys.set_string('Desktop Entry', 'Exec', 'nautilus "' + uri + '"');
        volumeKeys.set_string('Desktop Entry', 'StartupNotify', 'false');
        volumeKeys.set_string('Desktop Entry', 'Actions', 'mount;');
        volumeKeys.set_string('Desktop Action mount', 'Name', __('Mount'));
        volumeKeys.set_string('Desktop Action mount', 'Exec', 'gio mount "' + uri + '"');
        let volumeAppInfo = Gio.DesktopAppInfo.new_from_keyfile(volumeKeys);
        let volumeApp = new Shell.App({appInfo: volumeAppInfo});
        this._volumeApps.push(volumeApp);
        this.emit('changed');
    }
```
to :
```
    _onVolumeAdded(monitor, volume) {
        if (!volume.can_mount() || volume.get_name().search("gmail") > 0) {
            return;
        }

        let activationRoot = volume.get_activation_root();
        if (!activationRoot) {
            // Can't offer to mount a device if we don't know
            // where to mount it.
            // These devices are usually ejectable so you
            // don't normally unmount them anyway.
            return;
        }
        let uri = GLib.uri_unescape_string(activationRoot.get_uri(), null);

        let volumeKeys = new GLib.KeyFile();
        volumeKeys.set_string('Desktop Entry', 'Name', volume.get_name());
        volumeKeys.set_string('Desktop Entry', 'Icon', this._getWorkingIconName(volume.get_icon()));
        volumeKeys.set_string('Desktop Entry', 'Type', 'Application');
        volumeKeys.set_string('Desktop Entry', 'Exec', 'nautilus "' + uri + '"');
        volumeKeys.set_string('Desktop Entry', 'StartupNotify', 'false');
        volumeKeys.set_string('Desktop Entry', 'Actions', 'mount;');
        volumeKeys.set_string('Desktop Action mount', 'Name', __('Mount'));
        volumeKeys.set_string('Desktop Action mount', 'Exec', 'gio mount "' + uri + '"');
        let volumeAppInfo = Gio.DesktopAppInfo.new_from_keyfile(volumeKeys);
        let volumeApp = new Shell.App({appInfo: volumeAppInfo});
        this._volumeApps.push(volumeApp);
        this.emit('changed');
    }
```

philipl:
So, what we should be doing is excluding mounts that are identified as 'remote' - ie: accessed over the network. This is a gvfs filesystem attribute that should be set. There is no gmail backend in gvfs itself so I don't know where it's coming from, but it should be setting this attribute. If it does, it should already be excluded.

mohegan:
In **locations.js**, I think you can replace Nautilus by gio open (line 176). Replace :
`volumeKeys.set_string('Desktop Entry', 'Exec', 'nautilus "' + uri + '"');`
by :
`volumeKeys.set_string('Desktop Entry', 'Exec', 'gio open "' + uri + '"');`

philipl:
I'm going to leave this based on the 3.32 branch until I have upgraded to 19.10 and can use 3.34. Otherwise I can't actually test it.

I've fixed the `nautilus` vs `gio open`. Not sure how I missed that.

philipl:
> Can you add the possibility to translate the trash icon name and volume actions (_Trash_ and _Unmount_ for examples) ?
> 
> For now, it isn't easily translatable (in the locations.js file, I think).

What makes you say that? The strings all use `__("value")` which makes them translatable, does it not? This is used elsewhere in dash to dock.

philipl:
> So, what we should be doing is excluding mounts that are identified as 'remote' - ie: accessed over the network. This is a gvfs filesystem attribute that should be set. There is no gmail backend in gvfs itself so I don't know where it's coming from, but it should be setting this attribute. If it does, it should already be excluded.

If this is the `goa` backend in gvfs, that sets its class identifier to network, which is the check that I do to filter out volumes. So not sure why that doesn't work for you.

micheleg:
I've merged into master what was currently in the ubuntu-dock branch [80928137b], as I want to make a release for 3.34. 

I see further discusson here and I see also that @philipl you are working on a 3.32 based branch. I'm not sure how you want to evolve this PR now. I do not want to add features for older releases as I would not be able to support them (I might make exceptions for versions currently in ubuntu LTSs and Debian stable).

philipl:
@micheleg Thanks for merging! I'm shortly going to upgrade myself so I'm on 3.34, but in practice, there's only been a single line change on top of what you merged and I can open a separate PR for that.

You can merge this PR to the 3.32 branch if you think the you want the feature there, but otherwise we can close it.

3v1n0:
I've plans for this and even for devices, but not sure if I'll have time this cycle.

philipl:
@3v1n0 Can you close this PR in a way that indicates it was merged in the end? Due to how the merging was done, github isn't smart enough to notice.

3v1n0:
Looks like I can't do that as per https://github.com/isaacs/github/issues/2 but well, we can close this.

Merged as https://github.com/micheleg/dash-to-dock/commit/80928137b61ca94edab5ae1a0a676fb18b8057e7

PR #687, opened 2018-02-09
theming: explicitly update solid style on showing the overview

beidl:
Properly update top bar and dock styling with maximized windows in the overview when transparency-mode set is set to ADAPTIVE

### DIFF
--- a/theming.js
+++ b/theming.js
@@ -376,9 +376,13 @@ const Transparency = new Lang.Class({
             global.window_manager,
             'switch-workspace',
             Lang.bind(this, this._updateSolidStyle)
-        ],[
+        ], [
             Main.overview,
             'hiding',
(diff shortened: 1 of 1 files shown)


beidl:
Guess I have to explicitly mark a few Ubuntu guys? @didrocks

beidl:
"s/mark/tag/g" rather, don't want to give you the wrong idea :)

didrocks:
Hey! Sorry for not being reponsive earlier on (I was off).
Is this change in https://github.com/micheleg/dash-to-dock/commits/master? We only do regular rebase and not add features that aren't in dashtodock itself. I can't find a corresponding commit there. (And so, no need to then do a separate/duplicated PR).

beidl:
Created new pull request #706 as I'm not actively using Ubuntu with the Ubuntu session at the moment.

PR #703, opened 2018-03-09
http to https and a url fix

GitHaarek:
There are still 2 remaining http urls to be found in README.md that integrate the Flattr button. The button is broken, but you may want to remove it anyway and instead ask your visitors to use the Flattr extension which now seems to be the recommended method: [https://blog.flattr.com/2017/10/the-flattr-relaunch-what-to-expect/](https://blog.flattr.com/2017/10/the-flattr-relaunch-what-to-expect/)


micheleg:
Thanks (merged as [24f96f78ee2]). Yes, I should update the flattr button, haven't had time to investigate what's the best action to do.

PR #711, opened 2018-03-23
Add clickAction minimize-or-previews

veridiam:
**Needs translations.**

Adds a new clickAction, which minimized a single window, and shows a preview if there more than one (similar to Windows task bar).

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -84,6 +84,7 @@
                           <item translatable="yes">Cycle through windows</item>
                           <item translatable="yes">Minimize or overview</item>
                           <item translatable="yes">Show window previews</item>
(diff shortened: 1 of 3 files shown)


micheleg:
Thank you! I've merged your contribution in [b2e9bb7ca2d92f]. We are adding billions of behaviours. which we might one day reconsider... but the implementation is neat and self-contained - and the feature makes sense - so good for now.

peci1:
I'm sorry for a newbie question, but how can I get it on ubuntu bionic? I enabled DashToDock, chose this option, but the system doesn't seem to support is, because checking org.gnome.shell.extensions.dash-to-dock.click-action in dconf I see it gets set to default instead of minimize-of-previews.

Thanks for any advice

PR #719, opened 2018-04-05
Fix GtkCheckButton and GtkLabel allignement in Launchers tab.

squall77it:


micheleg:
Hi Gianni. Sorry for the late reply. I'm not sure I understand the goal of this patch. On my system, it results on the misalignment of the last checkbox ("Animate Show Applications"):

image

But this could be due to a different Gtk version or different theme from what you are using. Can you confirm your intended goal (a screenshot would help).

squall77it:
Hi, I use debian testing with gnome 3.28, you are right the problem is the theme i use.
With the adwaita default theme and arc-darc there is no misalignment

spazio di lavoro 1_012
spazio di lavoro 1_013

but with some other like Adapta-Nokto or Canta theme there is some misalignment

spazio di lavoro 1_010
spazio di lavoro 1_011

You can close this pull request because my patch work only with canta theme, sorry.

micheleg:
MIght still be some minor bug in our code exposed by that theme, but closing this for now. Thanks anyway!

PR #731, opened 2018-04-30
Feature: "Cycle + Minimize" windows

### LINKED ISSUE #721: Feature request: new action on click
I like "Dash to Panel" for it's on icon click action "Activate + Collapse", but I want to have something like it on the left side of screen. This action means:
 - Navigate over instances of the clicked application by some order
 - If all application's windows are visible, then hide all it

At least it would be good to just hide application window when it is visible and I click on it's icon on dash.

Tarik02:
Closes #721


05storm26:
@micheleg Could this be merged?

PR #733, opened 2018-05-02
Update README.md

Piterden:


micheleg:
Hi Denis, why do you think the git/ssh address is better than the https one?

Piterden:
Cause of with https the certificate doesn't work and you would need to enter login/pass each time.

micheleg:
Yes, I agree with that, but:

* That only applies to the case you are pushing to the repository. Users have no write permission, so they never have to write a password anyway. 
* Even if that was not the case, I would assume anybody having set up an ssh certificate would find out by himself about that option. 

From a quick search online, the main argument pro https seem to be:
* Avoiding potential firewall issues

Piterden:
But I had pushed, exactly)))

3v1n0:
There's no point for this.

Issue #671, opened 2018-01-14
Dash-to-dock on Ubuntu 17.10 : right click on Nautilus icon should show its bookmarks


Coeur-Noir:
Hi,

coming from here https://bugs.launchpad.net/ubuntu/+source/gnome-shell-extension-ubuntu-dock/+bug/1730201 there https://forum.ubuntu-fr.org/viewtopic.php?pid=21853368#p21853368 and https://bugzilla.gnome.org/show_bug.cgi?id=792483

On Ubuntu 17.10, bookmarks - default ones like documents, videos, pictures… or manually added ones through ctrl+d - are not displayed on right click on nautilus icon in the dock.

Only [ new window | remove from favorites | details ] appear.

I expected all the bookmarks to be displayed on right click, like in previous Ubuntu versions, see https://bug792483.bugzilla-attachments.gnome.org/attachment.cgi?id=366750

Maestroschan:
i tried that https://i.imgur.com/L8m6h6K.png as a standalone extension (very ugly code here https://github.com/Maestroschan/quicklist-gnome-extension) but it was incompatible with the "Ubuntu dock" for some reason (but not dash to dock...).

micheleg:
I think it depends on the settings: we (dash-to-dock and ubuntu-dock) reuse the popupmenu if no further customization are requested , and therefore show any customization provided by other extensions. I guess that if you disable the windows previews you will see the bookmarks. Full compatibility would be best achieved  implementing these quicklist actions in dash to dock directly.

frenchy82:
Having bookmarks shortcut on the nautilus icon must be an indispensable feature for à dock.
This will avoid installing an extension for shortcuts

Thanks

PR #758, opened 2018-06-19
Option for showing bookmarks or recent files in appicons menu

maoschanz:
related to issue #671 

i didn't find a .pot file so i didn't try to add new strings to translation-related files

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -36,7 +36,7 @@ const AppIconIndicators = Me.imports.appIconIndicators;
 
 let tracker = Shell.WindowTracker.get_default();
 
-let RecentManager = new Gtk.RecentManager();
+let recentManager = new Gtk.RecentManager();
 
 let DASH_ITEM_LABEL_SHOW_TIME = Dash.DASH_ITEM_LABEL_SHOW_TIME;
 
@@ -62,6 +62,10 @@ let recentlyClickedApp = null;
 let recentlyClickedAppWindows = null;
 let recentlyClickedAppIndex = 0;
 let recentlyClickedAppMonitor = -1;
+
+function sortByDate(x,y) {
+    return y.get_modified() - x.get_modified();
+}
 
 /**
  * Extend AppIcon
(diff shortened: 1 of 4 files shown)


micheleg:
Hi,

thanks for the contribution and sorry for the late reply. I quickly tested this feature and seems to be working just fine. I haven't thought about it enough, but it seems to be adding a reasonable funcitonality in a reasonably uninstrusive way.

I'm jsut a bit unconvinced about the bookmarks:
image

it's not obvous to me what those icons are and I'm not convinced bookmarks should be split  between icons and list. 

I've also noticed a little bug: the menu is not closed when you click on one of those icon, while it is when clicking on the bookmark in the list (like any other menu item).

Kinf of provocative question: is the bookmark menu any faster/more direct/more logical than just opening nautilus and using the sidebar?

Maestroschan:
>it's not obvous to me what those icons are and I'm not convinced bookmarks should be split between icons and list.

These 3 are "recent files", "trash" and "other locations", and since:
- they are not actually bookmarks;
- they are probably among the most useful items

... I didn't want to put them in the same submenu. But i agree the "other locations" icon is far from perfect

>I've also noticed a little bug: the menu is not closed when you click on one of those icon, while it is when clicking on the bookmark in the list (like any other menu item).

Exact, i'll fix that

>is the bookmark menu any faster/more direct/more logical than just opening nautilus and using the sidebar?

Faster: if Nautilus is already launched, clicking on the sidebar is faster, but otherwise it's not

More direct/logical: yes, imo at least, and it's what Unity and Windows Vista to 10 are doing (no idea what macOS does)

appIcons.js
```diff
@@ -36,6 +36,8 @@ const AppIconIndicators = Me.imports.appIconIndicators;
 
 let tracker = Shell.WindowTracker.get_default();
 
+let RecentManager = new Gtk.RecentManager();
+
```
micheleg: should be `recentManager` for consistency I believe.

micheleg:
> Faster: if Nautilus is already launched, clicking on the sidebar is faster, but otherwise it's not

I will try yout patches in the next few days. I'm curious to see how many times I will end up using those shortcuts :)

Maestroschan:
but it's not as fast as "Places status indicator" of course

maoschanz:
good news is that i fixed the issues, bad news is that i needed a working version so i updated the extension before fixing the issues, and it was a terrible idea since now there is a ton of conflicts (i only committed appIcons.js to avoid useless ones)

PR #764, opened 2018-06-28
appIcons: Use `.format` to allow proper translations

azzar1:


micheleg:
I think we did that way to to try to reuse existing translations, but yes this sound the riht thing to do. Merged as [842ba2726680]

PR #770, opened 2018-07-08
use a wrapper to access to Screen, Workspacess and Monitor

3v1n0:
As per [mutter xwayland-on-demand preparation branch refactory](https://gitlab.gnome.org/GNOME/mutter/merge_requests/95),
MetaScreen has been removed, thus replacing the dead objects.

Basically global.screen moved to global.display, while all the
workspace related features have been moved to the WorkspaceManager
and displays are managed by DisplayManager.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -114,7 +114,7 @@ var MyAppIcon = new Lang.Class({
             Main.layoutManager.monitors.length > 1) {
             this._signalsHandler.removeWithLabel('isolate-monitors');
             this._signalsHandler.addWithLabel('isolate-monitors', [
-                global.screen,
+                Utils.DisplayWrapper.getScreen(),
                 'window-entered-monitor',
                 Lang.bind(this, this._onWindowEntered)
             ]);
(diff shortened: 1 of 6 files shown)


3v1n0:
As a note, probably this should come with a version check on 2.30 (2.29).

Or better, we could just make the code abstract to work with both versions, I'm open to that, let me know what you prefer.

micheleg:
Hi Marco, sorry for the late reply and thatk you for the contribution. 

My understanding is that this is compatible only with the upcoming gnome-shell 3.30. I'll need to consider dropping support for older versions at a certain point, but I'd like to keep tha main branch compatible al least with the current Ubuntu LTS and possibly Debian stable (which I'm running daily). 

As it seems that the API haven't actualy changed but just being moved around, I would rahter make the code abstract to work with versions for now, with an explicit comment reminding to drop the upstraction in the future. At the moment we are supporting with the same codebase way back to 3.18. I would consider targetig 3.22+ if that's of any help.

3v1n0:
Yeah, I quite agree with that... As it wouldn't be too much complexity as instead maintaining different branches.

I'll rework on that soon.
Cheers

3v1n0:
So, @micheleg, i've refactored the change a bit to use the `DisplayWrapper` namespace (not really needed to have a class here) to redirect the calls to the proper object.

It works fine here in both gnome-shell 3.28 and 3.29.4 (and git).

micheleg:
Merged as  [8398d41c] with just a couple of modifications to maintain compatibility also with pre ES6 gjs versions (e.g gnome-shell 3.22).

Thank you!

Issue #781, opened 2018-08-02
Cannot change shortcut <Super>q


kebeda:
The text entry box in the settings screen, behavior tab, "Use keyboard shortcuts to activate apps", "Show dock and application numbers" pop-up, "Shortcut for the options above" doesn't appear to be enabled. Although I can highlight the text entry box with my mouse, I cannot change the text `"<Super>q"` in any way. Tabbing through the dialog seems to skip this text entry box altogether (I can tab/modify the "Hide timeout(s)" box right below it however). Please fix this, as I'd like to use something other than `"<Super>q"` for this function. Alternatively, a gsettings command would be helpful.

kebeda:
Here is a gsettings that will alter it (to Alt-d, in this case):

```
gsettings --schemadir ~/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/schemas/ set org.gnome.shell.extensions.dash-to-dock shortcut "['<Alt>d']"
gsettings --schemadir ~/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/schemas/ set org.gnome.shell.extensions.dash-to-dock shortcut-text '<Alt>d'

```
I would still like the settings dialog, fixed however.

kebeda:
Thank you, Fran, I applied PR!782 to my machine and the text entry box now works as it should!

micheleg:
Closed by #782 and and fix merged [082a76ca3c].

PR #782, opened 2018-08-03
Settings.ui: Fix GtkEntry property to make the shortcut editable.

franglais125:
Should fix https://github.com/micheleg/dash-to-dock/issues/781.

The issue was introduced with the update of Settings.ui to Glade 3.22 [7afffc7c2916b54a0dee9fbdb92a39518be731e2].


kebeda:
Thank you, Fran, I applied this change to my machine and the text entry box now works as it should!

micheleg:
Merged [082a76ca3c]. Thank you!

PR #825, opened 2018-10-25
dash, docking: remove Shell.GenericContainer

3v1n0:
[It has been removed upstream](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/153), and extending components instead of compositing them with the delegate pattern is suggested.

We still might use composition by using fake St.Widgets that replicate the same locally, but I guess it's better to follow what upstream does.

Not sure if this means breaking old compatibility though.

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -54,6 +54,7 @@ function extendDashItemContainer(dashItemContainer, settings) {
  */
 const MyDashActor = new Lang.Class({
     Name: 'DashToDock.MyDashActor',
+    Extends: St.Widget,
 
     _init: function(settings) {
         // a prefix is required to avoid conflicting with the parent class variable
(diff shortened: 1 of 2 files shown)


micheleg:
Thanks for the PR. I don't think this this will break anything in prior versions, as this is what I used to do long time ago. The reason I moved away (41a315cd5ac142a) was to avoid errors when live updating the extension (I was flooded with bug reports at each new version release on the extension website).

micheleg:
On my system (Debian stable, Gnome 3.22 this makes the shell crash. The problem seems to be related to the oject name, as ''.' seems not to be allowed. See https://github.com/GNOME/glib/blob/master/gobject/gtype.c#L747

I assume it's working somehow in newer Gnome versions or you would have encountered the issues.  I renamed  the classes (DashToDock.DashSlideContainer and DashToDock.MyDashActor) and merged as [421efeed7fb]

PR #838, opened 2018-11-15
docking: Fix leaking signal connection 

azzar1:
this.dash has a longer life-time that this._box so we need to manually
disconnect the signal connection to avoid invalid access to already destroyed
variables.

Fixes: #839

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -349,6 +349,13 @@ const DockedDash = new Lang.Class({
             Lang.bind(this, function() {
                 Main.overview.dashIconSize = this.dash.iconSize;
             })
+        ], [
+            // sync hover after a popupmenu is closed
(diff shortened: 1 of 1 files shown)


azzar1:
I removed the first commit because I moved it to: https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/294

3v1n0:
Better continuing using `Lang.bind` for retrocompatibility

azzar1:
> Better continuing using Lang.bind for retrocompatibility

Done

3v1n0:
LGTM, if @micheleg is fine, I can go forward and merge it.

micheleg:
Great! Thanks.

PR #843, opened 2018-11-20
 extension: Ensure signal disconnection 

azzar1:
Make sure we disconnect from the 'extension-state-changed' signal in the disable
function even if dockManager failed to destroy properly. If we fail to
disconnect from that signal, conditionallyenabledock will be called when the
next extension is disabled, and ubuntu-dock will re-enable itself.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -21,20 +21,24 @@ function enable() {
      * Listen to enabled extension, if Dash to Dock is on the list or become active,
      * we disable this dock.
      */
-    dockManager=null; // even if declared, we need to initialize it to not trigger a referenceError.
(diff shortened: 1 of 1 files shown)


extension.js
```diff
@@ -21,31 +21,31 @@ function enable() {
      * Listen to enabled extension, if Dash to Dock is on the list or become active,
      * we disable this dock.
      */
-    dockManager=null; // even if declared, we need to initialize it to not trigger a referenceError.
```
didrocks: Are you sure this isn't needed anymore?

azzar1: Yes, I don't get any warning (I tested it with and without dash-to-dock enabled).

didrocks:
Ack, let's merge then, thanks!

PR #848, opened 2018-11-30
SRU fix for LP: 1769383

azzar1:


logix2:
I applied this patch (well, this one which is the same: https://github.com/micheleg/dash-to-dock/pull/843/files) two days ago on my Ubuntu 18.10 system, and I even posted an article on my blog about the fix, but then last night the bug with Ubuntu Dock shown on the lock screen happened again, despite applying the patch.

PR #859, opened 2018-12-14
Avoid repainting an unchanging dock.

vanvugt:
Since we are usually visible but not usually changing, make sure
most repaint requests don't actually require us to repaint anything.

This eliminates about a third of gnome-shell's CPU usage when a window
redraws while touching the dock.

https://launchpad.net/bugs/1743976

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -74,6 +74,11 @@ const MyDashActor = new Lang.Class({
             layout_manager: layout,
             clip_to_allocation: true
         });
+
+        // Since we are usually visible but not usually changing, make sure
+        // most repaint requests don't actually require us to repaint anything.
+        // This saves significant CPU when repainting the screen.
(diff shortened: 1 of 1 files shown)


vanvugt:
gnome-shell CPU usage on an i7-7700:

Before:
Small window in the middle of the screen: 9%
Small window touching the dock: 14%
Maximized window touching the dock: 16%

After:
Small window in the middle of the screen: 9%
Small window touching the dock: **9%**
Maximized window touching the dock: **11%**

vanvugt:
Also, [the same fix is already upstream](https://gitlab.gnome.org/GNOME/gnome-shell/commit/f77b3da74fccfd8baa160e26a0ebf51c2baa5f83).

micheleg:
Thanks! Merged.

Issue #641, opened 2017-11-12
Dock shows on lockscreen on Ubuntu 17.10


1989gironimo:
Referring to this bugreport: https://github.com/micheleg/dash-to-dock/issues/569#issuecomment-343710348 - It's pretty much the same but in August that was already fixed.

It's not always the case but most of the time the dock is visible on the left bottom side if the lockscreen is active.

franglais125:
Could you check `sudo journalctl /usr/bin/gnome-shell` and post the output whenever this happens?

ernstp:
Could it be a conflict with ubuntu-dock? They're using the same variable name?
I have both extensions installed but Ubuntu Dock disabled. Or supposed to be disabled.

nov 15 13:30:03 semall0566 gnome-shell[5623]: JS WARNING: [/home/ernsjo/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/utils.js 34]: reference to undefined property Symbol.toPrimitive
nov 15 13:30:03 semall0566 gnome-shell[5623]: JS WARNING: [/usr/share/gnome-shell/extensions/ubuntu-dock@ubuntu.com/appIcons.js 1361]: unreachable code after return statement
nov 15 13:30:03 semall0566 gnome-shell[5623]: JS WARNING: [/usr/share/gnome-shell/extensions/ubuntu-dock@ubuntu.com/extension.js 44]: assignment to undeclared variable runningExtensions
nov 15 13:30:28 semall0566 gnome-shell[5623]: Extension "ubuntu-dock@ubuntu.com" had error: TypeError: dockManager is null

ernstp:
Yeah, uninstalling gnome-shell-extension-ubuntu-dock solved the issue for me.

1989gironimo:
Have done that (like the posts before) unintentionally and cannot reproduce that error anymore.
If it's coming back, I'll report.

1989gironimo:
Ok - had it now.
Had to copy the command from thunderbird, so there may be unnecessary stuff related to that:

`Nov 19 11:27:51 robot gnome-shell[2245]: Failed to apply DRM plane transform 0: Keine Berechtigung
Nov 19 11:27:51 robot org.gnome.Shell.desktop[2245]: Window manager warning: Failed to set power save mode for output LVDS-1: Keine Berechtigung
[...]

Issue #820, opened 2018-10-16
dash-to-dock crashes randomly (possibly when opening lid?)


rusins:
After upgrading to Gnome 3.30 (now on 3.30.1), dash-to-dock seems to crash every now and then. Have been trying to monitor when it happens, and from the attached journalctl log it appears opening the lid of my laptop triggered it. (Possibly related: I once locked my screen but still had the dock showing above the lock-screen. Moving my mouse off the dock hid it, and thankfully it didn't show up again when I moved my mouse back. I.e., stayed disabled like it should be, until I unlocked my session.)

Will try to catch it crashing again, to see if it really is the laptop lid / desktop locking triggering it. Sometimes all my extensions get disabled, but in this instance only dash-to-dock stopped working, so I believe it is the culprit.

```
10月 16 01:48:57 tanks systemd-logind[600]: Lid opened.
10月 16 01:48:59 tanks gnome-shell[1121]: ../glib/gobject/gsignal.c:2641: instance '0x562aa3960bd0' has no handler with id '351609'
10月 16 01:48:59 tanks gnome-shell[1121]: ../glib/gobject/gsignal.c:2641: instance '0x562aa3960bd0' has no handler with id '351610'
10月 16 01:48:59 tanks org.gnome.Shell.desktop[1121]: == Stack trace for context 0x562aa1382220 ==
10月 16 01:48:59 tanks org.gnome.Shell.desktop[1121]: #0   562ab6da9e30 i   resource:///org/gnome/gjs/modules/overrides/GObject.js:468 (7f98cdacc040 @ 25)
10月 16 01:48:59 tanks org.gnome.Shell.desktop[1121]: #1   562ab6d7e098 i   /home/raitis/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/theming.js:416 (7f9893ab1940 @ 20)
10月 16 01:48:59 tanks org.gnome.Shell.desktop[1121]: #2   7fff3006dfe0 b   self-hosted:261 (7f98cdac1dc0 @ 223)
10月 16 01:48:59 tanks org.gnome.Shell.desktop[1121]: #3   562ab6d7dfe0 i   /home/raitis/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/theming.js:415 (7f9893ab18b0 @ 172)
[...]

PR #868, opened 2019-01-15
theming: Ensure _trackingWindows contains valid windows

azzar1:
Populate _trackedWindows using global.window_group.get_children() instead of
global.get_window_actors() to guarantee a window is removed from the
_trackedWindows map when the global.window_group 'actor-removed' signal is
received. Without this guarantee we might keep a reference to a destoyed window
actor, causing troubles when trying to disable the extension.

Fixes: https://github.com/micheleg/dash-to-dock/issues/641

### DIFF
--- a/theming.js
+++ b/theming.js
@@ -388,11 +388,13 @@ const Transparency = new Lang.Class({
         ]);
 
         // Window signals
-        global.get_window_actors().forEach(function(win) {
+        global.window_group.get_children().filter(function(child) {
             // An irrelevant window actor ('Gnome-shell') produces an error when the signals are
             // disconnected, therefore do not add signals to it.
-            if (win.get_meta_window().get_wm_class() !== 'Gnome-shell')
-                this._onWindowActorAdded(null, win);
+            return child instanceof Meta.WindowActor &&
+                   child.get_meta_window().get_wm_class() !== 'Gnome-shell';
+        }).forEach(function(win) {
+            this._onWindowActorAdded(null, win);
         }, this);
 
         if (this._settings.get_enum('transparency-mode') === TransparencyMode.ADAPTIVE)


3v1n0:
The change looks good, however is there any advantage in using two `filter` functions?

Why not doing the class check in the first filter too?
As it would cause two loops otherwise.

azzar1:
> The change looks good, however is there any advantage in using two filter functions?

IMHO filter functions are more readable if the filters are separated. Performances are not a concern because the list of windows is small and this function is seldom called. Btw I can change it if you're strongly against having two filter functions.

azzar1:
Joined the two filter functions!

3v1n0:
Wasn't strongly against, but imho wasn't too readable either as because you need to remember you're filtering again and the syntax doesn't help with that (as you're still in the same indentation level as the previous one).

Anyway looks good to me.

micheleg:
Thank's for the patch. Happy to merge, it logically makes sense to pick the initial windows this way. Can you elaborate on what's going on. Were we adding signals to spurious actors? Was it some sort of race condition?

Also,  the check on the "Gnome-shell" wm_class could be irrelevent now, but maybe better leave it there for now.

azzar1:
> Can you elaborate on what's going on. Were we adding signals to spurious actors? Was it some sort of race condition?

Issue #641 is happening because in **Transparency::disable** we are accessing an already destroyed window. In particular:
```js
for (let key of this._trackedWindows.keys())
    this._trackedWindows.get(key).forEach(id => {
        key.disconnect(id);
    });
```
This error is trigger an exception and dash-to-dock fails to disable itself completely. 

**_trackedWindows** is populated in two ways:

- In **Transparency::enable**, using **global.get_window_actors()**
```js
        // Window signals
        global.get_window_actors().forEach(function(win) {
            // An irrelevant window actor ('Gnome-shell') produces an error when the signals are
            // disconnected, therefore do not add signals to it.
            if (win.get_meta_window().get_wm_class() !== 'Gnome-shell')
                this._onWindowActorAdded(null, win);
        }, this);
```
- In renspose to the **actor-added** signal (emmittedy by **global.window_group**). Vice versa, a window is removed from the **_trackedWindows** map in renspose to  a **actor-removed** signal.
```js
       this._signalsHandler.addWithLabel('transparency', [
            global.window_group,
            'actor-added',
            Lang.bind(this, this._onWindowActorAdded)
        ], [
            global.window_group,
            'actor-removed',
            Lang.bind(this, this._onWindowActorRemoved)
        ], [
```
It happens that a window in **global.get_window_actors()** is not in **global.window_group.get_children()**. This means that when this window is destroyed, **global.window_group** will not emit a **actor-removed** signal and we'll not remove the window from the **_trackedWindows** map.

Why some windows are in **global.get_window_actors()** but not in **global.window_group.get_children()**? TBH I'm not sure, I didn't have time to dig into this too much.

nkestrel:
This should also fix #820.

lgtrombetta:
Can you please include this fix in a release of the gnome extension (not only for ubuntu users via ubuntu-dock)? The most recent is v64 from August 2018.

micheleg:
@lgtrombetta the version is under review (https://extensions.gnome.org/review/9189) but it might take some time to be accepted. You can grab the zip here https://micheleg.github.io/dash-to-dock/releases.html.

conualfy:
The bug is present again in Ubuntu 20.04 LTS and I see it as a security vulnerability, although this should not exist at all, Gnome Shell should not allow extensions to visibly and accessibly run when the screen is locked.


> @lgtrombetta the version is under review (https://extensions.gnome.org/review/9189) but it might take some time to be accepted. You can grab the zip here https://micheleg.github.io/dash-to-dock/releases.html.

atleta:
FYI, there are other extensions too, that can surface this issue. I've just played with KStatusNotificationItem support and Ubuntu AppIndicator for an hour to figure out how their interaction can cause this same thing to happen.

PR #873, opened 2019-01-20
Fix/border radius with dynamic opacity

### LINKED ISSUE #666: Adaptive and Dynamic opacity style issue
When using the default shell theme and the position of the dock is set to bottom and the opacity is set to either Adaptive or Dynamic the very bottom right corner of the dock is rounded when it should be squared to meet the edge of the screen. 
dock
EDIT: After posting the screenshot I noticed the top left corner is squared when should be rounded.

micheleg:
The transparency machinery is overwriting the inline style of the dock,
as is the code to adjust the border radius when the position of the dock
is not left.

Ensure we append the opaque/transparent style to any existing styles.
We move the computation of the transparency style in the transparency
enable stage, which is always called when restting the adjusted
border-radius styles.

Fix #666.

### DIFF
--- a/theming.js
+++ b/theming.js
@@ -352,8 +352,7 @@ const Transparency = new Lang.Class({
         this._opaqueAlphaBorder = '0.5';
         this._transparentTransition = '0ms';
         this._opaqueTransition = '0ms';
-
-        this._updateStyles();
+        this._base_actor_style = "";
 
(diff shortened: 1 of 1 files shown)


3v1n0:
Code wise looks fine to me, maybe you should remove the debugging log lines though

PR #874, opened 2019-01-21
Use CSS to style progress bar colors

clobrano:
Progress bar color is hardcoded in the source code.
This PR adds a dummy class to take `background-color` and `border-color` from a new `.progress-bar` CSS class and uses the values for both `fill` and `stroke` styling. 

Hope this is the right approach,
Thanks


appIconIndicators.js
```diff
@@ -858,8 +859,20 @@ const UnityIndicator = new Lang.Class({
...
+        let dummyObject = new St.Bin({
+            name: 'dashtodockContainer',
+        });
+        Main.uiGroup.add_child(dummyObject);
+
+        dummyObject.add_style_class_name('progress-bar');
+        let themeNode = dummyObject.get_theme_node();
+        let bg = themeNode.get_background_color();
+        let bd = themeNode.get_border_color(St.Side.BOTTOM);
+
```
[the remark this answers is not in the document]
micheleg: I think the workaround can be avoided alltogeteher by just applying the class to a real element (the St.DrawingArea) and get the style from there.

micheleg:
Sorry for the late review. I have left few comments on the code. I think we can do this in a more clean and direct way as per my last comment.

clobrano:
Hi @micheleg, no worries and thanks for the review, I'll propose a new change

appIconIndicators.js
```diff
@@ -858,8 +859,20 @@ const UnityIndicator = new Lang.Class({
         height -= 2.0*lineWidth;
 
         let finishedWidth = Math.ceil(this._progress * width);
-        stroke = Cairo.SolidPattern.createRGBA(0.8, 0.8, 0.8, 1.0);
-        fill = Cairo.SolidPattern.createRGBA(0.9, 0.9, 0.9, 1.0);
+
+        // Create dumy object to get progress-bar style
```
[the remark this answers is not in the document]
clobrano: ack

appIconIndicators.js
```diff
@@ -794,7 +794,19 @@ const UnityIndicator = new Lang.Class({
             return;
         }
 
-        this._progressOverlayArea = new St.DrawingArea({x_expand: true, y_expand: true});
+        this._progressOverlayArea = new St.DrawingArea({name: 'dashtodockContainer', x_expand: true, y_expand: true});
```
micheleg: There is no need for this to have 'dashtodockcontainer' as id/name. The parent (one of the dock container actor) has this id, and it should not be duplicated (although the shell probably won't complain). Instead, the style would be defined as `` #dashtodockContainer .progress-bar`` (added space), and it should work.

appIconIndicators.js
```diff
@@ -794,7 +794,19 @@ const UnityIndicator = new Lang.Class({
...
+
+        // Get progress-bar background and border color from CSS
+        this._progressOverlayArea.add_style_class_name('progress-bar');
+        Main.uiGroup.add_child(this._progressOverlayArea);
+
```
micheleg: What I had in mind, was to apply permanently the ` progress-bar` class directly to the `stDrawinArea`, but then we cannot use the standard css properties as they would be rendered on the widget itself. However, we could use custom style defintion (someting lik `-progress-bar-background', or some better naming...), and then read via `them_node.get_color('-progress-bar-background')`. This way we avoid all this machinery of adding/removing elements, and we make clear that it a special styling.

appIconIndicators.js
clobrano: Ok, I didn't know this and I wasn't able to make it work without the name

appIconIndicators.js
clobrano: That's better for sure. I didn't think about custom style definition

appIconIndicators.js
clobrano: It looks like that removing the `id/name` does not work. I tried first as you said, removing the name and adding the space between object and class in the style, but then shell complains it couldn't find the color property.

micheleg: mmm, sorry, I was thinking ahead. At the moment you add the element to the uiGroup when you retrieve the style, therefore it is not a children of th dock container at that point. However, if  as I was suggesting you permanently apply the class to the StDrawingArea and retrieve the style when that is added to the icon (which is a children of the dashtodockContainer element already), it should work.

clobrano: Oh, I see now, thanks for your patience

micheleg:
Seems good to me, thank you!

A few points:
1. With the current styling class, are you able to override the style from yout third party theme? I think this is the main motivation behind this PR.
2. If so, does the style get updated when the theme is changed, or it is necessary to reload the extension?

clobrano:
>     1. With the current styling class, are you able to override the style from yout third party theme? I think this is the main motivation behind this PR.

I am not 100% sure right now.
Short story: No, I wasn't able to change the style through Yaru css
Long(er) story: of course I tested the original dash to dock application, while Yaru affects the Ubuntu version, so I am not sure whether I can't change the style because it simply does not work or because Yaru does not affect the version I changed.

micheleg:
As far as I remember the extension theme has he higherst priority, so the theme settings will be discarded. Even `!important` statements are ignored I believe. That's why I have the `.dashtodock` class which is applied only when the user select "use built-in theme". This could be the approach for this feature too, but we might need to specify a default style in the code for when no style is provided.

clobrano:
Sorry for the late reply

> This could be the approach for this feature too, but we might need to specify a default style in the code for when no style is provided.

So if a style is shipped in dashtodock stylesheet there is no way to change it, right? The only way is to check if the style is provided by the built-in theme and provide a default one in the code itself?

micheleg:
@clobrano yes, to my best understaning we should

implement an hardcoded style (the curent one) in the case no css rules are povided by the theme and move the css style under the `.dashtodock` class (could be the same or could be somethig specifically fitting the defualt gnome shell theme). Tthis would also act as a placeholder and guide for custom themes.

clobrano:
Not ready yet for review :)

clobrano:
Ready now :+1:

micheleg:
Sorry for the late review. Looks good to me. I pushed a minor change to maintain the existing colors as default. Assuming I have not broken anything I'm happy to merge this.

clobrano:
Oh no, it's perfectly fine. I assumed that the color was lightgrey, but this is certainly more precise

clobrano:
Sorry to push :)
Is there anything left I can do for this to be merged?

micheleg:
Squashed and merged as [94becd3].
@clobrano Thanks!

PR #881, opened 2019-02-08
Rewrite code to use modern ES6 code and remove Lang features

3v1n0:
As per upstream changes on https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/361 GNOME-Shell doesn't support anymore `Lang.Class`'es that inherit from the new ES6 classes (while it works the other way around), while as per the `ShellGenericContainer` removal the extension doesn't support anymore versions before than 3.31.x.

So this is the moment for modernizing the code, while other versions can be branched.


metadata.json
```diff
@@ -1,10 +1,10 @@
...
-"shell-version": ["3.18", "3.20", "3.22", "3.24", "3.26", "3.28", "3.30"],
+"shell-version": ["3.31", "3.32"],
 "uuid": "dash-to-dock@micxgx.gmail.com",
 "name": "Dash to Dock",
 "description": "A dock for the Gnome Shell. This extension moves the dash out of the overview transforming it in a dock for an easier launching of applications and a faster switching between windows and desktops. Side and bottom placement options are available.",
 "original-author": "micxgx@gmail.com",
 "url": "https://micheleg.github.io/dash-to-dock/",
 "gettext-domain": "dashtodock",
-"version": 64
+"version": 65
```
micheleg: This was probably not intended.

micheleg:
Thanks a lot for the boring work of migrating the code! Haven't tested the code, but everything seems fine to me. 
As mentioned in the email, the only concern is if we can (consider) keeping the minimum supported version to 3.28 or 3.30 (as in, is this code actually working  in 3.28/3.30? I do not want to introduce ad-hocs changes if that's not the case)

jnsh:
I tested this on 3.31, and there's one very apparent issue: The background of the dock is 100% transparent regardless of dash-to-dock Appearance settings, or the shell theme being used.

3v1n0:
@micheleg 

Well, wasn't too boring, some regex work was enough :stuck_out_tongue_winking_eye: 

Anyways, 3.30 would be supported reverting https://github.com/micheleg/dash-to-dock/commit/421efeed7fb0582e7f4188325aa17d63e5f2b1b0 but without this 3.31 (3.32) won't work.
While 3.28 can't be supported anyways (as per gjs changes too).

So we can't really support all these versions with one single codebase anymore since 3.32 will be an effective shell API extensions break.

@jnsh 
I have adaptive enabled here and it seem to work fine, however I've noticed the same behavior even before this rewrite, so I'm quite sure that there could be theme changes needed too as per upstream changes, but I need to check

metadata.json
3v1n0: Well I think we need a release bump after this, but probably should be put in a different commit I guess, right?

jnsh:
@3v1n0 
Yes, apparently the background is visible, if "Customize opacity" is set to "Fixed", "Adaptive" or "Dynamic". With opacity set to "Default", it remains fully transparent. I have never observed this behavior with earlier versions myself.

micheleg:
@3v1n0 I do not think 421efee prevents support for older versions. I'm running current master happily on Debian stable (gnome-shell 3.22). But yes, let's focus on new versions.

metadata.json
micheleg: Yes, I usually prefer to bump the version at release.

3v1n0: Ok, done

micheleg:
Great! Merging.

dash.js
```diff
@@ -69,19 +68,20 @@ const MyDashActor = new Lang.Class({
             orientation: this._isHorizontal ? Clutter.Orientation.HORIZONTAL : Clutter.Orientation.VERTICAL
         });
 
-        this.parent({
+        super._init({
             name: 'dash',
             layout_manager: layout,
             clip_to_allocation: true
         });
+        this.name = this.constructor.name;
```
micheleg: @3v1n0 This seems to create some styling problems, as the widget loses thr `#Dash` id/name and things like the dash background are not set. Any reason why this might be necessary?

micheleg: From my tests and understanding it is not necessary (a4cb6420bf07e5639).

3v1n0: Thanks for catching it!

PR #890, opened 2019-02-23
Remove the Adaptive transparency mode.

micheleg:
The panel dynamic transparency was dropped upstream:

https://gitlab.gnome.org/GNOME/gnome-shell/commit/9cfb51c106abd1f96012bf39f3d329cf060035cf

so this option does not make sense any more.

@3v1n0 @azzar1 : any objection? Is this something you are planning to keep supporting in Ubuntu? I'm even considering dropping the 'Dynamic' mode all together.

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -2160,7 +2160,6 @@
                                     <items>
                                       <item translatable="yes">Default</item>
                                       <item translatable="yes">Fixed</item>
-                                      <item translatable="yes">Adaptive</item>
                                       <item translatable="yes">Dynamic</item>
                                     </items>
                                   </object>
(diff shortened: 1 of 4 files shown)


azzar1:
@micheleg We're planning to drop both (adaptive and dynamic) from Ubuntu too.

micheleg:
>The only thing I'm not sure is how gsettings would handle the transition, but I guess that if the key value will be missing it the default value will be used again.

Yes, this is the expected behaviour. From https://developer.gnome.org/gio/stable/GSettings.html#g-settings-get-enum:
> If the value stored in the configuration database is not a valid value for the enumerated type then this function will return the default value.

PR #893, opened 2019-02-24
utils: use ES6 syntax to define _decimalToHex method

3v1n0:
Since we're in a class there should be no final comma, and the function can be
marked as static


micheleg:
Ops, yes. Thanks.

PR #936, opened 2019-04-26
Fix mouse trapped due to barrier set up with dock shown.

### LINKED ISSUE #743: Mouse gets stuck on secondary monitor after making a window full screen on primary monitor
The mouse cursor gets stuck on secondary monitor after making a window full screen on primary monitor (positioned on the left of the secondary monitor) when using Dash to Dock or Ubuntu Dock, with the following conditions:

1. Multi-monitor setups (obviously)
2. Ubuntu Dock / Dash to Dock with the auto-hide option enabled
3. Ubuntu Dock / Dash to Dock set to show on all displays
4. Ubuntu Dock / Dash to Dock position set to left (this is the default setting)
5. No window covering the Ubuntu Dock / Dash to Dock on the second monitor (so dock is visible on the second monitor)

The bug continues to occur after exiting full screen.

Changing any of this and the issue stops occurring.

I explained this with more details on my blog: [https://www.linuxuprising.com/2018/05/dual-monitor-fix-mouse-getting-stuck-on.html](https://www.linuxuprising.com/2018/05/dual-monitor-fix-mouse-getting-stuck-on.html)

I tested this in Ubuntu 18.04 with gnome-shell-extension-ubuntu-dock 0.9.1 / Dash to Dock v62 installed from the Gnome extensions website.

micheleg:
We were creating a barrier after each call to the updateBarrier
callback, regardless of the current dock state. If the dock is already shown,
this creat a barrier which is never removed, resulting in the mouse
being trapped. Check for the current dock state instead.

This is triggered for instance when an application enter/exit
fullscreen on any monitor.

The logic controlling pressure barriers and dock state needs
improvements but ths should fix it for now.

fix #743

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -974,16 +974,17 @@ var DockedDash = class DashToDock {
                 direction = Meta.BarrierDirection.NEGATIVE_Y;
             }
 
-            this._barrier = new Meta.Barrier({
-                display: global.display,
-                x1: x1,
-                x2: x2,
-                y1: y1,
-                y2: y2,
-                directions: direction
-            });
-            if (this._pressureBarrier)
+            if (this._pressureBarrier && this._dockState == State.HIDDEN) {
+                this._barrier = new Meta.Barrier({
+                    display: global.display,
+                    x1: x1,
+                    x2: x2,
+                    y1: y1,
+                    y2: y2,
+                    directions: direction
+                });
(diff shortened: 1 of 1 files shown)


Bloodsucker:
Why is this not merged and is part of a release? :/

PR #939, opened 2019-04-29
Show the "Quit" AppIconMenu entry only for running applications

jnsh:
The "Quit" AppIconMenu entry was displayed for every item on the dock, including favorited applications that aren't currently running.

Obviously you can't quit non-running applications, so it makes more sense to display the "Quit" menu item only if the application is actually running.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -880,11 +880,13 @@ const MyAppIconMenu = class DashToDock_MyAppIconMenu extends AppDisplay.AppIconM
         }
 
         // quit menu
-        this._appendSeparator();
-        this._quitfromDashMenuItem = this._appendMenuItem(_("Quit"));
-        this._quitfromDashMenuItem.connect('activate', () => {
-            this._source.closeAllWindows();
-        });
+        if (this._source.app.state == Shell.AppState.RUNNING) {
+            this._appendSeparator();
+            this._quitfromDashMenuItem = this._appendMenuItem(_("Quit"));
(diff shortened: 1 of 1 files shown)


azzar1:
I do not get the "Quit" menuitem if the application is not currently running. With which app are you able to reproduce this issue?

Also there is already some logic to hide/show the "Quit" menuitem: https://github.com/micheleg/dash-to-dock/blob/2c90645180d7a61d8e411263d972cd95261367e5/appIcons.js#L901

jnsh:
I can reproduce the issue with any application that's favorited to the dock and not running:
Screenshot from 2019-04-30 01-25-28

I didn't see the "Quit" menu item hiding/showing logic you mentioned though, and apparently it is only performed if the `show-windows-preview` option is enabled. If I enable "Show open windows previews" from dash-to-dock settings, the "Quit" menu entry is indeed (not) displayed correctly.

So, this PR would fix the "Quit" menu item only for when `show-window-preview` is disabled.

Additionally, I noticed that I'm never getting the "Quit \<number> Windows" menu entry with `show-window-preview` disabled, but it's always plain "Quit" regardless of the number of open windows. Therefore there could be more complete fix for the issue, that also implements the "Quit \<number> Windows" entry for people who don't enable window previews.

jnsh:
I added a new PR that should be a better fix: #940 

Thanks for the feedback on this one @azzar1

PR #940, opened 2019-04-30
Update AppIconMenu when window-previews are disabled

jnsh:
The AppIconMenu was only updated when show-window-preview option was enabled, which led to the "Quit" menu entry being displayed on non-running applications, and the number of windows to be quitted not being displayed when window previews were disabled.

Fix this by updating the "Quit" menu item and separators always, regardless of the show-window-preview setting.

Replaces #939


jnsh:
Updated to work and apply against the current master

PR #1011, opened 2019-10-09
MyAppIcon: Inherit from Dash.DashIcon

3v1n0:
Starting from GNOME/gnome-shell upstream commit ff3d32dd18 AppDisplay is
meant to be used only by the AppGrid and in fact it supports DnD drop
operations, being a drop target for other icons in order to be able to
create an app folder.

However dash-to-dock doesn't need anything of that, so in order to have
proper dock dnd inherit the app icons from the newly introduced (available
since the commit above, included in 3.34.0) Dash.DashIcon

LP: #1846477, #1847102

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -72,7 +72,7 @@ let recentlyClickedAppMonitor = -1;
  * - Update minimization animation target
  * - Update menu if open on windows change
  */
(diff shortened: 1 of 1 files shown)


micheleg:
Thanks.

PR #1012, opened 2019-10-09
Various cleanups and warning fixes

3v1n0:
Fixed warnings and errors accessing to dead objects as per upstream changes.

This meant to be added to #975, but I had no time to push in time :(


micheleg:
Grazie!

PR #1024, opened 2019-10-14
fixed a typo in a translatable string.

nishbone:
it should not be "Cutomize  opacity" but "Customize opacity".


micheleg:
Thank you.

PR #1026, opened 2019-10-17
Resolve animation hang issue in Gnome 3.34 with auto-hidden dock

acerspyro:
In Gnome 3.34, when activating an auto-hidden dock, quickly moving the mouse onto and off of the dock causes its animation to hang for the duration of the delay that was configured in the extension's settings.

This removes the delay if the dock is in the process of showing so that it returns to its hidden position as it was probably an accidental activation.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -697,14 +697,10 @@ var DockedDash = GObject.registerClass({
         // If no hiding animation is running or queued
         if ((this._dockState == State.SHOWN) || (this._dockState == State.SHOWING)) {
             let settings = DockManager.settings;
-            let delay;
+            // Remove delay if dock is in the process of showing
+            let delay = 0;
 
-            if (this._dockState == State.SHOWING)
-                //if a show already started, let it finish; queue hide without removing the show.
-                // to obtain this I increase the delay to avoid the overlap and interference
-                // between the animations
-                delay = settings.get_double('hide-delay') + settings.get_double('animation-time');
-            else
+            if (this._dockState != State.SHOWING)
                 delay = settings.get_double('hide-delay');
 
             this.emit('hiding');


micheleg:
Hi,

thanks for your contribution. I'm nost sure how this differs from the expected behaviour: "instead of chopping the animation just let it do its cycle naturally". Are you suggesting there is a bug or do you think this behaviour is not the ideal one?

acerspyro:
I believe the current behaviour is not ideal as the dock ends up just stopping abruptly mid-animation if you leave it too quickly. It feels like it's hanging.

micheleg:
Are you able to capture a video? There shouldn't be any abrupt stop of the animation. If so, then it might be a bug.

acerspyro:
I'll reply with a video shortly, this only began happening with 3.34.

Removing this code fixes it with no adverse effects.

acerspyro:
https://imgur.com/a/UQtDA6U

Bug shown here. Gnome 3.34, Manjaro

micheleg:
That looks surely bad :(

acerspyro:
> That looks surely bad :(

The video or the bug? lol

micheleg:
The bug :) I'm a bit puzzle because the comment to my code you are removing exactly suggest the delay is there to avoid this (and it makes logical sense). I suspect this might be due to the transition to clutter animations (662604b33f89f965e02664b63d1603be270e5c65).

micheleg:
I think that the issue is that while in tweener adding a transition with a delay used to do nothng until the transition start, with clutter animations the previous transitoin is immediately cancelled, and the animation is stalle for the above delay (https://github.com/GNOME/clutter/blob/master/clutter/clutter-actor.c#L273). Your solution avoid this problem but changes the behaviour from the desired one. Still better than the current situation but I'd like to see if it is possible to get back to the original behaviour.

3v1n0:
I think we should rely on the `onCompleted()` callback instead of doing timing assumptions.

So I've proposed https://github.com/micheleg/dash-to-dock/pull/1098 that addresses this using that.

@acerspyro, feel free to comment that PR, while I'm closing this.

Thanks for the debugging work though :)

Issue #1010, opened 2019-10-08
top right and bottom right corners only are always rounded


luca-viggiani:
Regardless the dock position chosen, top right and bottom right corners only are always rounded.
This is ok when the dock position is set to left. It is not for all other positions.
This is happening on GS 3.34 on Arch and Dash to Dock v67.
See attachment
![dash-to-dock](https://user-images.githubusercontent.com/25035914/66407529-a702a480-e9ed-11e9-8eeb-7a820908c664.png)

micheleg:
This seems to have been introduced in 5075934a5788e27cede7a67fbf568680775ec642.

PR #1027, opened 2019-10-17
Fix border radius for built-in theme

acerspyro:
The built-in theme does not account for Dash to Dock being able to place the dock anywhere on the screen, causing mismatched rounded corners if the dock is not placed on the left hand side while the built-in theme is enabled.

Fixes https://github.com/micheleg/dash-to-dock/issues/1010

### DIFF
--- a/stylesheet.css
+++ b/stylesheet.css
@@ -1,4 +1,4 @@
-/* Shrink the dash by reducing padding and border radius */
+/* Shrink the dash by reducing padding */
 #dashtodockContainer.shrink #dash,
 #dashtodockContainer.dashtodock #dash {
     border:1px;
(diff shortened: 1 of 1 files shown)


micheleg:
I have pushed a slightly different solutions:

1. Only apply styles with the build-in theme enabled (add .dashtodock class to the css rule)
2. Keep the border radius to 9px ( I believe this is what the default theme use).

As mentioned in the commit message, this bug was introduced in f2ce64a6277945b3a5494a090466d5c955162d65, in an attempt to maintain as much as possible theme compatibility. I believe this is a good compomize between the tow cases.

Does this still works for you?

acerspyro:
I'll take a look at it shortly, @micheleg

acerspyro:
Your solution works for me, this is a good workaround.

I believe a complete solution in the future would be to inherit the current theme's background, border values (colour, rounding, thickness, style, etc.) and shadow values and rotate them accordingly in JS.

PR #1030, opened 2019-10-17
docking.js: fix warning when using wheel

rastersoft:
When using the wheel over the dash to change from a workspace to another, a warning is shown in the log because it is still using .actor to access a widget. This patch fixes it.


micheleg:
Thanks.

PR #1042, opened 2019-11-03
Fixed typo in download.md

### LINKED ISSUE #1041: Typo in docs
There is an extra space after the backslash, that makes impossible to copy and paste the command a terminal. The issue is in https://micheleg.github.io/dash-to-dock/download.html under Installation under manual Installation.
Where can I make a PR for the website?

mone27:
fixes #1041


libreajans:
Your brach so old and I think it has no effect on the current version of dash-to-dock.

Even, there is no more download.md files.

libreajans:
Hi @mone27

There is also an extra/redundant space before "git clone" command.

` git clone https://github.com/micheleg/dash-to-dock.git`

Would you like to fix both of them?

PR #1044, opened 2019-11-05
locations: Fix missing behaviour for volume backed locations

### LINKED ISSUE #1037: "Show details" for volume mounts does nothing
If you right click on a volume mount, there's a "Show details" option like for other apps in the dock. Clicking it does nothing.

philipl:
I forgot to add this originally. Without it, we don't know that
the icon is a location rather than an app, and some undesired items
are added to the menu.

Mount icons already had this set and didn't have a problem.

Fixes #1037

### DIFF
--- a/locations.js
+++ b/locations.js
@@ -162,6 +162,10 @@ var Removables = class DashToDock_Removables {
             return;
         }
 
+        if (volume.get_identifier('class') == 'network') {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


philipl:
Added an additional commit that fixes certain network backed locations appearing in the dock when they should not.

@3v1n0 should be merged for ubuntu-dock too.

PR #1046, opened 2019-11-09
Update the screenshot to Gnome 3.34

xordspar0:
The current screenshot is getting a little old (for example, the default icon set has changed). This will update it to the latest version of Gnome.


xordspar0:
🎉

PR #1048, opened 2019-11-10
Fix unnecessary new line

libreajans:


pyroxar:
You have additional 0 in 3077 line. Fix It.

libreajans:
Fixed back. 

As a result of this patch, now we need to remove these kind unnecessary newlines from the PO files.

```
#: Settings.ui.h:70
msgid "Force straight corner\n"
msgstr "Zakázat zaoblené rohy\n"
``` 

Can I edit directly the PO files, or language teams must do it by themselves?

PR #1052, opened 2019-11-12
Improve Turkish translation

libreajans:
This patch improves Turkish translation using GNOME Shell's Turkish translation.


libreajans:
@TeknoMobil as you have known. It is done.

Issue #872, opened 2019-01-19
Disabling icon badge count not possible


LuFlo:
AFAIK it's not possible to disable the badge message count. Since it's not possible to disable this feature in Discord too, it is very annoying. Even if I disable all desktop notifications for Discord in Ubuntu, the numbers are still showing :(

fbruetting:
Hi! I came here to ask if you could please deactivate these badges by default. I have a very annoying issue with Geary:

E-Mail is not an instant messaging service, and therefore it’s counterproductive to often look or even reply to new emails (there are a lot of studies about that topic). This is pretty distracting because of its red error-like color and really creates pressure, because I often have to look there and it appears to me in a way like it wants to tell me all the time “Read it! Read it! Read it!”. This may not be felt by everyone, but I like things to be clean and as I have a good sight, I (sadly) recognize those badges all the time.

So to which behavior leads that issue for me? I open Geary, mark all new mails as read (without really realizing them, because they’re E-Mails and thus in almost all cases not important at all) and continue my work. Just for that badge to disappear. I don’t care if my mails are one or three days old, and I’m pretty sure that this counts for most people out. there. So at least in my case, the impact is purely negative.

So I’d therefore like to request to deactivate that badge by default, as this also is not contained in the GNOME HIG and leads to a LOT of distraction (at least for some people). I recommend just sticking to notifications and the notification center instead by default. But at the very minimum, an option for deactivation really is necessary here.

pvreevil:
+1

PR #1054, opened 2019-11-16
Make notification badges configurable

LuFlo:
Fixes: https://github.com/micheleg/dash-to-dock/issues/872
Disabling icon badge count not possible

In the past, the Unity-like notification badges where always enabled. It is now possible to enable or disable the badges in the settings UI in the Appearance Tab

### DIFF
--- a/schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
@@ -170,6 +170,11 @@
       <summary>TODO</summary>
       <description>TODO</description>
     </key>
+    <key type="b" name="show-notification-badges">
+      <default>false</default>
+      <summary>Show Unity-like notification and progress indicators.</summary>
+      <description>Show Unity-like notification and progress indicators.</description>
+    </key>
     <key type="b" name="custom-theme-customize-running-dots">
       <default>false</default>
       <summary>Customize the style of the running application indicators.</summary>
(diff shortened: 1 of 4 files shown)


schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
```diff
@@ -170,6 +170,11 @@
       <summary>TODO</summary>
       <description>TODO</description>
     </key>
+    <key type="b" name="show-notification-badges">
+      <default>false</default>
```
3v1n0: This should be `true` by default.

LuFlo: You are right, it should behave like the stock version of dash-to-dock. I somehow thought it was
on true... Thanks for pointing this out

LuFlo:
@3v1n0 is this ok how it is now? :)

schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
```diff
@@ -170,6 +170,11 @@
       <summary>TODO</summary>
       <description>TODO</description>
     </key>
+    <key type="b" name="show-notification-badges">
```
jtojnar, line 173: It might be nice to have this tri-state:

- `yes`
- `no`
- `when-not-in-dnd-mode` https://github.com/micheleg/dash-to-dock/issues/1329

3v1n0:
Sorry, I noticed I lost track of this, would you like to rebase it?

Ideally the tri-state thingy can be also done using an enum.

3v1n0:
This is handled now

PR #1058, opened 2019-11-21
update Simplified Chinese translation

zhmars:


laichiaheng:
@micheleg Would you approved these translations, thanks.

PR #1103, opened 2020-02-29
locations: Offer to filter out devices by regex

### LINKED ISSUE #1081: No selection possible which drive icons are shown on the dock
Searching for a "blacklist" of removable drives shown in the dock did not lead to anywhere.
The only thing that can be done is to disable the display of all or of none icons shown for removable/mounted devices.

Is there an option planned to be able to select e.g. only USB/DVD whatsoever volumes to be displayed in the dock instead of showing also mounted devices which I mount at startup (e.g. a bigger disk for data, which I do not need to be shown in the dock, as it works for me in the background and the icons just fill-up my dock)

br,
Kurt

philipl:
People have asked for the ability to filter out certain devices from
appearing in the doc and today it is all-or-nothing.

This change is somewhat a proof-of-concept as the user configuration
is literally a regex, which is very powerful but not terribly
intuitive. I would be interested in suggesitons on how to handle the
configuration.

Fixes #1081

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -1418,6 +1418,35 @@
                             <property name="top_attach">0</property>
                           </packing>
                         </child>
+                        <child>
+                          <object class="GtkEntry" id="hide_mounts_regex_entry">
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                            <property name="hexpand">True</property>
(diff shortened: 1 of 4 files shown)


philipl:
@3v1n0 any thoughts?

PR #1111, opened 2020-03-08
Fix bug with xscreensaver.

### LINKED ISSUE #1002: Dash no longer hidden by xscreensaver screen locking
### Current behavior
Since the Gnome 3.34 update, dash-to-dock is showing up on the screen when it's locked by xscreensaver.  The dock appears overlayed on top of the blanked screen/running screensaver.

### Expected behavior
Prior to this update, the dock was hidden along with the rest of the screen whenever xscreensaver locks the screen.

### Version info
OS: Arch

Dash-to-dock: ~~Installed via the aur package, which is basing off of this commit for gnome 3.34 compatibilty: https://github.com/micheleg/dash-to-dock/commit/908a05022a1ce2565d56af4c762fe95c5fb6b39c~~
v67 release installed via extensions.gnome.org

xscreensaver: Version 5.43

### Other
Nothing else appears to have broken in this way after the Gnome 3.34 update.
Some info on how xscreensaver works:
https://www.jwz.org/xscreensaver/man1.html
> When it is time to activate the screensaver, a full-screen black window is created on each screen of the display. Each window is created in such a way that, to any subsequently-created programs, it will appear to be a "virtual root" window. Because of this, any program which draws on the root window (and which understands virtual roots) can be used as a screensaver.

micheleg:
The dock is visible when xscreensaver is triggered. This occur in Gnome
Shell 3.34, but not in 3.30. Slighly changing the position of the dock
in the stack seems to fix the problem. This might be a consequence of
upstream commit 2bd80579eddba9804871ce6c8de742bdf9b1d800.  This also
requires to disable the machinery in place to keep the dock on top of
everying on screen grab seems. Luckily, this seems to not to be
necessary anymore.

Fix #1002

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -372,6 +372,8 @@ var DockedDash = GObject.registerClass({
 
         // Add aligning container without tracking it for input region
         Main.uiGroup.add_child(this);
+        if (Main.uiGroup.contains(global.top_window_group))
+            Main.uiGroup.set_child_below_sibling(this, global.top_window_group);
 
         if (settings.get_boolean('dock-fixed')) {
             // Note: tracking the fullscreen directly on the slider actor causes some hiccups when fullscreening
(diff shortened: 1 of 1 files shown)


3v1n0:
LGTM!

PR #1172, opened 2020-04-19
Fix regression (wrong sign when computing position).

micheleg:
Introduced in 4a40b1de64ba83656f0df40cc8c91eb34cd4d302. The problem was
occurring only when monitor.y !=0 (and with extended mode with execute
the code).


lazyfrosch:
This resolves my problem! Thanks :green_heart:

PR #1186, opened 2020-04-24
Using GVolumeMonitor references correctly.

### LINKED ISSUE #1025: Disabling "Show mounted volumes and devices" hides all icons on second monitor
Disabling _Show mounted volumes and devices_ hides all icons on second monitor. Dash to Dock version 67 on Ubuntu 19.10.

darkdragon-001:
Fixes #1025.

### DIFF
--- a/locations.js
+++ b/locations.js
@@ -101,7 +101,7 @@ var Removables = class DashToDock_Removables {
     constructor() {
         this._signalsHandler = new Utils.GlobalSignalsHandler();
 
-        this._monitor = Gio.VolumeMonitor.get();
+        this._monitor = Gio.VolumeMonitor.get().ref();
         this._volumeApps = []
         this._mountApps = []
 
@@ -138,7 +138,7 @@ var Removables = class DashToDock_Removables {
 
     destroy() {
         this._signalsHandler.destroy();
-        this._monitor.run_dispose();
+        this._monitor.unref();
     }
 
     _getWorkingIconName(icon) {


darkdragon-001:
Can somebody please review?

@3v1n0 @micheleg

3v1n0:
Mh, doing ref-unref/ from JS is generally bad but, why do you have to ref it all the times you connect to a signal?

I'd say you could just do it once at class initialization and destruction, although that would probably leak as well

darkdragon-001:
I don't know too much of the Gio internals. Doing it once on creation and once on destruction is enough.

Or is it done automatically in the JS destructor meaning that you don't have to ref/unref/dispose it manually at all?

3v1n0:
> Or is it done automatically in the JS destructor meaning that you don't have to ref/unref/dispose it manually at all?

Yes, the ref management is all done inside JS.
If you implement a "destroy" function (it means you release all the refs of it, and disconnect signals), then is fine to call `run_dispose()`, so that you actually release its resources.

darkdragon-001:
How should we proceed fixing #1025? Currently, the code tries to destroy a shared object multiple times, so we need some reference counting to avoid this situation. What do you think of the updated version? Does it free all resources correctly?

3v1n0:
Mh, I've to look at this better, but I assume we can make sure that we don't destroy the monitor instance if we've docks around.
While destroy this only together with the DockManager.

3v1n0:
This is now handled via #1576, so closing.

Issue #1224, opened 2020-05-14
Most of the translatable strings from appIcons.js not available for translation


AsciiWolf:
Most of the translatable strings that are present only in [appIcons.js](https://github.com/micheleg/dash-to-dock/blob/master/appIcons.js) does not seem to be detected by gettext and is not available in the generated potfile/pofiles. The only strings that seem to be added are `"All Windows"` and `"Dash to Dock %s"`, other strings are missing.

PR #1225, opened 2020-05-14
Update Czech translation

AsciiWolf:
Unfortunately, some strings are missing because of #1224. They are missing in most translations.

/cc @3v1n0 @dvorapa

**edit:** I was able to fix the missing strings, please merge this PR after #1226 is merged.

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -7,15 +7,15 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Dash to Dock\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-04-01 03:01+0200\n"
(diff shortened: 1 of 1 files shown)


po/cs.po
```diff
@@ -58,17 +58,70 @@ msgstr "Nastavit indikátory spuštěných aplikací"
...
+msgid "Add to Favorites"
+msgstr "Přidat mezi oblíbené"
+
+#: appIcons.js:868
+msgid "Show Details"
+msgstr "Zobrazit podrobnosti"
+
+#: appIcons.js:896 appIcons.js:914 Settings.ui.h:11
+msgid "Quit"
+msgstr "Ukončit"
```
AsciiWolf, line 87: I have changed this to "Ukončit" (which is a correct translation that is already used in GNOME Shell) instead of "Zastavit".

PR #1226, opened 2020-05-15
appIcons: Fix translatable strings

### LINKED ISSUE #1224: Most of the translatable strings from appIcons.js not available for translation

AsciiWolf:
Fixes #1224.

All translatable strings are now correctly exported by gettext, however I did not test running the actualy extension after this change and I don't fully understand the gettext domain/shell domain thing. But hopefully, it should work fine.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -805,7 +805,7 @@ const MyAppIconMenu = class DashToDock_MyAppIconMenu extends AppDisplay.AppIconM
                 let actions = appInfo.list_actions();
                 if (this._source.app.can_open_new_window() &&
                     actions.indexOf('new-window') == -1) {
-                    this._newWindowMenuItem = this._appendMenuItem(_("New Window"));
+                    this._newWindowMenuItem = this._appendMenuItem(_('New Window'));
                     this._newWindowMenuItem.connect('activate', () => {
                         if (this._source.app.state == Shell.AppState.STOPPED)
                             this._source.animateLaunch();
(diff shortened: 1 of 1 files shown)


AsciiWolf:
Some of these strings are also available (and translated) in GNOME Shell so I am not sure if they should be translatable in Dash to Dock, however at least the `Quit %d Windows` string is definitely not available anywhere in GNOME Shell.

dvorapa:
I found a similar `_('Dash')` string in docking.js

AsciiWolf:
@dvorapa Hmm, the docking.js file [does not seem](https://github.com/micheleg/dash-to-dock/blob/master/Makefile#L7) to be translatable at all.

3v1n0:
Mmh, I think we should use the upstream strings translations as much as possible, so just use `__()` when a string isn't translated there.

I think most of these are actually translated, no?

AsciiWolf:
@3v1n0 Yeah, seems so. The only string that is not in upstream GNOME Shell (3.36 and current master) is the `Quit %d Windows` one. I will fix this PR. I will also make a new Czech translation update PR afterwards.

AsciiWolf:
I have pushed updated commit that only adds `__(` for `Quit %d Windows` (and fixes quotation marks for all strings).

AsciiWolf:
@3v1n0 Please, review the changes and let me know if there is anything wrong. Thanks!

PR #1245, opened 2020-05-28
Support more Unity features (quicklist, urgent properties)

rhendric:
This adds support for the two properties of a libunity LauncherEntry object not yet supported: "quicklist" and "urgent". Quicklist entries are added to the right-click menu of the launcher as dynamic menu items. The urgent flag, if set, gives the icon a short dancing animation to call attention to it. These features don't seem to be widely used by applications, but there's a bit of a chicken-and-egg problem here; why would application authors spend time writing code to use them if they aren't supported by the desktop?

I encountered some trouble with the libunity communication, and I also had some trouble understanding the intent of the original libunity integrating code, so I took the liberty of refactoring it into a pattern that, among other advantages, is more robust to multiple DBus applications contending for the same LauncherEntry. Hopefully you see it as an improvement too; I can try to back it out if not, but it did make adding these two features much easier.

I've been using this branch for the better part of a year now, and tested the new features with my own custom applications; it all has been working quite well for me. I hope it does some good for someone else.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -32,6 +32,7 @@ const Docking = Me.imports.docking;
 const Utils = Me.imports.utils;
 const WindowPreview = Me.imports.windowPreview;
 const AppIconIndicators = Me.imports.appIconIndicators;
+const DbusmenuUtils = Me.imports.dbusmenuUtils;
 
 let tracker = Shell.WindowTracker.get_default();
 
@@ -293,7 +294,7 @@ class MyAppIcon extends Dash.DashIcon {
         this._draggable.fakeRelease();
 
         if (!this._menu) {
-            this._menu = new MyAppIconMenu(this);
+            this._menu = new MyAppIconMenu(this, this.remoteModel);
             this._menu.connect('activate-window', (menu, window) => {
                 this.activateWindow(window);
             });
@@ -768,7 +769,7 @@ class MyAppIcon extends Dash.DashIcon {
  */
 const MyAppIconMenu = class DashToDock_MyAppIconMenu extends AppDisplay.AppIconMenu {
 
-    constructor(source) {
+    constructor(source, remoteModel) {
         let side = Utils.getPosition();
 
         // Damm it, there has to be a proper way of doing this...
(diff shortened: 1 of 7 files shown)


3v1n0:
Oh, nice work... I will review this a bit later in the cycle.

But looks like a good thing to do and in my list as well.

appIcons.js
```diff
@@ -891,6 +920,22 @@ const MyAppIconMenu = class DashToDock_MyAppIconMenu extends AppDisplay.AppIconM
                 super._redisplay();
         }
 
+        // dynamic menu
+        const items = this._getMenuItems();
+        let i = items.length;
+        if (Shell.AppSystem.get_default().lookup_app('org.gnome.Software.desktop')) {
```
3v1n0, line 926: What was the reason for this special behavior?

3v1n0 (approved):
Overall this looks great, I've been through the code and this seems to do things as they were expected to be.

So approving.

I've some ideal changes to do though:
 - Make importing DBusmenu dynamic (may not be available in other distros than ubuntu)
 - Change the animation mechanism (doesn't work in 3.38, but mostly could be simplified).

However, I've implemented both as ideally I was thinking about in https://github.com/3v1n0/dash-to-dock/commits/unity-features-improvements

Feel free to propose another MR if you have better ideas about.
Merging this for now :)

appIcons.js
rhendric: This is to position the quicklist menu items just below the launcher action items. There are two conditional items in the menu that appear below the launcher actions (‘Add to/Remove from Favorites’, ‘Show Details’); depending on the value of the `show-windows-preview` setting, they are added by either [upstream code](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/8745a61e4002a3c6ac737e2b51118b900dbc5d26/js/ui/appDisplay.js#L2830-2868) or a few lines above in this file, but only if certain conditions are met. To insert the quicklist section in the correct place in either branch, I need to do the same tests that upstream/the code above does, and when true, decrement the position of the quicklist section by 2 (item + separator).

3v1n0: I see thanks!

PR #1256, opened 2020-06-15
Fixes typo on org.gnome.shell.extensions.dash-to-dock

pedropaulosuzuki:
Fixes 'appplication' typo on dash-to-dock description for some options.


micheleg:
Thanks!

PR #1283, opened 2020-07-27
Reset LRU on leave event of appIcon

gonzaarcr:
Like Unity, reset the list of windows when leaveing (unfocusing) the app icon; instead of in a random time.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -53,7 +53,6 @@ const scrollAction = {
     SWITCH_WORKSPACE: 2
 };
 
-let recentlyClickedAppLoopId = 0;
 let recentlyClickedApp = null;
 let recentlyClickedAppWindows = null;
 let recentlyClickedAppIndex = 0;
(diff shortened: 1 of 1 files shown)


micheleg:
The code and implementation loog good to me.

I'm not sure this is the best behaviour: when I implemented it this way I thought it was usefull for the icon to remember the position for a while. I agree this was randomly choosen, and your implementation is more determininstic.

Can you suggest a situation where this new behaviour is more natural the the original one?

gonzaarcr:
Well, it probably only affects me because I use this feature a lot, I guess nobody uses it. I think the most annoying scenario is when you are slowly changing windows to find one (specially with applications with tabs, which you don’t remember in what windows is what are you looking for) and then you scroll and it take you a while until you realize the scroll was rebooted, and you think you loop throgh all the windows but you didn’t.

3v1n0:
I guess this is the kind of case where having a preference to define this would be the best approach.

Issue #1262, opened 2020-06-30
Dock hiding magically after touching the bottom of the screen in Fedora 32 - Gnome 3.36


edier88:
Hi, I have written in the comments section of the Gnome Extensions page multiple times.
The extension works great in previous releases of Fedora with Gnome 3.34 and older, but with the new Fedora 32 with Gnome 3.36 it behaves terrible.
In random times, when I touch the bottom of the screen to make the dock appear, it magically hides again. 
That happens multiple times randomly and it's SO annoying cause if you click an icon of the Dock in the right moment it is hiding you can make a mistake, clicking an area of your screen making your work go to hell!
Please, is there any solution to this??
Thank you very much!!

binnih:
I have the same problem, this happens on all three of my Fedora 32 machines. It is really annoying. It almost always takes 2 to 3 attempts to get the dock to stay up so I can use it. I have to move the mouse up and down again until the dock comes up properly.
This has been like this for a while, started after upgrading from 31 to 32.

edier88:
Has anyone come across this problem in Fedora 32? 
I think this is a general problem for Fedora 32 and I don't know if other distros have the same problem. It's quite weird that they haven't this problem.

faraz-naeinian:
It's not just Fedora, I use Pop! OS on two PCs and I have this exact same problem, so annoying.

Split7fire:
I have similar problem on Fedora 32. I think I found a workaround: Increase "hiding timeout" to 0.5 sec or more.

edier88:
Hi all.
After upgrading to Fedora 33 is working great.
I had terrible issues in Fedora 32, now everything goes smoothly.
Thank you very much!

Issue #1251, opened 2020-06-08
Intellihide bounce with Manjaro 20.0.3


DaviDev-k:
Quite often when I move the mouse to the bottom of the screen to reveal the hidden dock, it starts to bounce for 1/4 of the height and it doesn't stop until I move the pointer out of the dock, it's very annoying.
Intellihide settings are set to default, the only difference is that "Dodge the windows" (idk if it's the right translation) is set to "All windows".

Guillaume-Savaton-ESEO:
I have the same issue in Ubuntu 20.04 with Gnome 3.36.2

Issue #1207, opened 2020-05-06
Stutters when tooltip is appears on th dock


Latesil:
Fedora 32, 64-it
GNOME 3.36.2
Dash-to-Dock 68

There is a short video that explains a lot: https://www.youtube.com/watch?v=bmvVXoLe7s8 
Looks like that happens when tooltip appears on the dock, dissapears, dock is moving down but then tooltip appears again, dock gets back to full size, tooltip dissapears again and so on.

cb3inco:
I experience this too on Pop_OS! 20.04 64-bit. Gnome 3.36.1, Dash to Dock v68.

JensVanhooydonck:
I have the exact sa,e thing on Pop_OS!20.04 LTS 64-bit with GNOME Version 3.36.1 and Dash to Dock version 68.

cb3inco:
> I have the exact same thing on Pop_OS!20.04 LTS 64-bit with GNOME Version 3.36.1 and Dash to Dock version 68.
> 
> When i set the Gap to 0, this doesn't happen anymore.

Thanks for the tip, that worked for me too, I also bumped up the gap to 3 and that seems to be working good too. Perhaps its a Pop_OS bug and not Dash to Dock.

Latesil:
> Perhaps its a Pop_OS bug and not Dash to Dock.

I am on Fedora and still experience that bug.

JensVanhooydonck:
> I am on Fedora and still experience that bug.

Did you install the Pop shell extension?

Latesil:
> > I am on Fedora and still experience that bug.
> 
> Did you install the Pop shell extension?

No, I installed this: https://micheleg.github.io/dash-to-dock/

JensVanhooydonck:
It seems it is caused by the resize window action. When the Gap in Pop Shell is set to 0, there is no resize window handle when touching the bottom of the screen. When the gap is set to 1 or 2, the resize symbol is shown when touching the bottom of the screen. When the gap is set to >= 3, the cursor isn't touching the window again and so not showing the resize icon.

A-UNDERSCORE-D:
Confirming this for Ubuntu 20.04 and GNOME version 3. 36.1

ceiphr:
Same issue Ubuntu 20.04 LTS. GNOME version 3.36.2.

Uchiha01Shivam:
[...]

PR #1296, opened 2020-08-16
Fix #1232

### LINKED ISSUE #1232: dash-to-dock glitches while in auto-hide setting
When "Auto-hide the Dock" is enabled and there is a maximized window, so that the dock is hidden, upon moving the mouse to the bottom of the screen, sometimes the dock glitches and initiates a hide-show animation without stopping instead from showing up until the cursor moves away from the dock.

https://youtu.be/Lw2HBmLhd6w

This bug was first noticed after upgrading to ubuntu 20.04 from 19.10

Description: Ubuntu 20.04 LTS
Release: 20.04

gnome-shell-extension-ubuntu-dock:
  Installed: 67ubuntu20.04.5
  Candidate: 67ubuntu20.04.5
  Version table:
  67ubuntu20.04.5 500
        500 http://gr.archive.ubuntu.com/ubuntu focal/main amd64 Packages
        500 http://gr.archive.ubuntu.com/ubuntu focal/main i386 Packages
        100 /var/lib/dpkg/status

ProblemType: Bug
ApportVersion: 2.20.11-0ubuntu27
Architecture: amd64
CasperMD5CheckResult: skip
CurrentDesktop: ubuntu:GNOME
DistroRelease: Ubuntu 20.04
InstallationDate: Installed on 2019-10-27 (204 days ago)
InstallationMedia: Ubuntu 19.10 "Eoan Ermine" - Release amd64 (20191017)
NonfreeKernelModules: nvidia_modeset nvidia
Package: gnome-shell-extension-ubuntu-dock 67ubuntu20.04.5
PackageArchitecture: all
ProcVersionSignature: Ubuntu 5.4.0-29.33-generic 5.4.30
Tags: focal
Uname: Linux 5.4.0-29-generic x86_64
UpgradeStatus: Upgraded to focal on 2020-05-09 (9 days ago)
UserGroups: adm cdrom dip lpadmin lxd plugdev sambashare sudo
_MarkForUpload: True

UA27:
Been using this fork for a week now and it fixes #1232 for me. It improves a few odd behaviours around the intellihide/autohide docking setup and mainly deals with the bouncing glitch using `_onPressureSensed`'s pointer-in-dock-area-check method.

On that, I want to state that the main issue came from `_box.hover` being unreliable if any of the maintainers wants to dig into the whole hover ordeal. I mostly worked around it the way I did because I don't know enough about how Gnome handles these things and it made sense to use the same check whatever the dock's state is in to avoid further weird, contradicting and unpredictable behaviours popping up.

_P.S. I was able to reproduce the issue using older commits of dash-to-dock's master on an Ubuntu 20.04 systen, so probably has something to do with newer versions of gnome-shell or I just suck at debugging Gnome, which I do, so feel free to discard anything I said and try it out yourself._

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -618,12 +618,11 @@ var DockedDash = GObject.registerClass({
                 this._ignoreHover = false;
                 // Do not hide if autohide is enabled and mouse is hover
                 if (!this._box.hover || !this._autohideIsEnabled)
-                    this._animateOut(settings.get_double('animation-time'), 0);
+                    this._hide();
             }
             else {
                 this._ignoreHover = true;
-                this._removeAnimations();
-                this._animateIn(settings.get_double('animation-time'), 0);
+                this._show();
             }
         }
         else {
@@ -655,16 +654,16 @@ var DockedDash = GObject.registerClass({
     }
 
     _hoverChanged() {
-        if (!this._ignoreHover) {
-            // Skip if dock is not in autohide mode for instance because it is shown
-            // by intellihide.
-            if (this._autohideIsEnabled) {
-                if (this._box.hover)
-                    this._show();
-                else
-                    this._hide();
-            }
+        // Skip if dock is not in autohide mode for instance because it is shown by intellihide or
+        // if the dock's `_ignoreHover` value is set to true.
+        if (this._ignoreHover || !this._autohideIsEnabled) {
+            return;
         }
+
+        if (this._triggerTimeoutId)
+            this._isPointerInZone() || this._box.hover ? this._show() : this._hide();
+        else
+            this._box.hover ? this._show() : this._hide();
     }
 
     getDockState() {
@@ -672,6 +671,9 @@ var DockedDash = GObject.registerClass({
     }
 
     _show() {
+        // Remove any delayed hide animation.
+        delete this._delayedHide;
+
         if ((this._dockState == State.HIDDEN) || (this._dockState == State.HIDING)) {
             if (this._dockState == State.HIDING)
                 // suppress all potential queued transitions - i.e. added but not started,
@@ -703,7 +705,6 @@ var DockedDash = GObject.registerClass({
 
     _animateIn(time, delay) {
         this._dockState = State.SHOWING;
-        delete this._delayedHide;
 
         this._slider.ease_property('slidex', 1, {
             duration: time * 1000,
@@ -721,7 +722,9 @@ var DockedDash = GObject.registerClass({
                     this._removeBarrierTimeoutId = GLib.timeout_add(
                         GLib.PRIORITY_DEFAULT, 100, this._removeBarrier.bind(this));
                 } else {
-                    this._hide();
+                    // If an animate-out transition was delayed, check if
+                    // it is still necessary.
+                    this._hoverChanged();
                 }
             }
         });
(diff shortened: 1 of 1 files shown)


docking.js
```diff
@@ -618,12 +618,11 @@ var DockedDash = GObject.registerClass({
...
                 // Do not hide if autohide is enabled and mouse is hover
                 if (!this._box.hover || !this._autohideIsEnabled)
-                    this._animateOut(settings.get_double('animation-time'), 0);
+                    this._hide();
             }
             else {
                 this._ignoreHover = true;
-                this._removeAnimations();
-                this._animateIn(settings.get_double('animation-time'), 0);
+                this._show();
```
UA27, line 625: Use `_hide` and `_show` instead to improve readability and keep control over transition behaviours. Only thing is, it now uses the extension's animation delay settings. I don't think it messes with any other setup, but someone with more experience of dash-to-dock's options should probably confirm it.

docking.js
```diff
@@ -655,23 +654,26 @@ var DockedDash = GObject.registerClass({
...
+        // Skip if dock is not in autohide mode for instance because it is shown by intellihide or
+        // if the dock's `_ignoreHover` value is set to true.
+        if (this._ignoreHover || !this._autohideIsEnabled) {
+            return;
         }
+
+        if (this._triggerTimeoutId)
+            this._isPointerInZone() || this._box.hover ? this._show() : this._hide();
+        else
+            this._box.hover ? this._show() : this._hide();
```
UA27, line 666: Keeping the `_box.hover` check in an else statement guarantee other docks that do not use `_onPressureSensed` will keep the same behaviour. Because of this, I think setups using intellihide/autohide without the push-to-show option (most likely) will still have the bouncing glitch issue.

Again, I'll leave this part to be reviewed by maintainers with better understanding of the many dash-to-dock's options.

docking.js
```diff
@@ -655,23 +654,26 @@ var DockedDash = GObject.registerClass({
...
     }
 
     getDockState() {
         return this._dockState;
     }
 
     _show() {
+        // Remove any delayed hide animation.
+        delete this._delayedHide;
+
```
UA27, line 676: Delete the `_delayedHide` value in `_show` instead of `_animateIn` since calls to said `_show` are ignored when the dock's state is either "shown" or "showing" and, when a delayed hide transition is queued, the dock's state is only updated AFTER the animate-in transition is completed.

docking.js
```diff
@@ -860,60 +863,67 @@ var DockedDash = GObject.registerClass({
...
                     y <= this._monitor.y + this._monitor.height) {
-                    shouldHide = false;
+                    return true;
                 }
+                break;
             }
-            if (shouldHide) {
+
+        return false;
+    }
```
UA27, line 908: Moved this big boy out so it can be used both in `_onPressureSensed` and `_hoverChanged`.

docking.js
```diff
@@ -721,7 +722,9 @@ var DockedDash = GObject.registerClass({
                     this._removeBarrierTimeoutId = GLib.timeout_add(
                         GLib.PRIORITY_DEFAULT, 100, this._removeBarrier.bind(this));
                 } else {
-                    this._hide();
+                    // If an animate-out transition was delayed, check if
+                    // it is still necessary.
+                    this._hoverChanged();
```
UA27, line 727: Replace the `_hide` call to `_hoverChanged` the same way `shouldHide` was (is) used in `_onPressureSensed` to catch any `_delayedHide` shenanigans by checking whether the delayed animate-out transition is still necessary instead of forcing the dock to hide right away and trigger a bounce from `_box`'s hover-changed signals.

UA27:
Lastly, this fixes my own issues with #1232, but it probably also concerns #1262, #1251, #1250 and #1207.

vltr:
oh my this issue is **so annoying** and I was almost loosing hope on this matter ... until this proposed fix. I'm using Gnome 3.36 on Arch Linux and the extension seems to be behaving itself "as normal" again! I almost can't believe it ... thank you, @UA27 !

GabMus:
What's blocking this from being merged?

k0lourful:
Thank you for this, helps me as well! I honestly didn't realise this issue is so common until you directed me here.

3v1n0 (approved):
Changes looks sane, thanks for looking at this.

3v1n0:
And thanks for commenting the changes as well!

PR #1308, opened 2020-09-10
Unity features improvements

3v1n0:
Cleanup a bit the changes introduced in #1245, as per the comment to @rhendric, and particularly:

 - Make importing DBusmenu dynamic (may not be available in other distros than ubuntu)
  - Change the animation mechanism (doesn't work in 3.38, but mostly could be simplified).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 UUID = dash-to-dock@micxgx.gmail.com
 BASE_MODULES = extension.js stylesheet.css metadata.json COPYING README.md
-EXTRA_MODULES = dash.js docking.js appIcons.js appIconIndicators.js fileManager1API.js launcherAPI.js locations.js windowPreview.js intellihide.js prefs.js theming.js utils.js Settings.ui
+EXTRA_MODULES = dash.js docking.js appIcons.js appIconIndicators.js fileManager1API.js launcherAPI.js locations.js windowPreview.js intellihide.js prefs.js theming.js utils.js dbusmenuUtils.js Settings.ui
 EXTRA_MEDIA = logo.svg glossy.svg highlight_stacked_bg.svg highlight_stacked_bg_h.svg
 TOLOCALIZE =  prefs.js appIcons.js locations.js
 MSGSRC = $(wildcard po/*.po)
--- a/appIcons.js
+++ b/appIcons.js
@@ -785,7 +785,7 @@ const MyAppIconMenu = class DashToDock_MyAppIconMenu extends AppDisplay.AppIconM
 
         this._signalsHandler = new Utils.GlobalSignalsHandler();
 
-        if (remoteModel) {
+        if (remoteModel && DbusmenuUtils.haveDBusMenu()) {
             const [onQuicklist, onDynamicSection] = Utils.splitHandler((sender, { quicklist }, dynamicSection) => {
                 dynamicSection.removeAll();
                 if (quicklist) {
--- a/dash.js
+++ b/dash.js
@@ -172,7 +172,7 @@ var MyDash = GObject.registerClass({
     }
 }, class DashToDock_MyDash extends St.Bin {
 
-    _init(remoteModel, monitorIndex, iconAnimator) {
+    _init(remoteModel, monitorIndex) {
         // Initialize icon variables and size
         this._maxHeight = -1;
         this.iconSize = Docking.DockManager.settings.get_int('dash-max-icon-size');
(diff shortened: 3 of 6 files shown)


rhendric:
Begging your pardon, but I don't think changing the urgent animation to use the upstream ‘wiggle’ is an improvement.

First, I think the wiggle animation is used exclusively by upstream for password fields where the password entered was incorrect, yes? I don't think the urgent animation needs to be made consistent with that because it's communicating a different idea—‘something needs your attention’ versus ‘that was wrong; try again’. (In much of the world, the wiggle evokes someone shaking their head ‘no’, which reinforces the specificity of the ‘that was wrong; try again’ interpretation.)

Second, I just tested it and it doesn't seem to work correctly for me on 3.36.6; it seems to make the urgent icon disappear.

Third, because I couldn't test it I couldn't verify the following, but just from looking at the code, it looks like the wiggle animation is a one-shot instead of repeated. If true, that doesn't seem like it would fulfill the purpose of having an urgency indicator—if the dock is hidden, or if the user's attention (or body) is elsewhere, they'll miss the animation and it will have added no value.

Fourth, the original code was written so that if multiple icons were urgent, their animations would be synchronized. This is pure aesthetics but I think that looks much more pleasing than out of sync animations. I don't think the wiggle transition does this (there's never a need for multiple password fields to tell you that you need to retry simultaneously!), but again, I couldn't test so I can't be sure.

Obviously, if the original patch doesn't work on 3.38 that needs to be fixed, so I am all for helping figure out how to do that. I am also open to either streamlining `IconAnimator` or adding more comments that explain the need for more complex bits. But I don't think it should be by using `Util.wiggle` like this.

(There is one piece of future-anticipating complexity in there that you *didn't* remove: I had an eye towards implementing multiple possible urgent animations that could be chosen by preference, which is why animations are named and `_animations` is an object of arrays instead of just an array. I would sleep just fine if you wanted this to be simplified out until there is another animation to choose.)

3v1n0:
> Begging your pardon, but I don't think changing the urgent animation to use the upstream ‘wiggle’ is an improvement.
> 
> First, I think the wiggle animation is used exclusively by upstream for password fields where the password entered was incorrect, yes? I don't think the urgent animation needs to be made consistent with that because it's communicating a different idea—‘something needs your attention’ versus ‘that was wrong; try again’. (In much of the world, the wiggle evokes someone shaking their head ‘no’, which reinforces the specificity of the ‘that was wrong; try again’ interpretation.)

Ok, fair enough... :+1:

I've reverted the timeline you were using, but adding fixes for 3.38. Now, however I'd prefer to use the easing animations for doing the same, but more to come.

However, in the animation side, I think the unity one was using some other logic (I need to check still) and I feel like it looked a bit nicer :).

> Second, I just tested it and it doesn't seem to work correctly for me on 3.36.6; it seems to make the urgent icon disappear.

Yeah, I assume that's the same reason why it didn't work for 3.37, but this should be fixed now

> Third, because I couldn't test it I couldn't verify the following, but just from looking at the code, it looks like the wiggle animation is a one-shot instead of repeated. If true, that doesn't seem like it would fulfill the purpose of having an urgency indicator—if the dock is hidden, or if the user's attention (or body) is elsewhere, they'll miss the animation and it will have added no value.

No, in the previous code the icon was shaking anyway. And happening when the launcher was shown.

What it wasn't doing, that your change is doing, is to dance every few minutes... I'm not sure that's the best approach, I'd in fact prefer to use what unity did:
 - Shake once it happens (presenting the launcher or the icon if it's hidden)
 - Until the user doesn't focus the application, we "glow" the icon so that it's still shows as needs attention.

> Fourth, the original code was written so that if multiple icons were urgent, their animations would be synchronized. This is pure aesthetics but I think that looks much more pleasing than out of sync animations. I don't think the wiggle transition does this (there's never a need for multiple password fields to tell you that you need to retry simultaneously!), but again, I couldn't test so I can't be sure.

Sure, they happened all together. Internally the clutter easing uses transitions anyways, so there's no difference.
The only difference was that if a new urgent icon was added only the new one was wiggling, not them all.

> (There is one piece of future-anticipating complexity in there that you _didn't_ remove: I had an eye towards implementing multiple possible urgent animations that could be chosen by preference, which is why animations are named and `_animations` is an object of arrays instead of just an array. I would sleep just fine if you wanted this to be simplified out until there is another animation to choose.)

Sure, i got this was the reason... Although probably there's no much need for keeping them as a map, as I'd imagine the animation would still be unique per all the icons, so we could just store the active icons and then apply the configured animation.

micheleg (approved):

PR #1311, opened 2020-09-14
Update to latest wording of the GPLv2 license

mikedep333:
Downloaded from:
https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt

This fixes the address, which was causing RPM lint error:
incorrect-fsf-address


3v1n0 (approved):
Thanks!

PR #1314, opened 2020-09-20
Fix getWindows for a location in the dock

n-nikol98:
On my system I have two drives mounted under `/mnt/Data` and `/mnt/Data 2`.

I noticed that browsing to the latter drive in Nautilus would also produce an open indicator in the dock for the former drive.

This simple tweak fixes the problem and should prevent issues of a similar nature from occurring.

### DIFF
--- a/fileManager1API.js
+++ b/fileManager1API.js
@@ -81,8 +81,14 @@ var FileManager1Client = class DashToDock_FileManager1Client {
      */
     getWindows(location) {
         let ret = new Set();
+    	let locationEsc = location;
+	    
+    	if (!location.endsWith('/')) { 
+		locationEsc += '/'; 
+	}
+	    
         for (let [k,v] of this._locationMap) {
-            if ((k + '/').startsWith(location + '/')) {
+            if ((k + '/').startsWith(locationEsc)) {
                 for (let l of v) {
                     ret.add(l);
                 }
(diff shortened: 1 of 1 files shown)


fileManager1API.js
```diff
@@ -82,7 +82,7 @@ var FileManager1Client = class DashToDock_FileManager1Client {
     getWindows(location) {
         let ret = new Set();
         for (let [k,v] of this._locationMap) {
-            if (k.startsWith(location)) {
+            if ((k + '/').startsWith(location + '/')) {
```
philipl, line 85: Can we say for sure that neither k nor location will ever end with a '/'? This logic will break if that happens.

n-nikol98: Thanks for the review.

This would indeed break the trash indicator if you browse to a trashed directory, which results in a `k` value of: `trash:///folder1` and a `location` of: `trash:///`.

I've added a check to make sure that `location` doesn't end with a '/'.

`k` will end with a '/' if you browse to the file-system root (e.g `file:////`), the trash root (`trash:////`) or the root of some other location, but I cant't see how that would break the current logic.

philipl (approved):
Thanks!

philipl:
@3v1n0 Can you merge?

micheleg (approved):

PR #1333, opened 2020-10-19
change .shrink's border to 0px

magicalfeyfenny:
In panel mode + shrunken padding mode + custom theme mode, the panel does not cover the corner pixel on one side due to the border being 1px. An example of this bug is here: https://queer.af/@fenny/105063613088716362

I don't know if this is an issue outside of Pop_OS! 20.04

### DIFF
--- a/stylesheet.css
+++ b/stylesheet.css
@@ -1,7 +1,7 @@
 /* Shrink the dash by reducing padding */
 #dashtodockContainer.shrink #dash,
 #dashtodockContainer.dashtodock #dash {
-    border:1px;
+    border:0px;
     padding:0px;
 }
 


micheleg:
The built-in theme customization are mostrly based on the default Gnome Shell theme, which has a border. Trying to account for all possible situation wouldbe a be impossible. Would removing the border rule fix the problem for you? That should achieve the desired effect both with the default theme (have a borders) and custom ones (keep the existing border-width)

stylesheet.css
```diff
@@ -1,7 +1,7 @@
 /* Shrink the dash by reducing padding */
 #dashtodockContainer.shrink #dash,
 #dashtodockContainer.dashtodock #dash {
-    border:1px;
+    border:0px;
```
micheleg, line 4:
```suggestion
```

PR #1346, opened 2020-11-09
appIcons: Add 'Focus, minimize or show previews'

l3nn4rt:
This action mixes the behaviour of 'Focus or show previews' and
'Minimize or show previews', ie. it acts as follows:

- if app is not running, create new window
- if app is running but not focused, focus (first) window
- if app is running and focused:
  - if app has multiple windows, show previews
  - otherwise, minimize window

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -86,6 +86,7 @@
                           <item translatable="yes">Show window previews</item>
                           <item translatable="yes">Minimize or show previews</item>
                           <item translatable="yes">Focus or show previews</item>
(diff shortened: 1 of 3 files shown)


edrisranjbar:
It is a very good feature to have. i hope it'll be merged soon.

skaldek (approved):
Good feature.

micheleg (approved):

micheleg:
Does not harm to have it as an option.

PR #1361, opened 2020-12-20
Quick Fix #1359

skaldek:
Quick fix #1359 causing application screen flickering.

PR #1369, opened 2021-01-10
Fix grid flashing on app grid animation

### LINKED ISSUE #1359: dash to dock flickering gnome 3.38
**Gnome-Shell** 3.38.2 , **Mutter** version 3.38.2 , running on Wayland.

WhatsApp-Video-2020-12-08-at-15 48 19

l3nn4rt:
This fixes the app grid flashing when the button is pressed from outside the overview.

It is still happening first time the apps page is opened from the workspaces page. This may be related to the shell itself.

It seems to work as expected (except for the case above) on `GNOME Shell 3.38.2`.

Edit: fixes #1359.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -1883,36 +1883,18 @@ var DockManager = class DashToDock_DockManager {
                 if (!Main.overview._shown) {
                     this._forcedOverview = true;
                     let grid = visibleView._grid;
-                    if (animate) {
-                        // Animate in the the appview, hide the appGrid to avoiud flashing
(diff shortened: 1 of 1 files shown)


seitbekir:
Used this code on my Manjaro with GNOME 3.38.3 Wayland. It works, The only comment is that windows disappearing is not smooth as top-bar, for example. But it is unrelated.

seitbekir (approved):
Checked on Gnome 3.38.3 on Wayland (Manjaro). Works fine.

vanvugt:
It looks like the same bug was already fixed in #1361.

PR #1416, opened 2021-04-17
Add drag-and-drop assist.

Valent-in:
Switch windows by dragging and holding item on dock.
Cycle windows if more than one. Launch app if pinned but not running.
Issue 278


Valent-in:
Does not work in Wayland. Dragging over icon emits ::hover, but I can not grab mouse modifier button by event.get_state_full()
https://gjs-docs.gnome.org/clutter4~4_api/clutter.event#method-get_state_full

I don't like this
`let scale = St.ThemeContext.get_for_stage(Meta.get_backend().get_stage()).scale_factor;`
But screen scaling factor does not applies to mouse coordinates. Probably a bug in toolkit.

dash.js
```diff
@@ -527,6 +527,72 @@ var MyDash = GObject.registerClass({
             }
         });
 
+        let switchTimerId = { id: 0 };
+        let isDndTarget = false;
+        let scale = St.ThemeContext.get_for_stage(Meta.get_backend().get_stage()).scale_factor;
```
3v1n0, line 532: You can get the sage via `global.stage`.

Valent-in: Thanks.

SIMULATAN:
Exactly what I need, just came here to check if there was any issue regarding this behavior.
Is there any possibility of a merge?

Valent-in:
Rebase to upstream. Merge conflicts resolved.
Also removed mouse coordinates scaling as it has been fixed in shell toolkit (tested only on version 41).

skoruppa (approved):

CosmicFusion:
Please add this, it will improve usability of gnome 100%

CosmicFusion:
Just tried this patch on ubuntu 22.10 gnome 43 wayland, it's awesome can't live without it

BarbzYHOOL:
@3v1n0 this looks awesome!!

sglbl:
Is there any update on this pull request?

PR #1423, opened 2021-04-23
Fixes #1421

### LINKED ISSUE #1421: Bug: The dock is shown outside of the activities view even when "Intelligent Autohide" enabled and "Intelligent Autohide->Autohide" disabled.
If you enable **Intelligent Autohide** but disable **Intelligent Autohide-> Autohide** (inside the **Intelligent Autohide** settings), the dock is shown when hovering the mouse over the edge of the screen.

Intelligent Autohide enabled
Autohide disabled dock shown

LorienLV:


3v1n0:
Some more explicative commit message would be appreciated.

LorienLV:
After disabling the intelligent autohide, the barrier is not updated, so the dock is still automatically shown. Now the updateBarrier() method is called when enabling/disabling the intelligent auto-hide.

3v1n0:
> After disabling the intelligent autohide, the barrier is not updated, so the dock is still automatically shown. Now the updateBarrier() method is called when enabling/disabling the intelligent auto-hide.

This would be nice, could you please amend the previous commit and force push here?

3v1n0:
This can be closed as was fixed as part of commit https://github.com/micheleg/dash-to-dock/commit/5bb37df467468ba8b9e491e4e616725821ad0ebb

PR #1454, opened 2021-06-04
Add new setting: preview size scale

kitzberger:
This adds a new option to the configuration interface to manipulate the preview size scale:
* scale from 0,00 to 1,00 in steps of 0,05
* default is 0,00 a.k.a. "auto" which is the current scale calculation

I've implemented this since the current scale calculation results in a scale of 0,1429 for my monitor which is waaaaay too small IMHO. I prefer having a scale of 0,40.

Bildschirmfoto vom 2021-06-04 11-38-49

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -574,6 +574,12 @@
     <property name="step_increment">1</property>
     <property name="page_increment">10</property>
   </object>
+  <object class="GtkAdjustment" id="preview_size_adjustment">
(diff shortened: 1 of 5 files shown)


3v1n0 (approved):
Looks good, maybe it could be also the case to improve the current scaling algorithm to take care of the monitor size, but this seems already an improvement to me.

PR #1461, opened 2021-06-20
[GNOME 40/DEPRECATION] Replaced deprecated property calls `margin_left/right` with proper `margin_start/end`

KoNickss:
**More info about the deprecation can be found here:** 

- https://valadoc.org/gtk+-3.0/Gtk.Widget.margin_left.html
- https://valadoc.org/gtk+-3.0/Gtk.Widget.margin_right.html


KoNickss:
ah fuck wrong repo

PR #1486, opened 2021-07-19
Remove version sed from Makefile

carlwgeorge:
Since version 63, the version has been set explicitly in metadata.json.

651234d42d1c94240b4aa528dfddf6f890d2f8a7


vanvugt:
Sorry this got ignored. It appears the `zip-file` target still relies on `$(VSTRING)` being set so this doesn't appear to be a safe change.

PR #1531, opened 2021-10-11
Add support to GNOME 41

3v1n0:
Include in our codebase some of the code that has been removed from GNOME 41 to handle the menus, and update them so that they will include the new features (such as better switcheroo management and parental control).

This is just a safer implementation so that we can cover both GNOME 40 and 41 without huge refactories.

I've locally [some experiments](https://github.com/3v1n0/dash-to-dock/commit/e403e16fd72c1d69a339c28f034bade0d9514907) to be able to use upstream's AppMenu as base for our menus, but I feel that the change would be to risky, and don't think would lead to much benefit either, given that we're already patching the menus quite a lot.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,7 @@
 {
 "shell-version": [
-    "40"
+    "40",
+    "41"
 ],
 "uuid": "dash-to-dock@micxgx.gmail.com",
 "name": "Dash to Dock",


rastersoft:
I tested this in Debian Sid with Gnome Shell 41, and works fine.

rkitover:
Works for me on NixOS unstable with Gnome 41.

rkitover:
Note for anyone installing this, change `"version"` in `metadata.json` to a value higher than `70` or GNOME will try to "upgrade" it to the current version.

PR #1541, opened 2021-10-15
Desktop icons integration

rastersoft:
When Dash to Dock is in autohide mode, Desktop Icons NG (DING) can't know how much space the dock need, so icons can be put below the dock and become inaccessible.

This is a generic fix for this. It adds a class (DingIntegration) that allows other extensions to communicate with Ding and any other extension that would need to know variable margins to avoid to put elements there.

The other part is at https://gitlab.com/rastersoft/desktop-icons-ng/-/merge_requests/196

### DIFF
--- a/desktopIconsIntegration.js
+++ b/desktopIconsIntegration.js
@@ -1,0 +1,159 @@
+/*
+ * The code in this file is distributed under a "1-clause BSD license",
+ * which makes it compatible with GPLv2 and GPLv3 too, and others.
+ *
+ * License text:
+ *
+ * Copyright (C) 2021 Sergio Costas (rastersoft@gmail.com)
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/*******************************************************************************
+ * Integration class
+ *
+ * This class must be added to other extensions in order to integrate
+ * them with Desktop Icons NG. It allows an extension to notify how much margin
+ * it uses in each side of each monitor.
+ *
+ * DON'T SEND PATCHES TO THIS FILE TO THE EXTENSION MAINTAINER. SEND THEM TO
+ * DESKTOP ICONS NG MAINTAINER: https://gitlab.com/rastersoft/desktop-icons-ng
+ *
+ * In the *enable()* function, create a *DesktopIconsUsableAreaClass()*
+ * object with
+ *
+ *     new DesktopIconsIntegration.DesktopIconsUsableAreaClass(object);
+ *
+ * Now, in the *disable()* function just call to the *destroy()* method before
+ * nullifying the pointer. You must create a new object in enable() the next
+ * time the extension is enabled.
+ *
+ * In your code, every time you change the margins, you should call first to
+ * *resetMargins()* method to clear the current margins, and then call to
+ * *setMargins(...)* method as many times as you need to set the margins in each
+ * monitor. You don't need to call it for all the monitors, only for those where
+ * you are painting something. If you don't set values for a monitor, they will
+ * be considered zero.
+ *
+ * The margins values are relative to the monitor border.
+ *
+ *******************************************************************************/
+
+const GLib = imports.gi.GLib;
+const Main = imports.ui.main;
+
+const ExtensionUtils = imports.misc.extensionUtils;
+const Me = ExtensionUtils.getCurrentExtension();
+
+const IDENTIFIER_UUID = "130cbc66-235c-4bd6-8571-98d2d8bba5e2";
+
+var DesktopIconsUsableAreaClass = class {
+    constructor() {
+        this._extensionManager = Main.extensionManager;
+        this._timedMarginsID = 0;
+        this._margins = {};
+        this._emID = this._extensionManager.connect('extension-state-changed', (_obj, extension) => {
+            if (!extension)
+                return;
+
+            // If an extension is being enabled and lacks the DesktopIconsUsableArea object, we can avoid launching a refresh
+            if (extension.state === ExtensionUtils.ExtensionState.ENABLED) {
+                this._sendMarginsToExtension(extension);
+                return;
+            }
+            // if the extension is being disabled, we must do a full refresh, because if there were other extensions originally
+            // loaded after that extension, those extensions will be disabled and enabled again without notification
+            this._changedMargins();
+        });
+    }
+
+    /**
+     * Sets or updates the top, bottom, left and right margins for a
+     * monitor. Values are measured from the monitor border (and NOT from
+     * the workspace border).
+     *
+     * @param {int} monitor Monitor number to which set the margins.
+     *                      A negative value means "the primary monitor".
+     * @param {int} top Top margin in pixels
+     * @param {int} bottom Bottom margin in pixels
+     * @param {int} left Left margin in pixels
+     * @param {int} right Right margin in pixels
+     */
+    setMargins(monitor, top, bottom, left, right) {
+        this._margins[monitor] = {
+            'top': top,
+            'bottom': bottom,
+            'left': left,
+            'right': right
+        };
+        this._changedMargins();
+    }
+
+    /**
+     * Clears the current margins. Must be called before configuring the monitors
+     * margins with setMargins().
+     */
+    resetMargins() {
+        this._margins = {};
+        this._changedMargins();
+    }
+
+    /**
+     * Disconnects all the signals and removes the margins.
+     */
+    destroy() {
+        if (this._emID) {
+            this._extensionManager.disconnect(this._emID);
+            this._emID = 0;
+        }
+        if (this._timedMarginsID) {
+            GLib.source_remove(this._timedMarginsID);
+            this._timedMarginsID = 0;
+        }
+        this._margins = null;
+        this._changedMargins();
+    }
+
+    _changedMargins() {
+        if (this._timedMarginsID) {
+            GLib.source_remove(this._timedMarginsID);
+        }
+        this._timedMarginsID = GLib.timeout_add(GLib.PRIORITY_DEFAULT, 100, ()=> {
+            this._sendMarginsToAll();
+            this._timedMarginsID = 0;
+            return GLib.SOURCE_REMOVE;
+        });
+    }
+
+    _sendMarginsToAll() {
+        this._extensionManager.getUuids().forEach(uuid =>
+            this._sendMarginsToExtension(this._extensionManager.lookup(uuid)));
+    }
+
+    _sendMarginsToExtension(extension) {
+        // check that the extension is an extension that has the logic to accept
+        // working margins
+        if (extension?.state !== ExtensionUtils.ExtensionState.ENABLED)
+            return;
+
+        const usableArea = extension?.stateObj?.DesktopIconsUsableArea;
+         if (usableArea?.uuid === IDENTIFIER_UUID)
+            usableArea.setMarginsForExtension(Me.uuid, this._margins);
+    }
+}
--- a/docking.js
+++ b/docking.js
@@ -1792,6 +1792,13 @@ var DockManager = class DashToDock_DockManager {
             this._settings,
             'changed::isolate-locations',
             () => this._ensureLocations()
+        ], [
+            this._settings,
+            'changed::intellihide',
+            () => {
+                if (!this._settings.intellihide)
+                    this._desktopIconsUsableArea.resetMargins();
+            }
         ]);
     }
 
--- a/docking.js
+++ b/docking.js
@@ -34,6 +34,7 @@ const DockDash = Me.imports.dash;
 const Locations = Me.imports.locations;
 const LauncherAPI = Me.imports.launcherAPI;
 const FileManager1API = Me.imports.fileManager1API;
+const DesktopIconsIntegration = Me.imports.desktopIconsIntegration;
 
 const DOCK_DWELL_CHECK_INTERVAL = 100;
 
@@ -548,7 +549,10 @@ var DockedDash = GObject.registerClass({
         ], [
             settings,
             'changed::intellihide',
-            this._updateVisibilityMode.bind(this)
+            () => {
+                this._updateVisibilityMode();
+                this._updateVisibleDesktop();
+            }
         ], [
             settings,
             'changed::intellihide-mode',
@@ -1102,6 +1106,21 @@ var DockedDash = GObject.registerClass({
         }
     }
 
+    _updateVisibleDesktop() {
+        if (!this._intellihideIsEnabled)
+            return;
+
+        const { desktopIconsUsableArea } = DockManager.getDefault();
+        if (this._position === St.Side.BOTTOM)
+            desktopIconsUsableArea.setMargins(this._monitorIndex, 0, this._box.height, 0, 0);
+        else if (this._position === St.Side.TOP)
+            desktopIconsUsableArea.setMargins(this._monitorIndex, this._box.height, 0, 0, 0);
+        else if (this._position === St.Side.RIGHT)
+            desktopIconsUsableArea.setMargins(this._monitorIndex, 0, 0, 0, this._box.width);
+        else if (this._position === St.Side.LEFT)
+            desktopIconsUsableArea.setMargins(this._monitorIndex, 0, 0, this._box.width, 0);
+    }
+
     _updateStaticBox() {
         this.staticBox.init_rect(
             this.x + this._slider.x - (this._position == St.Side.RIGHT ? this._box.width : 0),
@@ -1111,6 +1130,7 @@ var DockedDash = GObject.registerClass({
         );
 
         this._intellihide.updateTargetBox(this.staticBox);
+        this._updateVisibleDesktop();
     }
 
     _removeAnimations() {
@@ -1546,6 +1566,7 @@ var DockManager = class DashToDock_DockManager {
         this._vfuncInjections = new Utils.VFuncInjectionsHandler(this);
         this._propertyInjections = new Utils.PropertyInjectionsHandler(this);
         this._settings = ExtensionUtils.getSettings('org.gnome.shell.extensions.dash-to-dock');
+        this._desktopIconsUsableArea = new DesktopIconsIntegration.DesktopIconsUsableAreaClass();
         this._oldDash = Main.overview.isDummy ? null : Main.overview.dash;
 
         // Connect relevant signals to the toggling function
@@ -1608,6 +1629,10 @@ var DockManager = class DashToDock_DockManager {
 
     get trash() {
         return this._trash;
+    }
+
+    get desktopIconsUsableArea() {
+        return this._desktopIconsUsableArea;
     }
 
     getDockByMonitor(monitorIndex) {
@@ -2130,6 +2155,7 @@ var DockManager = class DashToDock_DockManager {
         // Remove extra features
         this._workspaceIsolation.destroy();
         this._keyboardShortcuts.destroy();
+        this._desktopIconsUsableArea.resetMargins();
 
         // Delete all docks
         this._allDocks.forEach(d => d.destroy());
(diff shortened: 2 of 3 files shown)


3v1n0:
Hey @rastersoft do you have progress on this?

I'd like to have it in by 22.04, so let me know if you need help

rastersoft:
Hi. I just ended a simplified code for the integration. It is here: https://gitlab.com/rastersoft/desktop-icons-ng/-/merge_requests/212

Can you give me your opinion to choose between that and the original idea? ( https://gitlab.com/rastersoft/desktop-icons-ng/-/merge_requests/196 )

rastersoft:
It's ready.

rastersoft:
Ok, NOW the MR is ready to be merged.

desktopIconsIntegration.js
```diff
@@ -0,0 +1,163 @@
...
+ * *setMargins(...)* method as many times as you need to set the margins in each
+ * monitor. You don't need to call it for all the monitors, only for those where
+ * you are painting something. If you don't set values for a monitor, they will
+ * be considered zero.
+ *
+ * The margins values are relative to the monitor border.
+ *
+ *******************************************************************************/
+
+const Mainloop = imports.mainloop;
```
3v1n0, line 58: Kind of deprecated upstream, can we just use GLIb.Sources directly?

desktopIconsIntegration.js
```diff
@@ -0,0 +1,163 @@
...
+const GLib = imports.gi.GLib;
+const Main = imports.ui.main;
+
+const ExtensionUtils = imports.misc.extensionUtils;
+const Me = ExtensionUtils.getCurrentExtension();
+
+var DesktopIconsUsableAreaClass = class {
+    constructor() {
+        this._extensionManager = Main.extensionManager;
+        this._extensionUUID = "130cbc66-235c-4bd6-8571-98d2d8bba5e2";
```
3v1n0, line 68: Can't you just use `Me.uuid` ?

desktopIconsIntegration.js
```diff
@@ -0,0 +1,163 @@
...
+     * @param {int} bottom Bottom margin in pixels
+     * @param {int} left Left margin in pixels
+     * @param {int} right Right margin in pixels
+     */
+    setMargins(monitor, top, bottom, left, right) {
+        this._margins[monitor] = {
+            'top': top,
+            'bottom': bottom,
+            'left': left,
+            'right': right
```
3v1n0, line 103: Why not using `Clutter.Margin`?

desktopIconsIntegration.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+        this._margins = null;
+        this._changedMargins();
+    }
+
+    _changedMargins() {
+        if (this._timedMarginsID) {
+            GLib.source_remove(this._timedMarginsID);
+        }
+        this._timedMarginsID = Mainloop.timeout_add(250, ()=> {
```
3v1n0, line 137: Having a timeout is better than an idle here? Do we require more time?

desktopIconsIntegration.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    _sendMarginsToExtension(extension) {
+        // check that the extension is an extension that has the logic to accept
+        // working margins
+        if ((!extension) ||
+            (extension.state != ExtensionUtils.ExtensionState.ENABLED) ||
+            (!extension.stateObj) ||
+            (!extension.stateObj.DesktopIconsUsableArea) ||
+            (extension.stateObj.DesktopIconsUsableArea.uuid != this._extensionUUID)) {
```
3v1n0, line 158: `extension?.stateObj?.DesktopIconsUsableArea?.uuid == this._extensionUUID`

docking.js
```diff
@@ -1096,6 +1105,24 @@ var DockedDash = GObject.registerClass({
...
+        this._updateVisibleDesktop();
+    }
+
+    _updateVisibleDesktop() {
+        if (!DockManager.settings.get_boolean('intellihide'))
+            return;
+        let monitor = this._monitorIndex;
+        if (this._isHorizontal) {
+            if (this._position == St.Side.BOTTOM)
+                Me.imports.extension.DesktopIconsUsableArea.setMargins(monitor, 0, this._box.height, 0, 0);
```
3v1n0, line 1117: Ad a `const` definition for this since we're reusing it multiple times

extension.js
```diff
@@ -3,19 +3,24 @@
...
 // GNOME Shell 3.26+ (mozjs52+).
 var dockManager;
+var DesktopIconsUsableArea;
 
 function init() {
     ExtensionUtils.initTranslations('dashtodock');
 }
 
 function enable() {
+    DesktopIconsUsableArea = new DesktopIconsIntegration.DesktopIconsUsableAreaClass();
```
3v1n0, line 18: Use DockManager for this

3v1n0 (changes requested):

desktopIconsIntegration.js
rastersoft: The idea is to reduce the number of times the margins are sent to all the extensions. Maybe an idle could work too...

desktopIconsIntegration.js
rastersoft: No. This UUID is used to ensure that the "destination extension" does really offer this interface: only if an extension has precisely this UUID "published", we will assume that it can accept "margins info".

desktopIconsIntegration.js
rastersoft: You are right, I should use GLib.add_timeout() instead of Mainloop.add_timeout(). Fixed.

desktopIconsIntegration.js
rastersoft: Yes, it would work too. But since it's an internal element, I didn't see a reason to use an "external" structure.

desktopIconsIntegration.js
rastersoft: Nice! Fixed.

extension.js
rastersoft: I don't understand. DockManager?

docking.js
rastersoft: Done.

rastersoft: But not exactly as you probably expected, because I'm accessing the actual object, not the file, so I can't define a CONST at the beginning, because the object changes each time the extension is disabled and enabled again, and also the code must wait until the object is created to be able to use it.

desktopIconsIntegration.js
3v1n0: Ok, I suppose normally that can be handled by an idle, but if you say it could not be the case, that's fine.

Maybe can be reduced to 100ms, it seems just more than enough.

rastersoft: Done.

extension.js
3v1n0: I'm saying to handle the lifecycle of this object in `Docking.DockManager()` (and `Docking.destroy()`).

If you need to expose this as part of the extension, just get it from that class via a property... I'd prefer to have all managed there.

desktopIconsIntegration.js
3v1n0: Ok... I imagined it by reading further, but...

Please expose this as a const in the top of the file instead, with a clearer name as otherwise it seems related to what this extension exposes of itself.

desktopIconsIntegration.js
```diff
@@ -0,0 +1,160 @@
...
+            let extension = this._extensionManager.lookup(uuid);
+            this._sendMarginsToExtension(extension);
+        }
+    }
+
+    _sendMarginsToExtension(extension) {
+        // check that the extension is an extension that has the logic to accept
+        // working margins
+        if ((!extension) ||
+            (extension.state != ExtensionUtils.ExtensionState.ENABLED) ||
```
3v1n0, line 154: Also here can just use `?.` operator here

desktopIconsIntegration.js
```diff
@@ -0,0 +1,160 @@
...
+            (extension.stateObj?.DesktopIconsUsableArea?.uuid != this._extensionUUID)) {
+                return;
```
3v1n0, line 156: You need to use inverse logic now, otherwise this would also match when such fields are not present (as `undefined !== this._extensionUUID`).

In any case use `===` operator and not just equality one.

docking.js
```diff
@@ -1096,6 +1105,25 @@ var DockedDash = GObject.registerClass({
...
+        this._updateVisibleDesktop();
+    }
+
+    _updateVisibleDesktop() {
+        if (!DockManager.settings.get_boolean('intellihide'))
+            return;
+        let monitor = this._monitorIndex;
+        let DesktopIconsUsableArea = Me.imports.extension.DesktopIconsUsableArea;
+        if (this._isHorizontal) {
+            if (this._position == St.Side.BOTTOM)
```
3v1n0, line 1117: `===`

docking.js
```diff
@@ -1096,6 +1105,25 @@ var DockedDash = GObject.registerClass({
...
+                DesktopIconsUsableArea.setMargins(monitor, 0, this._box.height, 0, 0);
+            else
+                DesktopIconsUsableArea.setMargins(monitor, this._box.height, 0, 0, 0);
+        } else {
+            if (this._position == St.Side.RIGHT)
```
3v1n0, line 1122: `===`

desktopIconsIntegration.js
```diff
@@ -0,0 +1,160 @@
...
+        this._extensionManager = Main.extensionManager;
+        this._extensionUUID = "130cbc66-235c-4bd6-8571-98d2d8bba5e2";
+        this._timedMarginsID = 0;
+        this._margins = {};
+        this._emID = this._extensionManager.connect('extension-state-changed', (obj, extension) => {
+            if (!extension) {
+                return;
+            }
+            // If an extension is being enabled and lacks the DesktopIconsUsableArea object, we can avoid launching a refresh
+            if (extension.state == ExtensionUtils.ExtensionState.ENABLED) {
```
3v1n0, line 75: `===`

extension.js
rastersoft: Done.

desktopIconsIntegration.js
rastersoft: Done.

desktopIconsIntegration.js
rastersoft: Done.

desktopIconsIntegration.js
rastersoft: That's the idea: if it is undefined, it must return.
Done.

docking.js
rastersoft: Done.

desktopIconsIntegration.js
rastersoft: Done.

docking.js
rastersoft: Done.

docking.js
```diff
@@ -1096,6 +1106,25 @@ var DockedDash = GObject.registerClass({
...
             }
         }
+        this._updateVisibleDesktop();
+    }
+
+    _updateVisibleDesktop() {
+        if (!DockManager.settings.get_boolean('intellihide'))
+            return;
+        let monitor = this._monitorIndex;
+        let desktopIconsUsableArea = Me.imports.extension.dockManager.desktopIconsUsableArea;
```
3v1n0, line 1116:
```suggestion
        const monitor = this._monitorIndex;
        const desktopIconsUsableArea = Me.imports.extension.dockManager.desktopIconsUsableArea;
```

You could probably avoid the `monitor` one at all.

desktopIconsIntegration.js
```diff
@@ -0,0 +1,160 @@
...
+            this._timedMarginsID = 0;
+            return GLib.SOURCE_REMOVE;
+        });
+    }
+
+    _sendMarginsToAll() {
+        for (let uuid of this._extensionManager.getUuids()) {
+            let extension = this._extensionManager.lookup(uuid);
+            this._sendMarginsToExtension(extension);
+        }
```
3v1n0, line 148:
```suggestion
        this._extensionManager.getUuids().forEach(uuid =>
            this._sendMarginsToExtension(this._extensionManager.lookup(uuid)):
```

desktopIconsIntegration.js
```diff
@@ -0,0 +1,160 @@
...
+ *
+ *******************************************************************************/
+
+const GLib = imports.gi.GLib;
+const Main = imports.ui.main;
+
+const ExtensionUtils = imports.misc.extensionUtils;
+const Me = ExtensionUtils.getCurrentExtension();
+
+const IdentifierUUID = "130cbc66-235c-4bd6-8571-98d2d8bba5e2";
```
3v1n0, line 64:
```suggestion
const IDENTIFIER_UUID = "130cbc66-235c-4bd6-8571-98d2d8bba5e2";
```

desktopIconsIntegration.js
```diff
@@ -0,0 +1,160 @@
...
+
+var DesktopIconsUsableAreaClass = class {
+    constructor() {
+        this._extensionManager = Main.extensionManager;
+        this._timedMarginsID = 0;
+        this._margins = {};
+        this._emID = this._extensionManager.connect('extension-state-changed', (obj, extension) => {
+            if (!extension) {
+                return;
+            }
```
3v1n0, line 74:
```suggestion
            if (!extension)
                return;

```

desktopIconsIntegration.js
```diff
@@ -0,0 +1,160 @@
...
+const Me = ExtensionUtils.getCurrentExtension();
+
+const IdentifierUUID = "130cbc66-235c-4bd6-8571-98d2d8bba5e2";
+
+var DesktopIconsUsableAreaClass = class {
+    constructor() {
+        this._extensionManager = Main.extensionManager;
+        this._timedMarginsID = 0;
+        this._margins = {};
+        this._emID = this._extensionManager.connect('extension-state-changed', (obj, extension) => {
```
3v1n0, line 71:
```suggestion
        this._emID = this._extensionManager.connect('extension-state-changed', (_obj, extension) => {
```

desktopIconsIntegration.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    _sendMarginsToExtension(extension) {
+        // check that the extension is an extension that has the logic to accept
+        // working margins
+        if ((extension?.state !== ExtensionUtils.ExtensionState.ENABLED) ||
+            (extension?.stateObj?.DesktopIconsUsableArea?.uuid !== IdentifierUUID)) {
+                return;
+        }
+        extension.stateObj.DesktopIconsUsableArea.setMarginsForExtension(Me.uuid, this._margins);
```
3v1n0, line 158: I'd like this style better :stuck_out_tongue: 

```suggestion
        if (extension?.state !== ExtensionUtils.ExtensionState.ENABLED))
            return;

        const usableArea = extension?.stateObj?.DesktopIconsUsableArea;
         if (usableArea?.uuid === INDENTIFIER_UUID))
                usableArea.setMarginsForExtension(Me.uuid, this._margins);
```

docking.js
```diff
@@ -1534,6 +1564,7 @@ var DockManager = class DashToDock_DockManager {
 
         Me.imports.extension.dockManager = this;
 
+        this.desktopIconsUsableArea = new DesktopIconsIntegration.DesktopIconsUsableAreaClass();
```
3v1n0, line 1567:
```suggestion
        this._desktopIconsUsableArea = new DesktopIconsIntegration.DesktopIconsUsableAreaClass();
```

I'm wondering if it would be relevant to add this on request in case of intellihide being enabled. So in case `this._desktopIconsUsableArea` is defined we can update it otherwise we ignore the changes.

docking.js
```diff
@@ -2105,6 +2136,7 @@ var DockManager = class DashToDock_DockManager {
     }
 
     _deleteDocks() {
+        this.desktopIconsUsableArea.resetMargins();
```
3v1n0, line 2139: I'd prefer this to be in `toggle()` instead.

docking.js
rastersoft: Done.

desktopIconsIntegration.js
rastersoft: Mmm... why?

3v1n0: Was not used, so that's the syntax we use in such way... At least, that's what eslint suggests.

docking.js
rastersoft: The point is that this object is accessed from outside, from the DashToDock class (line 1115 and nexts), so it should be "public".

docking.js
rastersoft: But they aren't "toggled", but "erased".

desktopIconsIntegration.js
rastersoft: Oh, good point.

docking.js
3v1n0: Ok that's fine, but in such cases I'd prefer to have a `get desktopIconsUsableArea()`, so that it's clear that can't be set outside. Just to follow the same we're doing with other objects dockManager exposes.

rastersoft: Good point. One second...

docking.js
3v1n0: Yeah, I know... But I'm saying with this to handle it in the caller function, as `_deleteDocks` is relevant to docks only. 

However, I'm wondering if we couldn't just handle it when destroying each dock, so basically ensure that `desktopIconsUsableArea.setMargins(this._monitorIndex, 0, 0, 0, 0);` is called on each dock's destroy().

docking.js
rastersoft: Done.

docking.js
rastersoft: I wanted to keep the API as simple as possible. That's why each time you want to remove one element, you have to delete all and reinsert the remainder ones. Also, that's another reason why I used a timer instead of the idle.

desktopIconsIntegration.js
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @param {int} monitor Monitor number to which set the margins.
+     *                      A negative value means "the primary monitor".
+     * @param {int} top Top margin in pixels
+     * @param {int} bottom Bottom margin in pixels
+     * @param {int} left Left margin in pixels
+     * @param {int} right Right margin in pixels
+     */
+    setMargins(monitor, top, bottom, left, right) {
+        this._margins[monitor] = {
```
3v1n0, line 99: As per https://github.com/micheleg/dash-to-dock/pull/1541#discussion_r790048572, I'd say we should do:

```suggestion
        if (top <= 0 && bottom <= 0 && left  <= 0 && right  <= 0) {
            delete this._margins[monitor];
            this._changedMargins();
         }
        this._margins[monitor] = {
```

rastersoft: As I said, I prefer to keep the API as simple as possible. Doing this would mean changing the API itself, so I think it is not a good idea (remember that this file is also used by dash-to-panel and hide-top-bar).

3v1n0: I see, would it change the result though?

3v1n0: There's still something I'm not sure about this, as with the current `DesktopIconsUsableAreaClass` code in case a single dock is removed (like a disconnected monitor) there would not be a way to notify a single change?

It's true that right now dash-to-dock delete all the docks on monitors-changed, but could actually change in theory (or in future), as it could just add / move docks around. In such case, will it break DesktopIcons?

docking.js
```diff
@@ -1096,6 +1106,24 @@ var DockedDash = GObject.registerClass({
                 this.remove_style_class_name('extended');
             }
         }
+        this._updateVisibleDesktop();
+    }
+
+    _updateVisibleDesktop() {
+        if (!DockManager.settings.get_boolean('intellihide'))
```
3v1n0, line 1113: This was meant to be

```suggestion
        if (!ockManager.settings.get_boolean('intellihide'))
```

Doesn't it?

docking.js
```diff
@@ -488,11 +489,17 @@ var DockedDash = GObject.registerClass({
         ], [
             settings,
             'changed::dash-max-icon-size',
-            () => { this.dash.setIconSize(settings.get_int('dash-max-icon-size')); }
+            () => {
+                this.dash.setIconSize(settings.get_int('dash-max-icon-size'));
+                this._updateVisibleDesktop();
```
3v1n0, line 494: This change is not needed as will be triggered by a reallocation anyways.

docking.js
```diff
@@ -488,11 +489,17 @@ var DockedDash = GObject.registerClass({
...
+            }
         ], [
             settings,
             'changed::icon-size-fixed',
-            () => { this.dash.setIconSize(settings.get_int('dash-max-icon-size')); }
+            () => {
+                this.dash.setIconSize(settings.get_int('dash-max-icon-size'));
+                this._updateVisibleDesktop();
```
3v1n0, line 501: This change is not needed as will be triggered by a reallocation anyways.

docking.js
```diff
@@ -544,7 +545,10 @@ var DockedDash = GObject.registerClass({
         ], [
             settings,
             'changed::intellihide',
-            this._updateVisibilityMode.bind(this)
+            () => {
+                this._updateVisibilityMode();
+                this._updateVisibleDesktop();
```
3v1n0, line 550: In this case, if intellihide is disabled, shouldn't we reset the usable area?

It's not probably something that I'd like to do here though, but rather in the DockManager

desktopIconsIntegration.js
rastersoft: No, there isn't possible to notify a single change. I preferred to keep the interface between extensions as simple as possible, so every time there is a change, the whole current status is passed. Since I wanted to also allow this interface to be used by other extensions, not only by DING (let's say: an extension that allows to paint widgets in the background with CPU/disk,memory/network/whatever usage, could also be interested in know where to not paint them to avoid being occluded by a dock/bar; using the same class than in DING would allow it without having to implement another set of MRs in dash-to-dock, dash-to-panel, hide-top-bar, and any other extension that uses this), so keeping a simple interface was paramount to avoid other programmers to back off.

docking.js
rastersoft: I, personally, would prefer to keep them, because I always have preferred being explicit than implicit. Also, that was precisely another reason for the timeout when sending the data: to coalesce all these changes in a single one. But, of course, if you consider that they can be safely remove, then we can remove them.

docking.js
rastersoft: That's a good point. I tried to do that, and the result was bad, because there seems to be a delay between the change notification in DockManager object and the notification in each dock (that, or the callback is being called before in the DockManager object than in the docks themselves), so it fails.

Anyway, since the margin values are the same in both cases, it is not needed, because when intellihide is disabled, the margin value will be the same than in the WorkingArea data, and the margin values are presumed to be from the screen border, not from the WorkingArea border.

rastersoft:
Ok, I squashed all the commits.

docking.js
3v1n0: Oh, ok... So you use the monitor size... Mh, this could be ok, even though I assume there could be problem in case the dock si then resized.

I'm wondering if we can just trigger a reset in a clean way in this case... Maybe just connect on the `changed::intellihide` in dockmanager and when it's false, just sending a reset?

3v1n0 (approved):

PR #1548, opened 2021-10-17
Add default.nix for NixOS users.

rkitover:
This allows installation on NixOS with:

nix-shell
make
make install

Signed-off-by: Rafael Kitover <rkitover@gmail.com>


3v1n0 (approved):
I assume it doesn't bother anyone :)

Issue #1553, opened 2021-10-19
Dash background color inverts green/blue


drakkar1969:
When customizing the dash background color, somehow the green and blue channels appear to be inverted.
E.g. setting the background color to #383c4a results in the dash background color being rendered as #384a3c.

Many thanks for the port to GNOME 40 btw, cannot live without this extension!

coreybruce:
Having the same weird issue, what is going on lol

drakkar1969:
Think this bug was introduced with https://github.com/micheleg/dash-to-dock/commit/c9a83048bb9c5e0e1cce5ac24e602df426f19da0

Line 77 of `theming.js` was changed to:
```js
`rgba(${color.red}, ${color.blue}, ${color.green}, ${newAlpha})`;
```

Which should probably be:
```js
`rgba(${color.red}, ${color.green}, ${color.blue}, ${newAlpha})`;
```

PR #1575, opened 2021-11-02
Fix inverted green/blue colors in background

drakkar1969:
Fixes https://github.com/micheleg/dash-to-dock/issues/1553


ZimbiX (approved):

3v1n0 (approved):
Indeed correct... This happens when you test your custom color using a fully red dock :joy:

PR #1576, opened 2021-11-03
locations: Do not manually dispose objects

3v1n0:
Leave the garbage collector do its job, especially on locations where
the object may be shared by multiple instances and its lifecycle can't
be controlled here.

See: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2020#note_1302901


ewlsh (approved):

darkdragon-001:
Should fix #1025, right?

PR #1577, opened 2021-11-04
Show urgent windows despite current workspace (optional)

mrEDitor:


docking.js
```diff
@@ -1487,15 +1476,13 @@ var WorkspaceIsolation = class DashToDock_WorkspaceIsolation {
...
-                'restacked',
-                dock.dash._queueRedisplay.bind(dock.dash)
-            ], [
-                global.window_manager,
-                'switch-workspace',
-                dock.dash._queueRedisplay.bind(dock.dash)
-            ]);
+            this._signalsHandler.addWithLabel(
+                'isolation',
+                [ global.display, 'restacked', dock.dash._queueRedisplay.bind(dock.dash) ],
```
3v1n0, line 1481: I'd like using the arrow functions better than binding now

schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
```diff
@@ -206,6 +206,11 @@
       <summary>Provide workspace isolation</summary>
       <description>Dash shows only windows from the currentworkspace</description>
     </key>
+    <key type="b" name="show-urgent-windows">
```
3v1n0, line 209: Maybe should be called something `always-show-urgent-windows`? Or `workspace-agnostic-urgent-windows`?

appIcons.js
```diff
@@ -1228,11 +1228,14 @@ function getInterestingWindows(windows, monitorIndex) {
...
 
     // When using workspace isolation, we filter out windows
-    // that are not in the current workspace
-    if (settings.get_boolean('isolate-workspaces'))
+    // that are neither in the current workspace nor marked urgent
+    if (settings.get_boolean('isolate-workspaces')) {
+		const showUrgent = settings.get_boolean('show-urgent-windows');
         windows = windows.filter(function(w) {
-            return w.get_workspace().index() == global.workspace_manager.get_active_workspace_index();
+            return w.get_workspace().index() == global.workspace_manager.get_active_workspace_index()
```
3v1n0, line 1235: Since you're changing this, maybe you can move the `global.workspace_manager.get_active_workspace_index()` call to the outside scope? Going to C calls when not needed is better to be avoided, and repeating it for each window it's just uneeded.

3v1n0 (commented):
Looks good overall, just few cleanups.

mrEDitor:
Yes, your suggestion for setting name is definitely better; done.

3v1n0 (approved):

PR #1599, opened 2021-11-23
windowPreview: Prevent an infinite retry loop of _cloneTexture

vanvugt:
Because for some hidden windows it will fail forever, like it does
for jEdit's "JavaEmbeddedFrame".

Such hidden windows are generally X11 windows used for events only,
so not all windows should be expected to ever return a texture.

Fixes: https://launchpad.net/bugs/1949572


3v1n0:
I was using another approach for this already, probably we can integrate them though

vanvugt:
> I was using another approach for this already, probably we can integrate them though

This PR is for master so if you have other fixes then they should be merged there too.

3v1n0:
Yep, mine is on master too. I had it done for some weeks but didn't mark the issue as such, sorry for that.

PR #1613, opened 2021-12-10
Fix urgent flag not being cleared

### LINKED ISSUE #1568: Shaking animation never stops
Whenever an application window is opened without the user clicking on the dock icon or the notification is received, the icon starts shaking. There is nothing wrong with that. But it never stops doing that. Whenever I click on the icon to focus the window, minimize, click again, etc, the icon is still shaking. I have to close the application and open it again to stop the shaking animation.

Version 71, GNOME 40.4.0

https://user-images.githubusercontent.com/17854950/139556297-610ea889-b31c-441d-8e46-d43c838c38e3.mp4

inthreedee:
Fixes #1568 

1. Make sure that when a window is focused, the urgent flag is cleared.  See the above issue for Firefox repro steps.
2. Don't mark focused windows as urgent.  This breaks click to minimize and requires the user to click off and back on the app to clear the shaking animation once the first fix is applied.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -142,8 +142,10 @@ var DockAbstractAppIcon = GObject.registerClass({
         });
 
         this.connect('notify::focused', () => {
-            if (this.focused)
+            if (this.focused) {
                 this.add_style_class_name('focused');
+                this.urgent = false;
+            }
             else
                 this.remove_style_class_name('focused');
         })
@@ -311,7 +313,7 @@ var DockAbstractAppIcon = GObject.registerClass({
     }
 
     _addUrgentWindow(window) {
-        if (this._urgentWindows.has(window))
+        if (this._urgentWindows.has(window) || window.has_focus())
             return;
 
         if (window._manualUrgency && window.has_focus()) {


BlueManCZ (commented):
Looks like it solved the bug, but did you find a reproducible test case to verify that the dance animation still works fine with urgent windows?

inthreedee:
@BlueManCZ I tested it with Signal.  When a contact sends me a message while Signal does not have focus, I still see the dancing animation and it clears away when I focus the app.  So it doesn't seem like we've broken any normal behavior.

It's worth noting that Signal's wobbling windows behave properly even without this fix.  I've never had that app get stuck in the same way Firefox does as described in #1568. The OP in that report, however, shows it happening with Telegram.  I can't explain this difference in behavior.

BlueManCZ:
I'm the OP in that report and the fun fact is I can't reproduce this dancing animation getting stuck in any version (with or without this patch) except Firefox. Tried to send emails to myself to only get Geary notification but not the animation. Chatted with bots on Telegram to get the same result. I just sometimes notice it's happening but have no idea what caused it. Simple notifications seem to not be the trigger.

BlueManCZ:
Alright, I found what is triggering it. The notification itself does not, but **clicking on that notification** does. So basically the whole process of starting the dancing animation seems to be wrong in my case. And I'm not able to get any dancing animations with this patch, which does not necessarily mean it's wrong, but I think there is another bug somewhere else.

inthreedee:
@BlueManCZ  That behavior seems consistent with the Firefox repro steps in the original issue.  Clicking a link in another app equates to clicking a notification in your example.  In both cases, the app is being brought to the foreground when you click either the link for Firefox or the notification for Telegram.  In these cases, I don't think the dancing animation should be happening because the app has been given focus at that point.

To that end, part of my fix is to prevent the urgent flag from being applied to apps that are focused in this way.  It should only be applied to apps that are not focused to get your attention.

Without that part of the fix, you would be experiencing this behavior:  Click the notification, Telegram is focused and dancing animation starts, then you have to select another app and select Telegram *again* to stop the dancing animation.  This makes no sense to me so I considered it a bug.

BlueManCZ (approved):

3v1n0:
Mh, I can't easily reproduce this, but it's not clear to me here, as if a window become urgent because of a shell signal, we should get the cancellation signal afterwards.

And it works for me:

```
gen 22 04:00:48 ubuntu-hero gnome-shell[2719]: JS ERROR: Add: urgent Window Mozilla Firefox, attention: true, urgent false, manual: undefined, focused: false: Error
                                                 _addUrgentWindow@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:322:18
                                                 _updateUrgentWindows/<@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:303:22
                                                 _updateUrgentWindows@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:301:28
                                                 _updateState@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:285:14
                                                 _updateWindows@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:275:14
                                                 _init/<@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:117:74
                                                 activateWindow@resource:///org/gnome/shell/ui/main.js:700:16
                                                 activate@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:539:30
                                                 vfunc_clicked@resource:///org/gnome/shell/ui/appDisplay.js:3234:14

 [ .... Then waiting a bit and focusing the urgent window .... ]

gen 22 04:00:48 ubuntu-hero gnome-shell[2719]: JS ERROR: Changed: urgent Window Mozilla Firefox, attention: false, urgent false, manual: undefined, focused: true: Error
                                                 onDemandsAttentionChanged@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:329:22
                                                 _addUrgentWindow/<@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:337:52
                                                 activateWindow@resource:///org/gnome/shell/ui/main.js:700:16
                                                 activate@/home/marco/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:539:30
                                                 vfunc_clicked@resource:///org/gnome/shell/ui/appDisplay.js:3234:14
```

This is printed with this debug code, could you try it to provide me some more details when the bug happens?

```diff
diff --git a/appIcons.js b/appIcons.js
index 8155e98..f4c880b 100644
--- a/appIcons.js
+++ b/appIcons.js
@@ -319,10 +319,15 @@ var DockAbstractAppIcon = GObject.registerClass({
             return;
         }
 
+        logError(new Error(),
+            `Add: urgent Window ${window.get_title()}, attention: ${window.demandsAttention}, urgent ${window.urgent}, manual: ${window._manualUrgency}, focused: ${window.has_focus()}`)
+
         this._urgentWindows.add(window);
         this.urgent = true;
 
         const onDemandsAttentionChanged = () => {
+            logError(new Error(),
+                `Changed: urgent Window ${window.get_title()}, attention: ${window.demandsAttention}, urgent ${window.urgent}, manual: ${window._manualUrgency}, focused: ${window.has_focus()}`)
             if (!window.demandsAttention && !window.urgent && !window._manualUrgency)
                 this._updateUrgentWindows();
         };
```

inthreedee:
I'm using release version 71.
Repro steps that I followed, same as in the parent issue #1568 

1. Clicked the "view it on Github" link for this PR in my email client
2. This link opens in Firefox, which gets brought to the foreground as the currently focused window
3. The Firefox icon is bouncing
4. Click another app to unfocus Firefox and click back on Firefox to re-focus it
5. Icon is still bouncing
6. Open a new Firefox window
7. Icon stops bouncing

My Logs show:
```
JS ERROR: Add: urgent Window Fix urgent flag not being cleared by inthreedee · Pull Request #1613 · micheleg/dash-to-dock · GitHub — Mozilla Firefox, attention: false, urgent false, manual: undefine>
                                        _addUrgentWindow@/usr/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:322:18
                                        _onWindowDemandsAttention@/usr/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:310:18
                                        _init/<@/usr/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js:119:18
```

The second debug message never shows up in my logs.

3v1n0:
Ok, can you please try this:

```diff
diff --git a/appIcons.js b/appIcons.js
index 8155e98..da54eec 100644
--- a/appIcons.js
+++ b/appIcons.js
@@ -298,10 +298,7 @@ var DockAbstractAppIcon = GObject.registerClass({
         this._urgentWindows.clear();
         if (interestingWindows === undefined)
             interestingWindows = this.getInterestingWindows();
-        interestingWindows.forEach(win => {
-            if (win.urgent || win.demandsAttention || win._manualUrgency)
-                this._addUrgentWindow(win);
-        });
+        interestingWindows.forEach(win => this._addUrgentWindow(win));
         this.urgent = !!this._urgentWindows.size;
     }
 
@@ -311,6 +308,9 @@ var DockAbstractAppIcon = GObject.registerClass({
     }
 
     _addUrgentWindow(window) {
+        if (!window.demandsAttention && !window.urgent && !window._manualUrgency)
+            return;
+
         if (this._urgentWindows.has(window))
             return;
 
```

inthreedee:
Edited the above steps to note that Firefox is already running in step 1

inthreedee:
> Ok, can you please try this:

That does seem to prevent the issue when I follow those same repro steps with Firefox. :tada: 

For urgent windows from, say, a chat program though, I'm not able to confirm it works as expected right now. It's late here and everyone I know who could message me to generate some notifications is asleep :P  I'll try tomorrow just to make sure normal behavior isn't broken by those changes.

inthreedee:
Limited testing with Signal seems to indicate the animation is still working as intended for chat notifications.

Thank you for looking into this, it's much appreciated!

3v1n0:
Ok I'm closing this, will handle it in another MR

PR #1627, opened 2021-12-28
Makefile compatible with documentation (README says default is `dart-sass`)

nobodyatandnothing:


3v1n0:
Please update the README instead, sassc is default in shell too and we'd prefer to be consistent with that.

3v1n0:
Oh, this is actually handled in https://github.com/micheleg/dash-to-dock/pull/1631, so let's close this.

Issue #1598, opened 2021-11-22
Dash to dock switching to a different screen when a (unused) screen is disconnected.


codingcatgirl:
I have a primary monitor and two secondary monitors. I have set up dash to dock to display on my first secondary monitor. The other secondary monitors is a projector which sometimes gets disconnected/switched over to a chromecast. Whenever that projector is connected or disconnected, dash to dock gets confused and moves to the projector or to an entirely different screen, even though i never set up dash to dock to be visible on the projector.

This leads me to believe that the screen to display the dash on is stored as an index in the list of screens, and not some kind of identifier of the screen/port, which should probably be changed.

PR #1630, opened 2022-01-06
add preferred-monitor-by-connector as an alternative to storing the index

codingcatgirl:
As described in #1598 i experienced issues when displaying the dock on a secondary monitor and unplugging or replugging another secondary monitor. This behavior was caused because the monitor to display the dock on is stored as an index in `preferred-monitor` and the index of that monitor changed. However, when studying the source code, i realized that this behavior needs to be retained for backwards compatibility purposes.

This pull request fixes the issue while retaining backwards compatibility: 

- There is a new setting called `preferred-monitor-by-connector`. It stores the monitor to display the dock on using the port it is connected to instead of its index.
- The setting `preferred-monitor-by-connector` is only used when `preferred-monitor` is set to `-2`. (a previously unused value, `-1` refers to the primary monitor).
- This means that the setting `preferred-monitor` still exists and fully works, retaining backwards compatibility with old user configurations or external apps configuring this extension such as ubuntu's gnome-control-center. If such an external app sets the `preferred-monitor` setting to the index of a monitor, this setting and monitor will be used.
- The settings UI was updated to use the new setting. If the monitor to display the dock on is changed via the extension's settings UI, `preferred-monitor-by-connector` will be used and `preferred-monitor` will be set to `-2`.

I've fully tested this pull request and would kindly ask you to merge it.

Thank you for your work!

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -178,19 +178,36 @@ var DashSlideContainer = GObject.registerClass({
 });
 
 var DockedDash = GObject.registerClass({
+    Properties: {
+        'is-main': GObject.ParamSpec.boolean(
+            'is-main', 'is-main', 'is-main',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
+            false),
+        'monitor-index': GObject.ParamSpec.uint(
+            'monitor-index', 'monitor-index', 'monitor-index',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
+            0, GLib.MAXUINT32, 0),
+    },
     Signals: {
         'showing': {},
         'hiding': {},
     }
 }, class DashToDock extends St.Bin {
+    _init(params) {
+        this._position = Utils.getPosition();
+        const positionStyleClass = ['top', 'right', 'bottom', 'left'];
 
-    _init(monitorIndex) {
+        // This is the centering actor
+        super._init({
+            ...params,
+            name: 'dashtodockContainer',
+            reactive: false,
+            style_class: positionStyleClass[this._position],
+        });
         this._rtl = (Clutter.get_default_text_direction() == Clutter.TextDirection.RTL);
 
         // Load settings
         let settings = DockManager.settings;
-        this._monitorIndex = monitorIndex;
-        this._position = Utils.getPosition();
         this._isHorizontal = ((this._position == St.Side.TOP) || (this._position == St.Side.BOTTOM));
 
         // Temporary ignore hover events linked to autohide for whatever reason
@@ -202,13 +219,13 @@ var DockedDash = GObject.registerClass({
         this._intellihideIsEnabled = null;
 
         // Create intellihide object to monitor windows overlapping
-        this._intellihide = new Intellihide.Intellihide(this._monitorIndex);
+        this._intellihide = new Intellihide.Intellihide(this.monitorIndex);
 
         // initialize dock state
         this._dockState = State.HIDDEN;
 
         // Put dock on the required monitor
-        this._monitor = Main.layoutManager.monitors[this._monitorIndex];
+        this._monitor = Main.layoutManager.monitors[this.monitorIndex];
 
         // this store size and the position where the dash is shown;
         // used by intellihide module to check window overlap.
@@ -227,22 +244,13 @@ var DockedDash = GObject.registerClass({
         this._dockDwellTimeoutId = 0
 
         // Create a new dash object
-        this.dash = new DockDash.DockDash(this._monitorIndex);
+        this.dash = new DockDash.DockDash(this.monitorIndex);
 
         if (Main.overview.isDummy || !settings.get_boolean('show-show-apps-button'))
             this.dash.hideShowAppsButton();
 
-        // Create the main actor and the containers for sliding in and out and
+        // Create the containers for sliding in and out and
         // centering, turn on track hover
-
-        let positionStyleClass = ['top', 'right', 'bottom', 'left'];
-        // This is the centering actor
-        super._init({
-            name: 'dashtodockContainer',
-            reactive: false,
-            style_class: positionStyleClass[this._position],
-        });
-
         // This is the sliding actor whose allocation is to be tracked for input regions
         this._slider = new DashSlideContainer({
             monitor_index: this._monitor.index,
@@ -341,7 +349,7 @@ var DockedDash = GObject.registerClass({
         this._slider.connect(this._isHorizontal ? 'notify::x' : 'notify::y', this._updateStaticBox.bind(this));
 
         // Load optional features that need to be activated for one dock only
-        if (this._monitorIndex == settings.get_int('preferred-monitor'))
+        if (this.isMain)
             this._enableExtraFeatures();
         // Load optional features that need to be activated once per dock
         this._optionalScrollWorkspaceSwitch();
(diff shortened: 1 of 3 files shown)


docking.js
```diff
@@ -1772,14 +1776,9 @@ var DockManager = class DashToDock_DockManager {
...
-        // In case of multi-monitor, we consider the dock on the primary monitor to be the preferred (main) one
-        // regardless of the settings
-        // The dock goes on the primary monitor also if the settings are incosistent (e.g. desired monitor not connected).
-        if (this._settings.get_boolean('multi-monitor') ||
-            this._preferredMonitorIndex < 0 || this._preferredMonitorIndex > Main.layoutManager.monitors.length - 1
-            ) {
-            this._preferredMonitorIndex = Main.layoutManager.primaryIndex;
-        } else {
+        if (this._preferredMonitorIndex == -2) {
+            this._preferredMonitorIndex = Meta.MonitorManager.get().get_monitor_for_connector(this._settings.get_string('preferred-monitor-by-connector'))
```
3v1n0, line 1780:
```suggestion
            this._preferredMonitorIndex = Meta.MonitorManager.get().get_monitor_for_connector(
                this._settings.get_string('preferred-monitor-by-connector'))
```

3v1n0: I think it's fine not to check for error here as in case it will just return `-1`, and we'll proceed as expected.

docking.js
```diff
@@ -1790,8 +1789,17 @@ var DockManager = class DashToDock_DockManager {
             this._preferredMonitorIndex = (Main.layoutManager.primaryIndex + this._preferredMonitorIndex) % Main.layoutManager.monitors.length ;
         }
 
+        // In case of multi-monitor, we consider the dock on the primary monitor to be the preferred (main) one
+        // regardless of the settings
+        // The dock goes on the primary monitor also if the settings are incosistent (e.g. desired monitor not connected).
+        if (this._settings.get_boolean('multi-monitor') ||
+            this._preferredMonitorIndex < 0 || this._preferredMonitorIndex > Main.layoutManager.monitors.length - 1
+            ) {
```
3v1n0, line 1797:
```suggestion
            this._preferredMonitorIndex < 0 ||
            this._preferredMonitorIndex > Main.layoutManager.monitors.length - 1) {
```

docking.js
```diff
@@ -1790,8 +1789,17 @@ var DockManager = class DashToDock_DockManager {
...
+            this._preferredMonitorIndex = Main.layoutManager.primaryIndex;
+        }
+
         // First we create the main Dock, to get the extra features to bind to this one
-        let dock = new DockedDash(this._preferredMonitorIndex);
+        let dock = new DockedDash(this._preferredMonitorIndex, true);
```
3v1n0, line 1802: See previous comment

prefs.js
```diff
@@ -368,7 +370,7 @@ var Settings = GObject.registerClass({
 
             this._monitors.push(monitor);
 
-            if (monitor.index === preferredMonitor)
+            if (monitor.index === preferredMonitor || (preferredMonitor == -2 && preferredMonitorByConnector == monitor.connector))
```
3v1n0, line 373:
```suggestion
            if (monitor.index === preferredMonitor ||
                (preferredMonitor == -2 && preferredMonitorByConnector == monitor.connector))
```

3v1n0 (changes requested):
Thanks for the contribution, it looks good, I've some comments though.

I'm also wondering whether instead of just supporting the "old value", it would be just preferred to mark `preferred-monitor` as deprecated, but:
 - On startup and on `changed::preferred-monitor` we call a function that checks its value and if it's a positive one we just convert it to use the safer (indeed!) `preferred-monitor-by-connector`.

Could you also update the commit message to reflect what you wrote in the PR?
Otherwise it's easy to loose track.

docking.js
```diff
@@ -184,7 +184,7 @@ var DockedDash = GObject.registerClass({
     }
 }, class DashToDock extends St.Bin {
 
-    _init(monitorIndex) {
+    _init(monitorIndex, isMain) {
```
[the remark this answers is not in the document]
codingcatgirl: sounds good, i just wasn't aware of this option since i'm in no way familiar with gobject internals, this is the first time i'm involved with it :) 

for some reason this change isn't listed as a "suggested change" so i can't commit it with one click. can you maybe change that?

codingcatgirl:
> Thanks for the contribution, it looks good, I've some comments though.
Thanks :)

> I'm also wondering whether instead of just supporting the "old value", it would be just preferred to mark `preferred-monitor` as deprecated, but:
> 
> * On startup and on `changed::preferred-monitor` we call a function that checks its value and if it's a positive one we just convert it to use the safer (indeed!) `preferred-monitor-by-connector`.
I don't use ubuntu or any of the software that we want to retain backwards compatibility with, so i don't know what the requirements are, so i just implemented it the way i did. 

Your solution sounds good, i think it brings one upside and a downside:

* Upside: When an ubuntu user selects a preferred monitor, it is also retained by the extension when un- or replugging a monitor.
* Downside: The native ubuntu settings dialog will show the wrong monitor afterwards, not the one that this extension is (rightfully) using.

I think the upsides outweigh the downsides here, so i'm on board with this idea. Of course one could also set the original preferred-monitor setting whenever monitors are un- or replugged. Either way, i don't think i hae the time and motivaton to implement either of these things.

> Could you also update the commit message to reflect what you wrote in the PR? Otherwise it's easy to loose track.

Gladly, i will just squash all the commits on my branch once i've implemented the suggested changes and add a better commit message. (Can you try to put them as "suggested changes" as mentioned in https://github.com/micheleg/dash-to-dock/pull/1630#discussion_r790132359?)

docking.js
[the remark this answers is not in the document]
3v1n0: Sadly I can't as I can't suggest a such big change... 

Just copy and paste it to a new file and apply it with `git apply file.patch` or `patch -p1 < file.patch`

3v1n0:
> Downside: The native ubuntu settings dialog will show the wrong monitor afterwards, not the one that this extension is (rightfully) using.

Yeah, but I will handle the fix upstream so that ubuntu's gnome control center will use the new setting, don't worry for that.
Actually, that will push me in fixing it to use new setting :)

docking.js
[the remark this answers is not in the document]
codingcatgirl: implemented and squashed (patch didn't work so i manually edited it, please check if it's okay)

codingcatgirl:
As far as i can tell i just implemented all the suggestions except the one about automatically translating preferred-monitor to preferred-monitor-by-connector automatically. If i see correctly, this would require migrate code over from the prefs.js so that the connector for a specific monitor index could be obtained. Since this would involve some significant refactoring and moving code from one file to another (or creating duplicate code), this is not a change i'm comfortable doing myself.

codingcatgirl:
I've also rebased and squashed the relevant commits and added the description to the commit message

codingcatgirl:
I have *not* tested this latest state so far.

3v1n0 (approved):

3v1n0:
Thanks, I've tested it, adding few more cleanups and it's now ready to merge!

3v1n0:
FYI this is now supported by ubuntu natively: https://salsa.debian.org/gnome-team/gnome-control-center/-/commit/07cc5147555e0cc1a5b500b7efed53e12a8f5d08

codingcatgirl:
That's so cool! Thank for mentioning it!

PR #1631, opened 2022-01-07
change README to document actual behavier

nobodyatandnothing:
this would be instead of https://github.com/micheleg/dash-to-dock/pull/1627


3v1n0 (approved):
Thanks

Issue #1629, opened 2022-01-04
Gnome 41.2: wrong windows counter indicators count


ask6483:
As seen on screenshot attached bellow windows counter indicators numbers are totally wrong:
- 3 opened nautilus (files) windows -> 4 indicators
- 1 STM32CubeIDE -> 4 indicators
- 1 Thunderbird window -> 2 indicators

Running gnome-shell over X server and reloading session doesn't help.

OS: Debian GNU/Linux bookworm/sid (testing)
Gnome 41.2

![image](https://user-images.githubusercontent.com/72493041/148085343-d00ee49a-0214-4cc4-8a64-1e2ac2c7fd11.png)

mbiebl:
I can reproduce this as well. Disabling the extension and running the GNOME Shell builtin overview, I see the correct number for the windows counters.

inthreedee:
I've noticed it has started picking up tooltips in the window count, which may be related to this.  Perhaps things like tooltips are confusing the count and causing it to get stuck at a wrong count

To reproduce, I'll use Firefox as an example
1. Open a site in a tab
2. Hover the mouse over the tab until the tooltip appears, showing the full title of the website
3. Dash to dock counts an extra window for Firefox
4. Move the mouse away from the tab and the count returns to what it was previously

GeoffreyCoulaud:
I confirm that the issue is reproducible, I have the same problem, it's quite annoying.

System : Manjaro (unstable branch)
Gnome version : 41.3

kanakshilledar:
I can also confirm the issue of wrong windows counter.
System: Endeavour OS Atlantis
Gnome version: 41.3

When browsing on firefox the window counter suddenly increases even if i have only one window present. I was able to get this issue in Nautilus when opening two different nautilus windows and copying one file from the other the window counter increments by 1.

GeoffreyCoulaud:
If you want to speed up the process, test MR #1637 and give feedback on it.

jesusignazio:
Same in GNOME Shell 41.3 on Archlinux.

3v1n0:
Can you provide some more infos of the windows causing the issue by using this in looking glass (from `alt+f2` → `lg`):

`Main.extensionManager._extensions.get('dash-to-dock@micxgx.gmail.com').imports.locations.getFileManagerApp().get_windows().map(w => [w.title, w.skip_taskbar,'|'])`

Or more generically, if you launch

[...]

PR #1637, opened 2022-01-13
Fix window counter (exclude hidden windows)

GeoffreyCoulaud:
Hide windows that hint not to be shown in the task bar.
Fix for #1629

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -1232,7 +1232,7 @@ function getInterestingWindows(windows, monitorIndex) {
         const showUrgent = settings.get_boolean('workspace-agnostic-urgent-windows');
         const activeWorkspace = global.workspace_manager.get_active_workspace_index();
         windows = windows.filter(function(w) {
-            const inWorkspace = w.get_workspace().index() == activeWorkspace;
+            const inWorkspace = w.get_workspace().index() === activeWorkspace;
             const isUrgent = w.urgent || w.demandsAttention || w._manualUrgency;
             return inWorkspace || (showUrgent && isUrgent);
         });
@@ -1240,7 +1240,7 @@ function getInterestingWindows(windows, monitorIndex) {
 
     if (settings.get_boolean('isolate-monitors')){
         windows = windows.filter(function(w) {
-            return w.get_monitor() == monitorIndex;
+            return w.get_monitor() === monitorIndex;
         });
     }
 
--- a/appIcons.js
+++ b/appIcons.js
@@ -1233,8 +1233,11 @@ function getInterestingWindows(windows, monitorIndex) {
         const showUrgent = settings.get_boolean('workspace-agnostic-urgent-windows');
         const activeWorkspace = global.workspace_manager.get_active_workspace_index();
         windows = windows.filter(function(w) {
-            return w.get_workspace().index() == activeWorkspace
-                || (showUrgent && (w.urgent || w.demandsAttention || w._manualUrgency));
+            // See GJS docs : Meta/Window
+            const inWorkspace = w.get_workspace().index() == activeWorkspace;
+            const isUrgent = w.urgent || w.demandsAttention || w._manualUrgency;
+            const isNotInTaskbar = w.is_skip_taskbar();
+            return !isNotInTaskbar && (inWorkspace || (showUrgent && isUrgent));
         });
     }
 
--- a/appIcons.js
+++ b/appIcons.js
@@ -1002,8 +1002,7 @@ const DockAppIconMenu = class DockAppIconMenu extends PopupMenu.PopupMenu {
             this._allWindowsMenuItem.hide();
             this.addMenuItem(this._allWindowsMenuItem);
         } else {
-            const windows = this._source.getInterestingWindows().filter(
-                w => !w.skip_taskbar);
+            const windows = this._source.getInterestingWindows();
 
             if (windows.length > 0) {
                 this.addMenuItem(
@@ -1233,20 +1232,19 @@ function getInterestingWindows(windows, monitorIndex) {
         const showUrgent = settings.get_boolean('workspace-agnostic-urgent-windows');
         const activeWorkspace = global.workspace_manager.get_active_workspace_index();
         windows = windows.filter(function(w) {
-            // See GJS docs : Meta/Window
             const inWorkspace = w.get_workspace().index() == activeWorkspace;
             const isUrgent = w.urgent || w.demandsAttention || w._manualUrgency;
-            const isNotInTaskbar = w.is_skip_taskbar();
-            return !isNotInTaskbar && (inWorkspace || (showUrgent && isUrgent));
+            return inWorkspace || (showUrgent && isUrgent);
         });
     }
 
-    if (settings.get_boolean('isolate-monitors'))
+    if (settings.get_boolean('isolate-monitors')){
         windows = windows.filter(function(w) {
             return w.get_monitor() == monitorIndex;
         });
+    }
 
-    return windows;
+    return windows.filter(w => !w.is_skip_taskbar());
 }
 
 /**
--- a/appIcons.js
+++ b/appIcons.js
@@ -1238,13 +1238,13 @@ function getInterestingWindows(windows, monitorIndex) {
         });
     }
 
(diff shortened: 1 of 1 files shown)


slavicd (approved):
Ran in Fedora 35

kanakshilledar:
nautilus window counter
Sorry, not working for me. I tried opening two nautilus windows and then tried copying one file to the other then after the transfer process ended the counter incremented by 2. I have attached a screenshot for reference.
tried logging out.

GeoffreyCoulaud:
@kanakshilledar I could not reproduce, following your procedure I have only 2 nautilus indicators before, during and after the copy.  

Did you test this MR properly ?   
You need to disable the extension, build and install my branch, reload `gnome-shell` (`alt`+`f2`, `r`, `return`) and then try again. 

If it's still broken on your machine, I'll need to investigate deeper.

slavicd:
I only needed the restart gnome shell part for the changes to take effect. I'll restate that the fix worked for me (I am, in fact, still running the code until the upstream merges this).

kanakshilledar:
@GeoffreyCoulaud  I did the following
- Removed the old extension from gnome-extension using the `remove` option
- Cloned your branch with `$ git clone https://github.com/GeoffreyCoulaud/dash-to-dock`
- Installed by `$ make` and `$ make install`
- reloaded the shell using `alt` + `f2`, `r`)
But still I am having that issue.
I am having Gnome shell version 41.3 and Endeavour OS (arch linux)
I have also attached a sample video [video_link](https://youtu.be/BVQG-aAdhW0)

kanakshilledar:
@GeoffreyCoulaud  I am using X11 and getting this issue.
```
$ loginctl show-session 7 -p Type
Type=x11 
```

GeoffreyCoulaud:
Yup, I tested again and It's broken in both now. I'm reading the mutter docs to see what's going wrong.

GeoffreyCoulaud:
@kanakshilledar You may want to report that to the EndeavourOS maintainers.  
I've checked with my Manjaro install, mutter is the same version so it's likely not that.  
My fix works on Arch, Manjaro and Fedora.

I've also tried filtering with the `get_window_type()` function and only keeping `NORMAL`, `DIALOG` and `MODAL_DIALOG` ([docs here](https://gjs-docs.gnome.org/meta9~9_api/meta.windowtype)) and even that is not working. Maybe it's somewhere earlier in the call stack that breaks ? I have no idea.

kanakshilledar:
Oh okay I will try reporting them... Thanks for the suggestion

kanakshilledar:
I tried Reborn OS and in the live image I added the extension and it was working fine only once I felt that issue. So maybe it's with Endeavour OS only

appIcons.js
```diff
@@ -1233,8 +1233,11 @@ function getInterestingWindows(windows, monitorIndex) {
         const showUrgent = settings.get_boolean('workspace-agnostic-urgent-windows');
         const activeWorkspace = global.workspace_manager.get_active_workspace_index();
         windows = windows.filter(function(w) {
-            return w.get_workspace().index() == activeWorkspace
-                || (showUrgent && (w.urgent || w.demandsAttention || w._manualUrgency));
+            // See GJS docs : Meta/Window
```
3v1n0, line 1236: This is not needed, it should be pretty known for who is into developing a shell extension :)

appIcons.js
```diff
@@ -1233,8 +1233,11 @@ function getInterestingWindows(windows, monitorIndex) {
...
+            const inWorkspace = w.get_workspace().index() == activeWorkspace;
+            const isUrgent = w.urgent || w.demandsAttention || w._manualUrgency;
+            const isNotInTaskbar = w.is_skip_taskbar();
```
3v1n0, line 1239: I was thinking to include this for some time now too... I've been always wondering if this would break the case of showing the windows into the previews-menu though.

In fact while the windows may not be to be shown in the lists, *maybe* we may want to show them in the previews overview... But I'm not either sure about :)

3v1n0 (changes requested):
idea looks ok though, but I think it must be at upper level, so outside the `settings.get_boolean('isolate-workspaces')` check scope.

Ideally you can just `return windows.filter(w => !w.skipTaskbar);`

3v1n0:
Ah, and please remove the duplicated filter on `_rebuildMenu`.

GeoffreyCoulaud:
Thanks for the review.  
I implemented the changes, I also added curly braces to the `isolate_monitors` check to be more consistent.  

About the ability to separate visibility in the indicator and in the preview, I'm not sure about that. I think having a different number of indicators and previews in unexpected for the user. Having options is great, still I believe the added complexity and the difference from the overview dock is not worth it.

appIcons.js
```diff
@@ -1233,17 +1232,19 @@ function getInterestingWindows(windows, monitorIndex) {
         const showUrgent = settings.get_boolean('workspace-agnostic-urgent-windows');
         const activeWorkspace = global.workspace_manager.get_active_workspace_index();
         windows = windows.filter(function(w) {
-            return w.get_workspace().index() == activeWorkspace
-                || (showUrgent && (w.urgent || w.demandsAttention || w._manualUrgency));
+            const inWorkspace = w.get_workspace().index() == activeWorkspace;
```
3v1n0, line 1235: Since you're there, can you please change to `===` equality check?

appIcons.js
```diff
@@ -1233,17 +1232,19 @@ function getInterestingWindows(windows, monitorIndex) {
...
+            const isUrgent = w.urgent || w.demandsAttention || w._manualUrgency;
+            return inWorkspace || (showUrgent && isUrgent);
         });
     }
 
-    if (settings.get_boolean('isolate-monitors'))
+    if (settings.get_boolean('isolate-monitors')){
```
3v1n0, line 1241:
```suggestion
    if (settings.get_boolean('isolate-monitors')) {
```

appIcons.js
GeoffreyCoulaud: Will do !

appIcons.js
```diff
@@ -1233,17 +1232,19 @@ function getInterestingWindows(windows, monitorIndex) {
...
         windows = windows.filter(function(w) {
             return w.get_monitor() == monitorIndex;
         });
+    }
 
-    return windows;
+    return windows.filter(w => !w.is_skip_taskbar());
```
3v1n0, line 1247: I like using the property better here:

```suggestion
    return windows.filter(w => !w.skipTaskbar);
```

3v1n0 (approved):

FreddieChopin:
Sorry for noob question, but is the thing you discuss above the same as the issue I'm seeing right now:
Screenshot from 2022-01-31 15-20-45
I have just one Firefox window opened (I had two, but I closed it, yet there are 2 dots), only 1 window from e-mail client (sent the e-mail I was writing, so these additional windows are closed too), 1 windows of Nautilus.

kanakshilledar:
I had the issue but it got fixed by the recent update.

sergiotca:
> Sorry for noob question, but is the thing you discuss above the same as the issue I'm seeing right now: Screenshot from 2022-01-31 15-20-45 I have just one Firefox window opened (I had two, but I closed it, yet there are 2 dots), only 1 window from e-mail client (sent the e-mail I was writing, so these additional windows are closed too), 1 windows of Nautilus.

Yep, this is the issue at hand here.

sergiotca:
@FreddieChopin Yep, that is the issue at hand here.

sergiotca:
> I had the issue but it got fixed by the recent update.

What update are you talking about buddy?

GeoffreyCoulaud:
@sergiotca Probably their distribution's package managers updated the extension.

To anyone unsure if the report applies to their bug : If it's about more indicators than should be, it's this issue and if you compile the extension from source you'll see it being resolved.

tareqimbasher:
Thanks guys. When can we expect this fix to be pushed to extensions.gnome.org?

kanakshilledar:
@sergiotca I manually compiled the Dash to Dock extension and the issue got resolved as suggested by @GeoffreyCoulaud.

Issue #1647, opened 2022-01-21
spanish translation duplicate message definition


doppelhelix:
```
po/es.po:519: duplicate message definition...
po/es.po:102: ...this is the location of the first definition
msgfmt: found 1 fatal error
```

aminvakil:
This is causing my AUR package to fail: https://aur.archlinux.org/packages/gnome-shell-extension-dash-to-dock-git/

https://github.com/aminvakil/aur/runs/4904194617?check_suite_focus=true#step:11:961

PR #1648, opened 2022-01-21
po/es.po: Remove duplicated entry for Quit translation

3v1n0:
And use "Cerrar" to be consistent with other cases.

Fixes: #1647


rastersoft:
This MR is critical. Without it, both `make` and `make install` fail.

rastersoft (approved):
Looks good.

3v1n0:
This was included in #1657 to make CI pass.

PR #1656, opened 2022-01-24
Locations refactory to use native Gio.AppIcon's (for GNOME 42 support)

3v1n0:
Refactor locations handling to have a more native Shell integration.

A part the various improvements and features, the main trigger for this was (as per the main commit message):

```
As per gjs 1.72 it won't be possible to inject an interface vfunc as it
used to work in previous versions, this would make gnome-shell to crash
as we may try to get the location "apps" ID which is now null (and not
customId anymore).

So, let's re-implement the things in a cleaner way by actually using
real custom AppInfo's that are still based on Gio.DesktopAppInfo
(beause the shell will still need to be such type), but that are
implementing the needed virtual functions of the parent interface.

As per this, we can finally use native shell mount operations to handle
mount options such as mounting encrypted disks.
```

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -1208,11 +1208,24 @@
                                 <property name="can_focus">0</property>
                                 <property name="hexpand">1</property>
                                 <property name="halign">start</property>
-                                <property name="label" translatable="yes">Show mounted volumes and devices</property>
+                                <property name="label" translatable="yes">Show volumes and devices</property>
 
                                 <layout>
                                   <property name="column">0</property>
                                   <property name="row">0</property>
+                                </layout>
+                              </object>
+                            </child>
+                            <child>
+                              <object class="GtkCheckButton" id="show_only_mounted_devices_check">
+                                <property name="label" translatable="yes">Only if mounted</property>
+                                <property name="halign">start</property>
+                                <property name="margin_top">12</property>
+
+                                <layout>
+                                  <property name="column">0</property>
+                                  <property name="row">1</property>
+                                  <property name="column-span">1</property>
                                 </layout>
                               </object>
                             </child>
--- a/Settings.ui
+++ b/Settings.ui
@@ -921,7 +921,7 @@
                                 <property name="can_focus">0</property>
                                 <property name="hexpand">1</property>
                                 <property name="halign">start</property>
-                                <property name="label" translatable="yes">Show favorite applications</property>
+                                <property name="label" translatable="yes">Show pinned applications</property>
 
                                 <layout>
                                   <property name="column">0</property>
--- a/Settings.ui
+++ b/Settings.ui
@@ -1229,6 +1229,19 @@
                                 </layout>
                               </object>
                             </child>
+                            <child>
+                              <object class="GtkCheckButton" id="show_network_volumes_check">
+                                <property name="label" translatable="yes">Include network volumes</property>
+                                <property name="halign">start</property>
+                                <property name="margin_top">12</property>
+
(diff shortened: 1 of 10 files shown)


timocapa:
not sure if this is the right place for this, but some issues ive ran into so far:

dock doesn't hide properly and app names when hovered over don't disappear most of the time, making it actually kinda annoying to use

sorry if this is unrelated here (since im not sure what this commit does)

stuarthayhurst:
> not sure if this is the right place for this, but some issues ive ran into so far:
> 
> dock doesn't hide properly and app names when hovered over don't disappear most of the time, making it actually kinda annoying to use
> 
> sorry if this is unrelated here (since im not sure what this commit does)

If you're referring to this branch specifically, I'm getting the same issues, using the standard release from EGO (Version 71), so I don't think it's related to this

timocapa:
> > not sure if this is the right place for this, but some issues ive ran into so far:
> > dock doesn't hide properly and app names when hovered over don't disappear most of the time, making it actually kinda annoying to use
> > sorry if this is unrelated here (since im not sure what this commit does)
> 
> If you're referring to this branch specifically, I'm getting the same issues, using the standard release from EGO (Version 71), so I don't think it's related to this

I'm quite confident this was introduced in a Gnome change a while ago, thought I'd mention it anyway

Etaash-mathamsetty:
could this get merged?

schtobia:
I can confirm this fix for GNOME 42.0-2 [as present in Debian sid](https://packages.debian.org/sid/gnome-shell).

fancypantalons:
Confirmed here as well.  Thank you @3v1n0 !

rastersoft:
There is a bug in this patch: in the file docking.js, the property `this._monitorIndex` is replaced with `this.monitorIndex`, but there are several references kept to `this._monitorIndex`, specially in the Desktop Icons integration (which, BTW makes the integration to fail).

phush0:
There is a bug in animation minimizing window. It will draw two animations one go to the left upper edge of screen and one minimize to dock. It is most visible with minimize Nautilus window

martindmtrv:
Running this branch on my fresh Gnome 42 install all seems swell! Only issue I had was with the trash icon it had a white background. When disabling the "show trash icon" then re-enabling it seems to have fixed that issue

OctoRocket:
Is this pull ready to go? Can't wait to use it on my computer!

3v1n0:
@rastersoft I notice the issue is also upstream: could you please do a quick MR to fix them?
https://github.com/micheleg/dash-to-dock/blob/ubuntu-dock/docking.js#L1109-L1122

as if I do it myself I can't merge it afterwards :)

MacTavishAO:
> not sure if this is the right place for this, but some issues ive ran into so far:
> 
> dock doesn't hide properly and app names when hovered over don't disappear most of the time, making it actually kinda annoying to use
> 
> sorry if this is unrelated here (since im not sure what this commit does)

Building mutter from source fixes this issue.
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2321

micheleg (approved):

rastersoft:
It seems that in the file `docking.js` there are still three references to this._monitorIndex.

3v1n0:
@rastersoft I don't see your comment here, however the mentions of `this._monitorIndex` present are present only on functions/vfuncs overrides.

So they refer to the overridden type.

Zorono (approved):

ashb (approved):

Issue #1654, opened 2022-01-23
Simple CI trying to build package


aminvakil:
What do you think about a CI which just tests if `make` and `make install` works correctly in order to prevent something like https://github.com/micheleg/dash-to-dock/issues/1647 happen.

It would be nice and I can help with that if that's acceptable.

3v1n0:
Yeah, I was thinking about that...

Indeed it would be welcome! Also doing some eslint would be nice to have a basic JS syntax validator, sadly we're not at the point of having full eslint, but wouldn't be bad to use it (probably can be done by [disabling all the rules](https://eslint.org/docs/user-guide/configuring/) though).

PR #1657, opened 2022-01-24
Simple make CI

aminvakil:
Fixes https://github.com/micheleg/dash-to-dock/issues/1654

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -21,5 +21,11 @@ jobs:
       - name: Make
         run: make
 
+      - name: Install modules
+        run: sudo npm install eslint -g
+
+      - name: Make Check
+        run: make check
+
       - name: Make Install
         run: make install
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -1,0 +1,25 @@
+---
+name: Make
+'on':
+  pull_request:
+  push:
+    branches:
+      - master
+      - ubuntu-dock
+
+jobs:
+  make:
+    name: Make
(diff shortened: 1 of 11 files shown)


aminvakil:
Currently run actions can be viewed from https://github.com/aminvakil/dash-to-dock/actions.

Merging #1648 fixes the currently broken make obviously and can be verified [here](https://github.com/aminvakil/dash-to-dock/actions/runs/1738576432) too.

.github/workflows/make.yml
```diff
@@ -0,0 +1,27 @@
+---
+name: Make
+'on':
+  pull_request:
+  push:
+    branches:
+      - master
+  schedule:
+    - cron: '0 0 * * *'
```
3v1n0, line 9: There's no need for having a cron check here... I think it's fine on pushes and cron.

.github/workflows/make.yml
```diff
@@ -0,0 +1,27 @@
+---
+name: Make
+'on':
+  pull_request:
+  push:
+    branches:
+      - master
```
3v1n0, line 7:
```suggestion
      - master
      - ubuntu-dock
```

3v1n0 (changes requested):

.github/workflows/make.yml
aminvakil: Fixed in https://github.com/micheleg/dash-to-dock/pull/1657/commits/f219115cde96a1743d73d3aa8da7378afe7f2216

3v1n0:
Thanks, i've also added some eslint checks

Issue #1620, opened 2021-12-17
Improve nonlinear animation curve, refer to Plank


conan81412B:
In another popular dock [Plank](https://github.com/ricotz/plank), many different nonlinear curves are used, such as EASE_IN_CUBIC, EASE_IN_EXPO, EASE_IN_QUINT, etc.

https://github.com/ricotz/plank/blob/43fe952ae5eede0a537b48e037daacf6c23a703a/lib/DockRenderer.vala#L259
https://github.com/ricotz/plank/blob/43fe952ae5eede0a537b48e037daacf6c23a703a/lib/DockRenderer.vala#L633
https://github.com/ricotz/plank/blob/43fe952ae5eede0a537b48e037daacf6c23a703a/lib/DockRenderer.vala#L652

However, there are only ordinary QUAD animation curves in dash-to-dock.

https://github.com/micheleg/dash-to-dock/blob/39e5597da42124a23d3f0d5dbca69fea1f3fa96b/docking.js#L734
https://github.com/micheleg/dash-to-dock/blob/39e5597da42124a23d3f0d5dbca69fea1f3fa96b/windowPreview.js#L531
https://github.com/micheleg/dash-to-dock/blob/39e5597da42124a23d3f0d5dbca69fea1f3fa96b/dash.js#L707

If we adjust these curves, maybe the look and feel of dash-to-dock will become better :)

conan81412B:
We can preview different animation curves [here](https://api.flutter.dev/flutter/animation/Curves-class.html), thanks to Flutter.

3v1n0:
Such curves were mostly selected to be consistent with what upstream code does, but feel free to propose other animation curves in specific places through a PR and we can see if we get an improvement :-)

PR #1662, opened 2022-01-28
Improve nonlinear animation curve

conan81412B:
These animation improvements will make the dash-to-dock feel more responsive than before.
https://github.com/micheleg/dash-to-dock/issues/1620

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -735,7 +735,7 @@ var DockedDash = GObject.registerClass({
       this._slider.ease_property('slide-x', 1, {
             duration: time * 1000,
             delay: delay * 1000,
-            mode: Clutter.AnimationMode.EASE_OUT_QUAD,
+            mode: Clutter.AnimationMode.EASE_IN_OUT_CUBIC,
             onComplete: () => {
                 this._dockState = State.SHOWN;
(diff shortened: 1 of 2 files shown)


emansom:
How will it make things more responsive?

conan81412B:
> How will it make things more responsive?

animation_spacing

I found a GIF from [here](https://5pmcasual.com/post/everything_is_animation/). It can be seen that while the duration of the first two animations is the same, people feel that the second animation is more responsive (or faster). That's the difference between linear animation and non-linear animation - non-linear animation looks more "high-end".

In addition, it can also be noticed that more and more non-linear animations have appeared in recent GNOME versions (such as window maximization/minimization animations, activation animations of activity overview), it can be considered that GNOME has a tendency to replace linear animations with non-linear animations.

You can experience the different feel of different nonlinear animations [here](https://api.flutter.dev/flutter/animation/Curves-class.html).

nikodunk:
So this PR basically changes the animation from [ease-out](https://api.flutter.dev/flutter/animation/Curves/easeOutQuad-constant.html) to [ease-in-out](https://api.flutter.dev/flutter/animation/Curves/easeInOutCubic-constant.html), which I believe the bar will feel "stickier".

PR #1666, opened 2022-02-04
Add 42 since it works there

shemgp:
Tested using Ubuntu Jammy with GNOME PPA.

Prefs works too.


3v1n0:
Sorry, this is not enough as it could lead to crashes, reason why of https://github.com/micheleg/dash-to-dock/pull/1656

PR #1677, opened 2022-02-27
slow reloading dock when open/close window

leanh128:


jaredfeng-ca:
Hi I have encountered the same problem after I set up the gnome extension stuff. Now I have uninstalled all those extension setups but the dock is still slow to respond

leanh128:
For me, the gnome-shell's crash log  literally showed that the `firstButton` was `undedfined ` after turning off  `show-mounted`.

franciskafieh:
I tried locally merging your pr into upstream ubuntu-dock, but my dock is still not updating icons....

here's my log when I grep for undefined:

image

would appreciate your help! :)) tysm

jaredfeng-ca:
I fixed this error following instructions from https://askubuntu.com/questions/1254414/ubuntu-20-04-not-updating-icons-on-dock.

franciskafieh:
> I fixed this error following instructions from https://askubuntu.com/questions/1254414/ubuntu-20-04-not-updating-icons-on-dock.

hmm, it seems like that was old because the current source already has this:
```js
// if no icons there's nothing to adjust
if (!firstIcon)
return;
```

this guy seems to agree..
image

(not on ubuntu by the way, so i'm just cloning the latest ubuntu-dock branch and manually patching and rebuilding)

franciskafieh:
uhh... not sure what i did differently... but it's fixed for me!

step 1: `git clone https://github.com/micheleg/dash-to-dock.git -b ubuntu-dock`
step 2: rename and cd into this new dir
step 3: run `make`
step 4: reload shell and enable in extensions app

i guess this pull request can be closed, as it's not necessary?

jaredfeng-ca:
> > I fixed this error following instructions from https://askubuntu.com/questions/1254414/ubuntu-20-04-not-updating-icons-on-dock.
> 
> hmm, it seems like that was old because the current source already has this:
> 
> ```js
> // if no icons there's nothing to adjust
> if (!firstIcon)
> return;
> ```
> 
> this guy seems to agree.. image
> 
> (not on ubuntu by the way, so i'm just cloning the latest ubuntu-dock branch and manually patching and rebuilding)

hmm perhaps it's because my os version is 20.04 LTS, which might not be the latest?

3v1n0:
I've handled this in https://github.com/3v1n0/dash-to-dock/commit/de1299bf918d30f1079dbaffac31c42b553ed06d (also part of ubuntu-dock branch).

So closing this, thanks @leanh128 for catching it!

PR #1678, opened 2022-03-02
locations: Reintroduce manual rate limiting for trash monitoring

vanvugt:
Because `g_file_monitor_set_rate_limit` doesn't work. It's a stub in
glib!

Fixes: https://launchpad.net/bugs/1962699

### DIFF
--- a/locations.js
+++ b/locations.js
@@ -586,7 +586,7 @@ class TrashAppInfo extends LocationAppInfo {
 
         try {
             this._monitor = this.location.monitor_directory(0, this.cancellable);
-            this._monitor.set_rate_limit(UPDATE_TRASH_DELAY);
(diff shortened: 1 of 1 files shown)


vanvugt:
CC @3v1n0

locations.js
```diff
@@ -621,10 +621,18 @@ class TrashAppInfo extends LocationAppInfo {
     }
 
     _onTrashChange() {
+        if (this._schedUpdateId)
+            GLib.source_remove(this._schedUpdateId);
```
3v1n0, line 625: Please unset this also on `destroy()`.

We're already cancelling the cancellable there, so indirectly we should avoid to do thing anyways, but still better to trigger the callback.

vanvugt: Done.

Issue #1697, opened 2022-04-04
Add to Favorites/Remove from Favorites should be translatable


m-shibata:
The current code expects "Add to Favorites" and "Remove from Favorites" will be translated by GNOME Shell.

https://github.com/micheleg/dash-to-dock/blob/f928de04659210eb9876cd48811ce3c65af4e969/appIcons.js#L851-L857

However these labels are renamed to "Pin to Dash" and "Unpin" on GNOME Shell.

https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2136/diffs

Could you replace these labels to GNOME Shell domain like or set translatable by dashtodock domain?

gunnarhj:
Those are two of the most visible strings on the desktop, so I think this should be fixed instantly.

Considering that Ubuntu calls the thing "Dock" and not "Dash", moving the existing strings to the `dashtodock` domain makes most sense IMO.

gunnarhj:
Related gnome-shell merge request in salsa:

https://salsa.debian.org/gnome-team/gnome-shell/-/merge_requests/61

PR #1710, opened 2022-04-11
Move "Add to Favorites" and "Remove from Favorites" to dashtodock

gunnarhj:
Fixes: https://github.com/micheleg/dash-to-dock/issues/1697


gunnarhj:
Which LP bug? The issue was reported on the ubuntu-translators mailing list.

PR #1720, opened 2022-04-21
Fix stylesheet for gnome 42

daniruiz:
With the update to gnome 42 the new updated branch ubuntu-dock works great but the stylesheet didn't get completely updated. Some of the spacings and margings look odd. This patch fixes it 

### Old:
Captura desde 2022-04-21 10-40-23
### New:
image

### DIFF
--- a/_stylesheet.scss
+++ b/_stylesheet.scss
@@ -1,26 +1,25 @@
 // From https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/c17dc9c8ecba0b542aab75f13da238f7b0690031/data/theme/gnome-shell-sass/_common.scss#L28
 $base_padding: 6px;
 $base_margin: 4px;
-$base_spacing: 6px;
 
+// border radii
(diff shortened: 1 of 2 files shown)


daniruiz:
It's also the patch I created for the dash-to-dock extension for Kali Linux. Here is the source
https://gitlab.com/kalilinux/packages/gnome-shell-extension-dashtodock/-/blob/kali/master/debian/patches/fix-stylesheet-for-gnome-42.patch

Zorono (approved):

proninyaroslav:
@daniruiz 
What could be the reason? I removed the old extension (for GNOME 41) and installed this.

proninyaroslav:
This seems to be a problem with f36 itself, I see the same problem in the desktop extension.

3v1n0:
Some of this stuff was included in #1826 and #1827, so I'm closing this.

Please, make a new one in case there's something we missed.

PR #1724, opened 2022-04-24
Add support for GNOME 42 of dash to dock

vikdevelop:


RyzenDew:
This is wrong so wrong and the pr like this was already closed … also a pr is already open use that thanks

RyzenDew:
https://github.com/micheleg/dash-to-dock/pull/1656

PR #1732, opened 2022-05-06
Pop os changes

13r0ck:
Because of https://github.com/pop-os/cosmic-dock/issues/139, thank you for the reminder!


micheleg (approved):
All seems sensible to me. Apart from the binary indicators :) but they are non invasive anyway.

3v1n0:
Thanks a lot for addressing this, I've rebased this dropping the hiding fix as a similar fix was merged already via #1751.

PR #1740, opened 2022-05-13
Quick fix to allow use on Gnome 42

sterndata:
This just adds 

`, "42"`

to the extension's metadata so it will load on Gnome 42 (Fedora 36)

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,7 +1,8 @@
 {
 "shell-version": [
     "40",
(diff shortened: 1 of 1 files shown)


emansom:
Duplicate, irrelevant etc. See #1656 and next time check the git history across branches before contributing to anything. `ubuntu-dock` has latest patches.

PR #1752, opened 2022-05-19
Add auto show preview to dock

postnick:
See this request, implament the same request so this will work for those of use not on POP. 

[Cosmic-Dock/pull/142](https://github.com/pop-os/cosmic-dock/pull/142/)


3v1n0:
The idea is fine, but please only cherry-pick the relevant commits from that PR.

postnick:
I would, but frankly I don't know where the changes are made for my request.

3v1n0:
I think the change you wanted already went in via #1732

PR #1759, opened 2022-05-28
Draft: Update show apps icon for GNOME 42 changes

ewlsh:
Since https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/49b9ae08d81345e17b7b7794709a374d5d57e058, our show apps icon implementation has been triggering errors due to a missing `forcedHighlight` implementation, this is a refactor to better align show apps with upstream code.

### DIFF
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -32,7 +32,7 @@ const MAX_WINDOWS_CLASSES = 4;
 
 
 /*
- * This is the main indicator class to be used. The desired bahviour is
+ * This is the main indicator class to be used. The desired behavior is
  * obtained by composing the desired classes below based on the settings.
  *
  */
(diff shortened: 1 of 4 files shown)


appIconIndicators.js
```diff
@@ -169,10 +169,10 @@ var RunningIndicatorBase = class DashToDock_RunningIndicatorBase extends Indicat
     }
 
     _updateDefaultDot() {
-        if (this._source.running)
+        // if (this._source.running)
             this._source._dot.show();
-        else
-            this._source._dot.hide();
+        // else
+        // this._source._dot.hide();
```
psgreco, line 175: This block makes the dot always visible, seems like debug leftover

appIcons.js
```diff
@@ -1225,77 +1225,65 @@ function getInterestingWindows(windows, monitorIndex) {
...
+
+        this._iconActor = null;
+        
+        // Create the icon
+        this.icon = new IconGrid.BaseIcon(_('Show Applications'), {
+            setSizeManually: true,
+            showLabel: false,
+            createIcon: this._createIcon.bind(this),
+        });
+        this.icon.y_align = Clutter.ActorAlign.CENTER;
```
cuviper, line 1252: I have my dock in **left** panel mode, and I believe this `CENTER` change is the reason my icon is no longer at the end (bottom). Could that orientation be detected here?

(I'm here because Fedora applied your patch to its package.)

PR #1767, opened 2022-06-23
Add compatibility with DING for Dynamic Transparency

skoruppa:
This patch adds compatibility with Desktop Icons NG for Dynamic Transparency functionality by taking into account that, under Wayland, the "skip-taskbar" property can be set only under very specific circumstances (that, currently, only DING does fulfill).

You can read about issue details at https://gitlab.com/rastersoft/desktop-icons-ng/-/issues/230

### DIFF
--- a/theming.js
+++ b/theming.js
@@ -456,7 +456,8 @@ var Transparency = class DashToDock_Transparency {
         let windows = activeWorkspace.list_windows().filter(function(metaWindow) {
             return metaWindow.get_monitor() === dash._monitorIndex &&
                    metaWindow.showing_on_its_workspace() &&
-                   metaWindow.get_window_type() != Meta.WindowType.DESKTOP;
+                   metaWindow.get_window_type() !== Meta.WindowType.DESKTOP &&
+                   (!Meta.is_wayland_compositor() || !metaWindow.skip_taskbar);
         });
 
         /* Check if at least one window is near enough to the panel.


3v1n0:
Mh, while I think this can be fine, I feel it could be too generic, and we could end up considering other windows other than DING, can we maybe use a stricter filter here?

skoruppa:
Let me just quote DING's developer (@rastersoft) comment from https://github.com/lamarios/gnome-shell-extension-transparent-top-bar/issues/9 (exact same thing but for different extension)

> Hi, I'm the creator of Desktop Icons ND (DING). Unfortunately, under Wayland, there aren't thinks like a "Desktop window" or similar, so I had to use several tricks to emulate a desktop with icons. But fortunately I think that there is a way of solving this: in Wayland, no application can set the "skip_from_window_list" flag, which is mandatory for things like a desktop window. But an extension can set it under some very specific circumstances (which were put there specifically for DING, BTW). So a solution is to check if we are under Wayland AND a window has the "skip_from_window_list" flag set, and in that case set the top bar as transparent.

So it seems no other windows will use this flag. I based my change on his

vanvugt:
This looks like exactly the right fix.

I share Marco's concerns that the mutter window flags here are poorly named and vague, but that ship has sailed some years ago. If we want to fix that then we can discuss it separately.

vanvugt (approved):
This looks right to me, so approved. It should help with https://launchpad.net/bugs/1981837

vanvugt:
Confirmed to fix https://launchpad.net/bugs/1981837

Coeur-Noir:
Unfortunately problem is back - noticed today on a fresh Ubuntu 22.04 installation.

gnome-shell-extension-ubuntu-dock is 72~ubuntu5.22.04.1
and
gnome-shell-extension-desktop-icons-ng is 43-2ubuntu1

vanvugt:
Since this bug is closed already, please open a new one.

Coeur-Noir:
Not related to DiNG but… **wslu**
→ https://askubuntu.com/questions/1410764/ubuntu-22-04-fresh-install-desktop-entries-do-not-open

Removing **wslu** package restores expected behavior.

PR #1777, opened 2022-07-20
Ensure that PanelCorner has given properties before checking the value

calexandru2018:
Since Gnome42 has deprecated panel corners, for optimum backwards compatibility the method `_adjustPanelCorners` and `_revertPanelCorners` should first check if the `PanelCorner` has the left and right corner, and only then check for its value.

Source for Gnome42 port documentation: https://gjs.guide/extensions/upgrading/gnome-shell-42.html#panel-corner


3v1n0:
I've cleaned it, thanks.

3v1n0 (approved):

PR #1793, opened 2022-08-24
docking: Check for corners availability using standard checks

3v1n0:
Object.hasOwn isn't supported by mozjs 91


sidevesh:
Seeing the `Object.hasOwn is not a function` error on Gnome 42.4 with the latest code on master.

PR #1804, opened 2022-08-28
Preserve dock opacity when draging window preview in shell overview

yellowgh0st:
In order to maintain visual rhythm, symmetry and consistency I suggest this change that preserves dock opacity during the window preview drag user event.

With this change, the workspace body that allows user to drop the window on to workspace or go to desired workspace by click, gets on both left and right sides visually equal width. I say visually because technically the workspace bodies still differ
because the part of the left one is being hidden under the dock.

During the drag event dock doesn't provide any functionality and it actually prevents dropping the window which can be confusing when is transparent.

By leaving the dock's opacity same as is top bar's we achieve even more consistency.

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -241,18 +241,6 @@ var DockDash = GObject.registerClass({
             Main.overview,
             'item-drag-cancelled',
             this._onItemDragCancelled.bind(this)
-        ], [
-            Main.overview,
-            'window-drag-begin',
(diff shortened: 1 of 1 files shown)


3v1n0:
Mh, personally I think it makes sense to make it opaque in case it was not to make clear that such operation has not to be performed on dock.

You can make it an option, but in general this should happen IMHO.

Issue #1798, opened 2022-08-26
Dock icon animation does not obey OS setting "disable animations"


grandinj:
what the title says.

Specifically, this is the animation that occurs when the application is triggered somehow.

grandinj:
Thanks Marco

PR #1808, opened 2022-08-31
docking: Respect St.Settings for animation duration

3v1n0:
Ensure that the animation duration is computed according to the St
Settings values, taking in account both the slowdown factor and the
animation disablement.

Closes: #1798


3v1n0:
Nobody against, so let's get this in.

Issue #1493, opened 2021-07-29
Dominant color CSS doesn't work for some apps


Misery7100:
Ubuntu 20.04.2 LTS
GNOME Shell 3.36.9

I've seen similar issue, but don't know why now it like that. All is fine without dominant color, naturally

![image](https://user-images.githubusercontent.com/68801526/127558848-89f9eeba-460e-4b79-a75e-4543c277a142.png)

Misery7100:
At least adding fallback with default `white` color makes it better

https://github.com/micheleg/dash-to-dock/blob/302c6932617fc12b8f9382db71906e7ec32fff3b/appIconIndicators.js#L385-L391

cjkr:
@Misery7100, the solution you offered worked for me if I add a try-catch block as so:

```javascript
// Apply dominant color if requested
if (settings.get_boolean('running-indicator-dominant-color')) {
   try {
      let colorPalette = this._dominantColorExtractor._getColorPalette();
      if (colorPalette !== null) {
         this._bodyColor = Clutter.color_from_string(colorPalette.original)[1];
      } 
   } catch {
      this._bodyColor = Clutter.color_from_string('white')[1];
   }
}
```

jrom99:
I don't know if this is any better, but I found it easier to just reuse the other setting for this, since they are mutually exclusive at this point.

```javascript
 // Apply dominant color if requested
 if (settings.get_boolean('running-indicator-dominant-color')) {
     try {
        let colorPalette = this._dominantColorExtractor._getColorPalette();
        if (colorPalette !== null) {
            this._bodyColor = Clutter.color_from_string(colorPalette.original)[1];
        } else {
             this._bodyColor = Clutter.color_from_string(settings.get_string('custom-theme-running-dots-color'))[1];
        }
    } catch {
        this._bodyColor = Clutter.color_from_string(settings.get_string('custom-theme-running-dots-color'))[1];
    }
 }
```

PR #1809, opened 2022-08-31
Fix some icons missing dominant color

jrom99:
This is an attempt to fix micheleg/dash-to-dock#1493

Some icons are not Gio.ThemedIcon, but Gio.LoadableIcon, so we can get their pixel buffer from memory. Also, if this fails, we can use the customized setting to set a sensible fallback.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -12,5 +12,5 @@
 "original-author": "micxgx@gmail.com",
 "url": "https://micheleg.github.io/dash-to-dock/",
 "gettext-domain": "dashtodock",
-"version": 83
+"version": 84
 }


appIconIndicators.js
```diff
@@ -379,8 +379,11 @@ var RunningIndicatorDots = class DashToDock_RunningIndicatorDots extends Running
             // Apply dominant color if requested
             if (settings.runningIndicatorDominantColor) {
                 let colorPalette = this._dominantColorExtractor._getColorPalette();
+                let fallbackColor = Clutter.color_from_string(settings.customThemeRunningDotsColor)[1];
                 if (colorPalette !== null) {
                     this._bodyColor = Clutter.color_from_string(colorPalette.original)[1];
+                } else {
+                    this._bodyColor = fallbackColor;
```
3v1n0, line 386:
```suggestion
                const [gotFallbackColor, fallbackColor] = Clutter.color_from_string(settings.customThemeRunningDotsColor);
                if (colorPalette !== null) {
                    this._bodyColor = Clutter.color_from_string(colorPalette.original)[1];
                } else if (gotFallbackColor) {
                    this._bodyColor = fallbackColor;
```

appIconIndicators.js
```diff
@@ -941,14 +944,17 @@ var DominantColorExtractor = class DashToDock_DominantColorExtractor {
         if (iconTexture instanceof Gio.FileIcon) {
             // Use GdkPixBuf to load the pixel buffer from the provided file path
             return GdkPixbuf.Pixbuf.new_from_file(iconTexture.get_file().get_path());
+        } else if (iconTexture instanceof Gio.ThemedIcon) {
+            // Get the pixel buffer from the icon theme
+            let icon_info = themeLoader.lookup_icon(iconTexture.get_names()[0], DOMINANT_COLOR_ICON_SIZE, 0);
```
3v1n0, line 949: Please handle the case we can't find any icon.

appIconIndicators.js
```diff
@@ -941,14 +944,17 @@ var DominantColorExtractor = class DashToDock_DominantColorExtractor {
...
+            if (icon_info !== null)
+                return icon_info.load_icon();
+            else
+                return null;
+        } else {
+            // Use GdkPixBuf to load the pixel buffer from memory
+            return GdkPixbuf.Pixbuf.new_from_stream(iconTexture.load(DOMINANT_COLOR_ICON_SIZE, null)[0], null);
```
3v1n0, line 956: I think it would be possible to do this in an async way, by doing some rerfactor so that the the value is cached.

3v1n0 (changes requested):

appIconIndicators.js
jrom99: What should happen when you can't get the fallback color? Is it possible?

appIconIndicators.js
jrom99: The case where you can't find any icon within the `Gio.ThemedIcon`?

3v1n0: I mean if `icon_info` is null (also use `const` for it).

appIconIndicators.js
3v1n0: Check https://gjs.guide

appIconIndicators.js
3v1n0: Well that's what the bindings provide us, and no it won't never be a null value in such case IIRC.

appIconIndicators.js
jrom99: Got it, do you think it would be better to use then [`choose_icon`](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/IconTheme.html#Gtk.IconTheme.choose_icon) instead of [`lookup_icon`](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/IconTheme.html#Gtk.IconTheme.lookup_icon) or try to iterate over the names ourselves?

PR #1838, opened 2022-09-30
Draft: docking: Call back for the overview runStartupAnimation ASAP

vanvugt:
docking: Call back for the overview runStartupAnimation ASAP
    
That means when it is finished, before we start our own dock animation.
    
Although it was theoretically OK to call back after we finish our dock
slide animation, it was unnecessary to wait that long and would also fail
to call back under certain conditions:
    
1. Log into a Xorg session in a virtual machine.
    
2. The startup animations begin.
    
3. The hypervisor changes the Xorg display resolution.
    
4. The startup animations continue but never receive any onComplete or
   onStopped.
    
5. global.window_group would never have its 
   set_clip(<old monitor geometry>) removed.
    
Fundamentally this is a bug somewhere in clutter or gnome-shell, but 
working around it in the dock is not a change we will ever need to undo
because this change is also logically correct; the callback for the 
overview animation still occurs after the overview animation. Just more
reliably now.
  
Fixes: https://launchpad.net/bugs/1989170, https://launchpad.net/bugs/1989726

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -2097,8 +2097,8 @@ var DockManager = class DashToDock_DockManager {
                 DockManager.allDocks.forEach(dock => (dock.opacity = 0));
                 injections.add(dockManager.mainDock.dash, 'ease', () => {});
                 let callbackArgs = [];
-                const ret = await originalMethod.call(this,
-                    (...args) => (callbackArgs = [...args]));
+                const ret = await originalMethod.call(this, () => {});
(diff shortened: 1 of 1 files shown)


vanvugt:
Back to draft: Turns out the overview animation hasn't really finished at this point so it still needs work.

vanvugt:
Replaced by #1840.

PR #1840, opened 2022-10-03
docking: Handle when dash is destroyed during the login animation

vanvugt:
This can happen in a VM where the hypervisor changes the Xorg display resolution at startup, during the login animation.

Fixes: https://launchpad.net/bugs/1989170, https://launchpad.net/bugs/1989726

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -2018,8 +2018,14 @@ var DockManager = class DashToDock_DockManager {
             }
 
             const mainDockProperties = {};
-            if (dock === this.mainDock)
-                mainDockProperties.onComplete = callback;
+            if (dock === this.mainDock && callback !== undefined) {
+                const destroyId = dash.connect('destroy',
+                    () => mainDockProperties.onStopped(false));
+                mainDockProperties.onStopped = (finished) => {
+                    dash.disconnect(destroyId);
+                    callback();
+                }
+            }
 
             dash.ease({
                 opacity: 255,


docking.js
```diff
@@ -2017,9 +2017,17 @@ var DockManager = class DashToDock_DockManager {
...
+            if (dock === this.mainDock) {
+                let destroyId = 0;
+                newCallback = () => {
+                    if (destroyId)
+                        dash.disconnect(destroyId);
+
+                    callback();
+                };
+                destroyId = dash.connect('destroy', newCallback);
+            }
```
3v1n0, line 2030: I'd prefer to keep using `mainDockProperties`, so can be extended in future, so here:

```suggestion
            const mainDockProperties = {}
            if (dock === this.mainDock) {
                const destroyId = dash.connect('destroy', () =>
                    mainDockProperties.onStopped());
                mainDockProperties.onStopped = () => {
                    dash.disconnect(destroyId);
                    callback();
                };
            }
```

docking.js
```diff
@@ -2028,7 +2036,7 @@ var DockManager = class DashToDock_DockManager {
                 delay: STARTUP_ANIMATION_TIME,
                 duration: STARTUP_ANIMATION_TIME,
                 mode: Clutter.AnimationMode.EASE_OUT_QUAD,
-                ...mainDockProperties,
+                onStopped: newCallback,
```
3v1n0, line 2039: Thus here can be:

```suggestion
                onStopped: () => {},
                ...mainDockProperties,
```

3v1n0 (changes requested):
I see the point, and looks fine. But please see the code cleanups, so that we can keep the specific props for main dock, in case we'd need in future again for more stuff.

As per the switch from `onComplete` to `onStopped`, why that (for the non-destroy case)?

Should we maybe avoid calling `callback()` when the `onStopped()` `isFinished` argument is `false`?

3v1n0:
> Should we maybe avoid calling `callback()` when the `onStopped()` `isFinished` argument is `false`?

If that's is the case we can just use:

```js
if (dock === this.mainDock) {
    const destroyId = dash.connect('destroy', () =>
        mainDockProperties.onStopped());
    mainDockProperties.onStopped = () => dash.disconnect(destroyId);
    mainDockProperties.onComplete = () => callback();
}
```

docking.js
3v1n0: Oh, actually this is not needed at all, don't think we need to call anything at all, so just keep `...mainDockProperties`.

docking.js
vanvugt: I wouldn't call `mainDockProperties` a cleanup. It's code that doesn't need to exist and is misleading because it's being passed in to all docks, not just the main one. Only its value changes. So the name is wrong at least: `s/mainDockProperties/dockProperties/`.

vanvugt:
> 

That doesn't work because you forgot to callback when the destroy happens. Also onStopped takes a boolean. But I will adapt it thanks.

docking.js
vanvugt: Done.

docking.js
vanvugt: Done.

3v1n0:
> > 
> 
> That doesn't work because you forgot to callback when the destroy happens. Also onStopped takes a boolean. But I will adapt it thanks.

Yeah, sorry should have been calling `mainDockProperties.onComplete()` during  on ondestroy cb, if the parameter is not used, there's no need to expose it.

3v1n0:
However, my point is that we're now calling the original callback also when the animation didn't complete, shouldn't we just call it still when `onComplete`?

Also is it needed to complete the call the callback on destruction?

vanvugt:
> However, my point is that we're now calling the original callback also when the animation didn't complete, shouldn't we just call it still when `onComplete`?

onStopped(true) is the same thing as onComplete.

> 
> Also is it needed to complete the call the callback on destruction?

Yes completing the callback on destruction is the whole bug fix. Without that this is pointless exercise :)

3v1n0:
> onStopped(true) is the same thing as onComplete.

Indeed, but `finished` is not used

docking.js
```diff
@@ -2018,8 +2018,14 @@ var DockManager = class DashToDock_DockManager {
             }
 
             const mainDockProperties = {};
-            if (dock === this.mainDock)
-                mainDockProperties.onComplete = callback;
+            if (dock === this.mainDock && callback !== undefined) {
```
3v1n0, line 2021:
```suggestion
            if (dock === this.mainDock && callback) {
```
Otherwise we'd could call a `null` callback too.

docking.js
```diff
@@ -2018,8 +2018,14 @@ var DockManager = class DashToDock_DockManager {
...
+                const destroyId = dash.connect('destroy',
+                    () => mainDockProperties.onStopped(false));
+                mainDockProperties.onStopped = (finished) => {
+                    dash.disconnect(destroyId);
```
3v1n0, line 2025: Can just use

```suggestion
                    () => mainDockProperties.onStopped());
                mainDockProperties.onStopped = () => {
                    dash.disconnect(destroyId);
```

Since we're the one defining that function and `finished` being `undefined` is fine anyways for the way is checked everywhere.

vanvugt: Half correct.

If anyone else is listening to onStopped then they expect to receive a boolean. Only we don't use it, so you could remove the `finished` but can't remove the `false`. I recommend keeping both to make it clear there's a parameter whose value should be correct even when we don't use it.

docking.js
vanvugt: Sure, not important. Although I think a cleaner solution is to modify the caller to pass in `() => {}` since it's an expected parameter.

docking.js
3v1n0: I was thinking that, but we're definitely the only callback since we're setting it locally.

However, I'm fine if you want to keep it to remember the actual type "signature", but if you do please use the `_mainDockProperties.onStopped = _finished => {` syntax.

vanvugt: That suggestion seems to contain two mistakes:
 * No parentheses (which I assume is a mitsake); and
 * The actual parameter (according to `environment.js`) is called `isFinished`. But I didn't think the name mattered much.

So maybe now I will just remove the parameter to avoid further discussion because this is a waste of our time.

vanvugt: I don't think either of those underscores make sense in the existing function or in the existing coding convention (we don't underscore locals or parameters) so I'm going to ignore that and merge as-is.

docking.js
vanvugt: I'm not sure null as a callback is something we need to set a precedent for or support. Anyway, I've solved it in https://github.com/micheleg/dash-to-dock/pull/1843 instead.

docking.js
3v1n0: > No parentheses (which I assume is a mitsake)

No it's the expected style (see eslint, we don't complain here but it would if would be stricter as shell/gjs is in future, and we will)

> The actual parameter (according to environment.js) is called isFinished. But I didn't think the name mattered much.

Of course not

> I don't think either of those underscores make sense in the existing function or in the existing coding convention (we don't underscore locals or parameters)

We do underscore when it's not used, please there is a style we want to follow and we should in order to make this extension to be soon more compatible with what upstream linting expects us to do.

I would have merged once fixed (or I would have taken care), please it would be nice if you'd follow the suggestions without merging by yourself.

3v1n0:
> It would be used in other peoples' onStopped callbacks, if they exist. So the boolean must exist.

_Impossible_, unless someone monkey patches our own extension, overrides that `ease` call and eventually try to reuse our callback function. And anyways FWIW, whether is `undefined` or `false`, `!!isFinished` or `!isFinished` has the same return value.

PR #1842, opened 2022-10-05
docking: _prepareStartupAnimation earlier when disableOverviewOnStartup

### LINKED ISSUE #1770: If 'Show overview on startup' is disabled, dock "jumps up" onсe during GNOME Shell launch
After disabling the 'Show overview on startup' option, overview on startup is really disabled, but the dock "jumps up" once. This happens with both version 72 of Fedora's 36 official repo and the version maked from git sources.

But if you leave the 'Show overview on startup' option enabled and at the same time use the [No overview at start-up by fthx](https://extensions.gnome.org/extension/4099/no-overview/) extension, when starting the GNOME Shell, the desktop and dock are simply shown without any unnecessary animations.

vanvugt:
So that dock.dash.opacity=0 before the original animation runs.

Definitely fixes: https://launchpad.net/bugs/1965208 but might also resolve #1770.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -2281,6 +2281,7 @@ var DockManager = class DashToDock_DockManager {
 
         if (Main.layoutManager._startingUp && Main.sessionMode.hasOverview &&
             this._settings.disableOverviewOnStartup) {
+            this._prepareStartupAnimation();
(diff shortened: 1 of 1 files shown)


vanvugt:
Replaced by #1846

PR #1843, opened 2022-10-06
docking: Remember the parameter to _runStartupAnimation()

vanvugt:
It was never meant to be optional.


3v1n0:
No need for that, it's our code, and this would also imply us handling the destruction that is not needed in this case, as there's no cb.

null handling is now in #1844

PR #1844, opened 2022-10-06
Startup animation cleanups

3v1n0:


vanvugt:
I will test this soon.

Please also remember #1842 which is probably the fix that regular users will notice.

3v1n0:
This is just a temporary cleanup, while I'd cover #1842 case and the one
fixed in another upcoming branch.

Because with your previous fix handling destruction indeed we fix the issue
but we don't animate anymore in such cases. While it's a minor thing, I'd
also like to get that handles handled too.

Il ven 7 ott 2022, 03:58 Daniel van Vugt ***@***.***> ha
scritto:

vanvugt (approved):
Yes this seems to work. Sorry about the semicolon.

PR #1846, opened 2022-10-07
docking: Ensure we perform the startup animation completely on docks updates

3v1n0:
If docks gets destroyed early during startup animation (as it may happen
when there are resolution or other monitor changes), we need to
completely finish the animation, but we can safely perform a new one in
these cases, as this happens early enough.

Closes: #1770
LP: #1965208

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -2123,19 +2123,23 @@ var DockManager = class DashToDock_DockManager {
 
         this._methodInjections.addWithLabel(Labels.MAIN_DASH, ControlsManager.prototype,
             'runStartupAnimation', async function (originalMethod, callback) {
-                const injections = new Utils.InjectionsHandler();
-                const dockManager = DockManager.getDefault();
-                dockManager._prepareStartupAnimation(callback);
-                injections.add(dockManager.mainDock.dash, 'ease', () => {});
-                let callbackArgs = [];
-                const ret = await originalMethod.call(this,
-                    (...args) => (callbackArgs = [...args]));
-                injections.destroy();
+                try {
+                    const injections = new Utils.InjectionsHandler();
+                    const dockManager = DockManager.getDefault();
+                    dockManager._prepareStartupAnimation(callback);
+                    injections.add(dockManager.mainDock.dash, 'ease', () => {});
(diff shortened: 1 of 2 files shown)


vanvugt:
> If docks gets destroyed early during startup animation (as it may happen
> when there are resolution or other monitor changes), we need to
> completely finish the animation, but we can safely perform a new one in
> these cases, as this happens early enough.

I think you've confused three separate bugs:
 * Animation does not restart if interrupted (no bug ID exists yet).
 * Clipping is broken if interrupted (LP#1989170, LP#1989726)
 * Login animation runs twice (#1770 and LP#1965208)

From what I can tell you are only fixing the first one so shouldn't mention any bug IDs.

vanvugt:
Please also remember we need very small patches to backport to jammy.

3v1n0:
> Clipping is broken if interrupted (LP#1989726, LP#1989726)

The fact is that the fix we've now is not covering two possible cases in which the destruction happens even earlier (which was already somewhat work-arounded, but not in the nicest way), and the clipping was still adjusted a bit late still causing the animation to happen with the wrong screen size leading to a side of the screen being "gray" for a while till animation completed.

This handles them all and that's all tiny enough to be backported alltogether.

3v1n0:
Ah, and per

> From what I can tell you are only describing the first issue so shouldn't be declaring any bug IDs as fixed here.

The other issues are definitely fixed by this, as the reason they were happening was that we were showing the dash too early as per callback completion, but this is not the case anymore as once we have a monitor reconfiguration we're just resetting everything to the start (including the dash being hidden).

Something I was just unsure about  was whether to restart the whole animation all the times or just on the first "destroy", while just complete the startup without animation if this is happening more than once. But that's only doubt I've.

vanvugt:
OK, manual testing is showing good but weird results. The old bugs are fixed, but there's a new issue (probably mutter's fault): When I change the display resolution during the login animation it only works 75% of the time. The remaining 25% it stays in the old resolution. Although I have also changed from HDMI to DisplayPort this week so :man_shrugging: 

Good enough for me.

PR #1854, opened 2022-10-15
Windows previews improvements

3v1n0:
Various cleanups and improvements to the window previews, and finally fixing a very-old bug that was causing the previews not to be properly aligned.

### DIFF
--- a/windowPreview.js
+++ b/windowPreview.js
@@ -95,6 +95,7 @@ var WindowPreviewList = class DashToDock_WindowPreviewList extends PopupMenu.Pop
             name: 'dashtodockWindowScrollview',
             hscrollbar_policy: St.PolicyType.NEVER,
             vscrollbar_policy: St.PolicyType.NEVER,
+            overlay_scrollbars: true,
             enable_mouse_scrolling: true
         });
 
--- a/windowPreview.js
+++ b/windowPreview.js
@@ -25,6 +25,8 @@ const PREVIEW_MAX_HEIGHT = 150;
 const PREVIEW_ANIMATION_DURATION = 250;
 const MAX_PREVIEW_GENERATION_ATTEMPTS = 15;
 
+const MENU_MARGINS = 10;
+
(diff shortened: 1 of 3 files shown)


windowPreview.js
```diff
@@ -362,13 +374,28 @@ class WindowPreviewMenuItem extends PopupMenu.PopupBaseMenuItem {
...
+        // For some crazy clutter / St reason we can't just have this handled
+        // automatically or here via vfunc_allocate + vfunc_get_preferred_*
+        // because if we do so, the St paddings on first / last child are lost
+        const themeNode = this.get_theme_node();
+        let [minWidth, naturalWidth] = this._box.get_preferred_width(-1);
+        let [minHeight, naturalHeight] = this._box.get_preferred_height(naturalWidth);
+        [minWidth, naturalWidth] = themeNode.adjust_preferred_width(minWidth, naturalWidth);
+        [minHeight, naturalHeight] = themeNode.adjust_preferred_height(minHeight, naturalHeight);
+        this.set({ minWidth, naturalWidth, minHeight, naturalHeight });
+    }
```
3v1n0, line 397: @fmuellner Hey, do you have any clue why this ugly stuff is the only way to be able to get the size to be adjusted to use the actual padding to that is defined for first and last children?

To explain, this is basically just a `PopupMenu.PopupBaseMenuItem`, if from css I set something like:

```scss
       &:ltr:first-child {
            padding-left: 100px;
            padding-top: 40px;
        }
        &:ltr:last-child {
            padding-right: 100px;
        }
```
then, the children are positioned like if that is taken into account, but the size is not adjusted (in fact `vfunc_get_natural_width` is never called after the first time is set), and forcing allocating would never get the first/last items size in account.

So getting this:

image

Instead of the expected

image

3v1n0: Sure, sure... Here it seems though that it's correctly computed and assigned, the problem is that the size of `PopupBaseMenuItem` is not adjusted when the first children is set to have some specific padding values.

It's weird, as applying something similar upstream works in all the other menu items I tried, while this one has nothing special.

PR #1879, opened 2022-11-18
Fix intellihide conflict with desktop icons gnome shell extension

dr3mro:
Hello, 
I have reported this and managed to find a patch, Please I would appreciate accepting it into main branch.
Thank you. 
Amr

### DIFF
--- a/intellihide.js
+++ b/intellihide.js
@@ -267,7 +267,7 @@ var Intellihide = class DashToDock_Intellihide {
                         return true;
 
                     // Desktop icons extension is not an application so it should be ignored
-                    else if (meta_win.get_wm_class() == 'com.desktop.ding')
+                    else if (meta_win.is_skip_taskbar())
                         return false;
 
                     let currentApp = this._tracker.get_window_app(meta_win);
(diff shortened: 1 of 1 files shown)


smedir:
Hi,

This commit will not fix the problem, probably for the Gtk3 version of DING (Although I have not tried it with the Gtk3 version).

The better very specific solution may be- 

`
At the top of the function at line 322, in _handledwindow() -

const ignoreApps = [ 'com.rastersoft.ding',  'com.desktop.ding' ];
const wmClass = meta_win.get_wm_class();

else if (ignoreApps.includes(wmClass) && meta_win.is_skip_taskbar())
        return false;`

A more generic solution at the same level could simply be

`else if (meta_win.is_skip_taskbar())
       return false;`

Would have to be checked in X11, but I think it will work....

dr3mro:
That's awesome 😍 but I can confirm my solution works with gnome 43 on fedora 37

smedir:
Yes it will work perfectly for the Gtk4 DING! :smile:

dr3mro:
Anyways, please do what ever necessary to fix it.
have a nice day 
👍

dr3mro:
please accept pull request

dr3mro:
I have updated the code please recheck and accept the pull request.

smedir:
I tried this out, this will not work when case IntellihideMode.ALL_WINDOWS is considered in dodge windows. The dock will hide all the time.

I will submit a new pull request with the very specific fix I proposed earlier.

3v1n0:
Duplicate of #1890

PR #1888, opened 2022-12-09
Fix Transparency for Gtk4 DING

smedir:
In Gtk4-DING, the Gtk4 port of the desktop icons extension, the desktop X11 window cannot be detected by "DESKTOP" type. It is handled by the same code as wayland windows, and needs to be detected by the skip_taskbar hint. This enables correct transparency of the dock when on X11 with Gtk4-DING.

This, works with #1879, which is needed for fixing intellihide of Dock with Gtk4-DING.


3v1n0:
Is this creating any problem with the gtk3 version?

smedir:
No, seems to work with both the Gtk3 and Gtk4 branches on both X11 and 
Wayland.

Issue #1859, opened 2022-10-19
Please offer a setting to turn off icon shaking


hosford42:
Please offer a setting under Accessibility to completely turn off the icon shaking feature. Anyone with sensory processing disorder, like myself, is likely to struggle significantly with filtering out motion in their peripheral vision (along with any other sensory stimulus that isn't directly relevant to the current task). This is not a rare condition; it is common for neurodivergent people -- those with ADHD or ASD, for example.

NOTE: I just upgraded to Ubuntu 22.04.01 LTS. I don't have this extension installed, but it appears to be incorporated into the OS now. Apologies if that's not the case and I have the wrong repo.

vanvugt:
Two workaround suggestions:
 * Edit `/usr/share/gnome-shell/extensions/ubuntu-dock@ubuntu.com/docking.js` and change `ICON_ANIMATOR_DURATION` to zero. At least that looks like the offending code.
 * Disable the dock and use the default GNOME one instead:
   `gnome-extensions disable ubuntu-dock@ubuntu.com`

lordspace:
yeah, that's super annoying. Who came up with that stupid idea and who OK'ed it?!?
Our attention is pulled in so many ways anyways.

PR #1896, opened 2023-01-04
Make urgent application icon dance toggleable

ZimbiX:
Resolves https://github.com/micheleg/dash-to-dock/issues/1859.

The distracting dance/shake animation for urgent application icons has always really annoyed me. I finally thought I'd do something about it and contribute a setting for whether dancing should be enabled; and being mostly a copy-paste job, was easier than I expected.

Tested using gnuplot, which when not minimised and not the foreground application, makes its icon urgent upon `reread`:

```bash
echo 'plot sin(x); pause 1; reread' > x && gnuplot x; rm x
```

Edit: I wasn't sure what to do about translations. Is there something automated? Should we be using Google Translate?

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -162,9 +162,11 @@ var DockAbstractAppIcon = GObject.registerClass({
         this.connect('notify::urgent', () => {
             const icon = this.icon._iconBin;
             this._signalsHandler.removeWithLabel(Labels.URGENT_WINDOWS)
-            if (this.urgent && Docking.DockManager.settings.danceUrgentApplications) {
-                icon.set_pivot_point(0.5, 0.5);
-                this.iconAnimator.addAnimation(icon, 'dance');
+            if (this.urgent) {
+                if (Docking.DockManager.settings.danceUrgentApplications) {
+                    icon.set_pivot_point(0.5, 0.5);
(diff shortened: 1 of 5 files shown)


appIcons.js
```diff
@@ -162,7 +162,7 @@ var DockAbstractAppIcon = GObject.registerClass({
         this.connect('notify::urgent', () => {
             const icon = this.icon._iconBin;
             this._signalsHandler.removeWithLabel(Labels.URGENT_WINDOWS)
-            if (this.urgent) {
+            if (this.urgent && Docking.DockManager.settings.danceUrgentApplications) {
                 icon.set_pivot_point(0.5, 0.5);
                 this.iconAnimator.addAnimation(icon, 'dance');
```
3v1n0, line 167:
```suggestion
            Mh, I feel it would be better to do something like:

```suggestion
            if (this.urgent) {
                if (Docking.DockManager.settings.danceUrgentApplications) {
                    icon.set_pivot_point(0.5, 0.5);
                    this.iconAnimator.addAnimation(icon, 'dance');
                }
```

Otherwise we might break other logic that depends on urgent windows state.

ZimbiX: Thanks for the review. I don't know what it'd break, but that seems to work the same, so alright

PR #1908, opened 2023-01-25
translation: Mark some string translatable again

sabriunal:
Makefile uses --keyword=__ to extract strings.

This patch fixes string extraction error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,8 @@ EXTRA_MODULES = \
                 theming.js \
                 utils.js \
                 dbusmenuUtils.js \
-                Settings.ui desktopIconsIntegration.js \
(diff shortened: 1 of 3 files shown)


3v1n0:
I'm not sure about, for these strings we should already have upstream translations so there's no need to duplicate.

sabriunal:
Sorry for not insufficient information.  This patch fixes string extraction error. This error occurs when you extract string with xgettext. It up to developer to fix issue. I've tested and got error extracting these strings.

3v1n0:
Sure, the point is that not all the strings must be extracted, as many are already provided by gnome-shell and there's no need to duplicate the translations efforts.

PR #1937, opened 2023-02-27
fix: this._launchMaxWaitIds is undefined

### LINKED ISSUE #1935: this._launchMaxWaitIds is undefined
Hi,

On my Fedora Linux 37/Gnome 43 i receive this error: `this._launchMaxWaitIds is undefined`
Which results in a not working extension for me.
I dont see an UI Endpoint to set this value.

As far as i see, other people seem to have the same Problem as me: https://extensions.gnome.org/extension/307/dash-to-dock/

dav1app:
fix #1935

### DIFF
--- a/locations.js
+++ b/locations.js
@@ -394,7 +394,7 @@ var LocationAppInfo = GObject.registerClass({
     }
 
     destroy() {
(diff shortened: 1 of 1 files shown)


3v1n0:
Thanks!

PR #1953, opened 2023-03-01
Hide tooltip

SubZ69:
Hide application tooltip option

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -1326,6 +1326,43 @@
                         </property>
                       </object>
                     </child>
+                    <child>
+                      <object class="GtkListBoxRow" id="hide_tooltip_row">
+                        <property name="child">
+                          <object class="GtkGrid">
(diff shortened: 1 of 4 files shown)


docking.js
```diff
@@ -574,6 +574,12 @@ const DockedDash = GObject.registerClass({
             'changed::dance-urgent-applications',
             () => this.dash.resetAppIcons(),
             Utils.SignalsHandlerFlags.CONNECT_AFTER,
+        ], [
+            settings,
+            'changed::hide-tooltip',
+            () => {
+                this.dash.resetAppIcons();
+            },
```
3v1n0, line 582: I'm not sure this is needed, at the end this should be applied immediately once you hover the icon anyways.

SubZ69: You are right, this is not needed. 
Do you plan to accept this PR? Do I need to make a new one without docking.js change?

PR #1963, opened 2023-03-07
Update German translation

Amr-Ibra:
Update German translation

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -4,136 +4,136 @@
 # Morris Jobke <hey@morrisjobke.de>, 2014
 # Jonatan Zeidler <jonatan_zeidler@hotmail.de>, 2012
 # jonius <jonatan_zeidler@gmx.de>, 2012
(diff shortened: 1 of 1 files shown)


po/de.po
```diff
@@ -367,70 +367,90 @@ msgid "Isolate volumes, devices and trash windows from file manager"
 msgstr "Datenträger, Geräte und Papierkorbfenster vom Dateimanager isolieren"
 
 #: Settings.ui.h:57
+msgid "Dance urgent applications"
```
Amr-Ibra, line 370: I translated _Dance_ (en) as _Wiggle_ (en) as if the original string would be “_Wiggle urgent applications_”, because a literal translation of _Dance_ would sound weird in German.

3v1n0: We should probably fix the upstream code too, please do a MR changing the string :)

Amr-Ibra: https://github.com/micheleg/dash-to-dock/pull/1979

Issue #1975, opened 2023-03-24
Dash to Dock GUI settings in Extensions is too large for lower-res screens


Anduril97:
Hey! You guys have done great work with this, so thanks! I am stuck using a 720p laptop and I have this issue where the window size of the Settings for Dash to Dock (in Extensions) is way too large for my screen so I can never see the bottom settings as the window is not resizeable. Conversely, if I try to drag the window up to reach the bottom the title bar goes offscreen, again due to the window not being resizable. Could you allow the settings window to be smaller with a scrollbar maybe?
Thank you!

Franciel-Victor:
I run into the same problem!

vanvugt:
The GNOME Extensions GUI does seem to be resizable *and* automatically shrinks to fit small screens. I just tested 800x600 on both GNOME 42 and 44. What GNOME versions are you using?

vanvugt:
Is the window not scrollable, or just not obviously scrollable?

Anduril97:
Although the gnome extensions gui does resize to fit the screen resolution, for some reason the Settings page for Dash to dock still displays full size to the point that I cannot access the lower options without moving the title bar beyond the screen (upwards). I tried scrolling it and it did not work for me.

I'm on Gnome version 43.3 (Manjaro)

PR #1977, opened 2023-03-28
prefs: Set default window height to 850, if it fits

vanvugt:
Instead of enforcing a minimum window height of 850. The height is now resizable.

Fixes: https://github.com/micheleg/dash-to-dock/issues/1975

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -227,7 +227,7 @@ var Settings = GObject.registerClass({
         // Set a reasonable initial window height
         this.widget.connect('realize', () => {
             const rootWindow = this.widget.get_root();
-            rootWindow.set_size_request(-1, 850);
+            rootWindow.set_default_size(-1, 850);
             rootWindow.connect('close-request', () => this._onWindowsClosed());
         });
 


prefs.js
```diff
@@ -227,7 +227,6 @@ var Settings = GObject.registerClass({
         // Set a reasonable initial window height
         this.widget.connect('realize', () => {
             const rootWindow = this.widget.get_root();
-            rootWindow.set_size_request(-1, 850);
```
3v1n0, line 230: Please use instead:

```js
if (!Me)
    rootWindow.set_size_request(-1, 850);
```

So that it will still have some reasonable size (can be even smaller) when running as standalone window (i.e. just running `gjs prefs.js` for development purposes).

vanvugt: Not quite that syntax because 850 is still too big for some displays.

3v1n0: That's just for the debugging case, were we need to provide a default window size... Can be smaller but we still need to have one.

PR #1986, opened 2023-04-11
Fix consistency of full stops in strings

stuarthayhurst:
Convention for this extensions seems to be that titles don't have full stops, but descriptions do. This PR just fixes the 9 cases I found that break that convention.

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -520,7 +520,7 @@
                             </child>
                             <child>
                               <object class="GtkCheckButton" id="multi_monitor_button">
-                                <property name="label" translatable="yes">Show on all monitors.</property>
+                                <property name="label" translatable="yes">Show on all monitors</property>
                                 <property name="margin_top">12</property>
 
                                 <layout>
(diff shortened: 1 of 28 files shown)


3v1n0:
This looks good, but it would imply also changing all the translations... This is fine but please also remove the full stop from all the po files so that translators won't have to review all the strings just because of this.

stuarthayhurst:
As in update the translation files? Do you happen to know the commands to regenerate the files? I had a brief look when I submitted the PR, but it complained about dependencies so I figured I'd just ask instead.

3v1n0:
Just using `make` should rengerate them, however here the problem is that we want to modify the existing ones, so something like `sed 's,\."$,",g' -i po/*.po`, but I'm quite unsure that's not safe as you need to replace it only in the ones you've modified here.

stuarthayhurst:
I just wrote a script to find matches for all the strings and open the files to those lines, then fix them by hand. Seems to have worked, let me know if there's anything else to do :)

3v1n0:
Thanks a lot, this should work!

Issue #1964, opened 2023-03-07
Notification badge counter persistant or wrong number displayed


sekurilabs:
After the latest update I have noticed the notification badge count is +1 than the actual number of new notifications. For example I get an email, Geary icon badge shows 2 new emails.
For Teams for Linux for example the number of notification doesn't go away after the message is read, it's persistent and only way to clear the count is to restart the shell. Application restart doesn't reset the counter.

3v1n0:
Notification badges now also includes the tray notifications you have. This could be probably be re-considered in case an application provides its own counting though.

vageliskatsiotis:
Noticed this today after the last update of D2D in Thunderbird. When I deactivated the extension and the initial dock did not have the notification number. Reactivated and it was there again. Also  does not go away even if I read all of the emails. I have to restart gnome shell for it to go away

fabd:
Why am I seeing a (1) now on the Files icon in the dock? Just rebooted after the extension upgrade.

jrom99:
I have the same problem on Firefox after a recent update (dash-to-dock 79, gnome-shell 43.3, fedora 37).

3v1n0:
Can you verify whether you've notifications involving these apps in the notifications tray?

The reason why the emblems are added is the fix for #20, they can be disabled from d2d settings or using the do-not-disturb mode.

jrom99:
Yes, I had some Outlook notifications in the calendar top popup, but they had already been addressed within Firefox.

3v1n0:
Eh, that's not something we can check, the feature here is to show the number of unread notifications in the tray.

In case you don't want this feature, there's an option to disable it.

vladamarek:
I have the same problem and on some launchers I can't get rid of those notifications:

![image](https://user-images.githubusercontent.com/82888469/223994549-ec667432-aee5-4288-8ea2-667ebc9316de.png)

The issue appeared after upgrading gnome-shell-extension-dash-to-dock from 78-1 to 79-1.

jrom99:
In that case, the badge's presence doesn't seem to be consistent, since I can close the terminal to erase its badge, but closing Firefox won't do anything. I'll take a look if a report has been issued about the badge's number's size, since it's too small to be visible.

![image](https://user-images.githubusercontent.com/47498278/224026743-10180e2b-2c3f-4dc1-ab94-dd599517d7e8.png)

[...]

PR #1988, opened 2023-04-12
Add more options to control the emblems visibility

3v1n0:
It's now possible to:
 - Toggle the emblems at all
 - Disable the notification counters
 - Control whether the notification numbers are summed with the applications-provided counts (no by default)

Closes: #1964

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -1414,6 +1414,20 @@
                               </object>
                             </child>
 
+                            <child>
+                              <object class="GtkCheckButton" id="notifications_counter_check">
+                                <property name="label" translatable="yes">Show the number of unread notifications</property>
+                                <property name="halign">start</property>
+                                <property name="margin_top">12</property>
+
+                                <layout>
+                                  <property name="column">0</property>
+                                  <property name="row">2</property>
+                                  <property name="column-span">2</property>
+                                </layout>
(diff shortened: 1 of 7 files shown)


christophlsa:
I tested only my "case" by deactivating the checkbox "Show the number of unread notifications" and it's working fine for me. :)

vageliskatsiotis:
In my case in Thunderbird, after reading the emails, when it gets down to 1, the emblem does not go away. Tried with applications-provided counts off and on with same result. A notice in Files file manager which I happen to come across, worked as expected

3v1n0:
> Tried with applications-provided counts off and on with same result

That option is not meant to disable them (there's the global option for it), but only to sum or not the two values.

Apparently in your case there's something still sending us that the number is > 0, because in my tests we hide them otherwise.

3v1n0:
Also you can try the `hello-unity` snap and see how it behaves: https://snapcraft.io/hello-unity

vageliskatsiotis:
Thanx Marco. Using hello-unity I figured out the functionality notification wise. As long as the notification on the unity bar is not dismissed, the count remains. And has nothing to do with the fact that I have read or not the emails. Glad to clear this out. Thanx again for contributing on this!

Issue #1746, opened 2022-05-18
Feature Request: Centered Icons in Panel Mode


maxlahn:
Basically the title. I use the "panel mode" (which extends the dock), but all my favorite apps are on the left-hand side. It'd be cool we could choose for them to be centered.

Anderson0xFF:
I liked, I'm looking for something about it.

janpansa:
Please add this feature, should be easy enough?

Alt37:
Centered panel would definitely improve both presentation and usability.
1. Centered items on the panel look much closer to original dash design than current "favorite/running apps on the left, apps icon on the right"
2. I can use "Dock" instead of "Panel" to achieve similar effect. But panel look so much better and far less distracting with some bright wallpaper
3. Most other popular desktop OS (macOS, ChromeOS, Windows) center apps on the panel.

![dash_to_dock__bottom_panel](https://user-images.githubusercontent.com/44649402/195506539-7c46cb51-0c65-44e2-97d6-eb36ab811d86.jpg)
![dash_to_dock_dock](https://user-images.githubusercontent.com/44649402/195506584-fc9b657d-4dee-4910-984f-ce096aeb4a68.jpg)
![dash_to_dock__bottom_panel_centered](https://user-images.githubusercontent.com/44649402/195506604-1e24c911-ba33-4655-876d-53ab9f23810e.jpg)
![original_dash](https://user-images.githubusercontent.com/44649402/195506619-c7286b76-930a-4ca3-b991-2d9e6356005f.jpg)

I've made a simple "proof of concept" patch which works well but not ready for merge because:
- With this patch some options ("Move the application button at the beggining of the dock", "Panel mode") don't update panel properly in a real time (you also need to change "Positiion on the screen" or reload extension/Shell)
- it doesn't add option, just partly replaces current design

@3v1n0 , @micheleg 
What is your opinion about adding such option and maybe even making it default for horizontal panel?
Thanks!

3v1n0:
I'm ok adding an option, not to making or default though

roracle:
> I'm ok adding an option, not to making or default though

How long before we can expect the change? I would like this as well. :)

SassyJammy:
Yes, I would love such a feature. The only extensions that achieve this are Dash to Panel and Dash to Dock for COSMIC.

However, they both come with their downsides in my opinion:

Dash to Panel:
- Looks really square and gets rid of the GNOME feel

Dash to Dock for COSMIC:
[...]

PR #1998, opened 2023-04-19
Support centered icons in panel mode

3v1n0:
Added two options:
 - Keep icons always centered when in panel mode (under Position and Size)
 - Show _Show Apps_ button in the edge when using Panel mode (under Launchers)

This allows getting these results (same when show-apps is shown at the beginning):

image

image

image

image

Closing: #1746


kalinchernev:
@3v1n0 thanks for this :clap:

Issue #2013, opened 2023-05-18
Handling popup (or dropdown?) menus in intellihide in wayland


taoky:
Currently when Intelligent autohide is on (Dodge windows = Only focused application's windows) in GNOME wayland session, it seems that this could not handle popup menus triggered by right clicks, especially in Qt programs.

Example 1: Okular in Wayland (`QT_QPA_PLATFORM=wayland`). The menu is covered by dock, making it impossible to interact with.

![wayland-okular](https://github.com/micheleg/dash-to-dock/assets/2109893/8800a41c-0a80-4114-8eb3-1e83ddc72858)

Example 2: Telegram in Wayland. Also, the bottom part of the menu is very hard to use.

![wayland-telegram](https://github.com/micheleg/dash-to-dock/assets/2109893/cef2cfdc-0cc2-417e-834a-a1940d07f686)

Although I have not tested it in GNOME X11 session, Okular in Xwayland seems not have this issue, as the menu is above the dock:

![x11-okular](https://github.com/micheleg/dash-to-dock/assets/2109893/7588bf79-cf65-4359-bfa4-6ca0545257a8)

A partially working workaround is to append `Meta.WindowType.DROPDOWN_MENU` to `handledWindowTypes` in `intellihide.js` -- I don't know why it is `DROPDOWN_MENU` instead of `POPUP_MENU` as Meta both has these two types of windows. However only when `DROPDOWN_MENU` is set in this array does it work.

This workaround is not perfect (when "Show dock for urgent notifications" is on):

https://github.com/micheleg/dash-to-dock/assets/2109893/d6758ac0-4453-419c-8315-719d6b1f41a3

A better implementation, I think, may put the dock behind "dropdown menus". Unfortunately, I'm not quite familiar with how StBox works and I also don't know if it is possible to implement.

taoky:
Forgot to add versions...

- GNOME 44.1
- dash-to-dock 81

taoky:
And this issue also affects GTK programs:

![image](https://github.com/micheleg/dash-to-dock/assets/2109893/6628ee07-e730-4e3c-b74c-88af2e516534)

(GNOME OS 44.1)

I would submit a pull request soon about my proposed workaround.

vanvugt:
This is covered by #1887 although I'm not sure if it is solvable for the autohide case.

vanvugt:
Duplicate of #1887

taoky:
[...]

PR #2014, opened 2023-05-20
intellihide: add dropdown menu to handled window types

taoky:
This is a workaround for issue #2013. As far as I could tell this works in most cases of popup menus (by right clicks) in GNOME wayland session. The only exception I meet is the multilevel popup menu from Telegram (illustrated in #2013), but with showing urgent notifications off it also works without problems.

`Meta.WindowType.POPUP_MENU` is not added to this list as I find that it has side effects for Qt xwayland programs: on my machine the intellihide is triggered every time I right-clicked in Okular (xwayland). However I could not reproduce this within my GNOME OS VM so it might be caused by other factors.

### DIFF
--- a/intellihide.js
+++ b/intellihide.js
@@ -41,6 +41,7 @@ const handledWindowTypes = [
     Meta.WindowType.MENU,
     Meta.WindowType.UTILITY,
     Meta.WindowType.SPLASHSCREEN,
+    Meta.WindowType.DROPDOWN_MENU,
 ];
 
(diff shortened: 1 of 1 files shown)


taoky:
Just tested this in GNOME Xorg session, and confirmed that this PR will not change the behavior of dock in Xorg: menus are still above the dock and they are clickable.

vanvugt (approved):
Tested and works for me.

Issue #1612, opened 2021-12-07
Enabling dash-to-dock messes up search results navigation


yoursred:
GNOME: 41.1
dash-to-dock: 71
Kernel: 5.14.21-2-MANJAR

Description: With dash-to-dock enabled, side-to-side navigation with the arrow keys is nonfunctional, and trying to navigate takes the search bar out of focus.

yoursred:
Update: intelligent autohide seems to be cause

yoursred:
Update: enabling it, regardless of its settings, seems to be the cause

3v1n0:
Please, explain exactly how to reproduce because it's not clear to me.

If it's about the fact that using arrow-keys while moving around the search text, moves the focus to the app icons, that's an upstream bug/feature as well.

yoursred:
I guess it's a combination of gnome and the extension
![ezgif-3-028a676f4f63](https://user-images.githubusercontent.com/86563891/145079984-6f6b9cdc-728e-47fb-bc83-3f40a82156a4.gif)
.

JustCryen:
I've been having the same issue for quite some time.
I just narrowed it down to Dash To Dock but nothing more.
For now I can confirm that either disabling the extension or Intelligent Autohide feature fixes it.
OS: archlinux
Gnome: gnome-shell 1:41.1-1

yoursred:
I want to add another thing, it only happens with intelligent autohide when the dock is not extended and is on the left side of the screen.

JustCryen:
You're right… sadly that's exactly how I use it.

yoursred:
I would recommend extending it until a fix is found

JustCryen:
Actually, under Appearance, disabling "Shrink the dash" makes it work again.
That's something I can tolerate, extending it is too distracting for me.

yoursred:
Quite the bizarre bug

rollflick:
The issue in Activities Overview only occurs for me after suspending, not after a reboot. Occurs whether using Wayland or Xorg (though maybe need to do reboot rather than just logging out, which I didn't do).

Turning off Ubuntu Dock in extensions seems to have sorted it for me. Never needed to do that before for this extension to work well.

yoursred:
I can confirm that it happens regardless of session type (Xorg or Wayland), but for me at least, it doesn't care about being woke up or a cold boot.

JustCryen:
Same as @yoursred.
[...]

PR #2020, opened 2023-05-31
docking: Do ControlsManagerLayout allocation with box-adjusted workArea

3v1n0:
We were allocating the ControlsManagerLayout using the whole work area geometry but then in the auto-hide mode we were reducing the size of the available area again, creating problems when the available space was reduced. Leading to non-clickable or activatable search results.

To prevent this, perform layout allocation using the available space even when the dock is in autohide mode.

Closes: #1612
LP: #1979096

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -2313,8 +2313,10 @@ var DockManager = class DashToDockDockManager {
                 const searchBox = this.overviewControls._searchEntry.get_allocation_box();
                 const { shouldShow: wsThumbnails } = this.overviewControls._thumbnailsBox;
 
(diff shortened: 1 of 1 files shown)


vanvugt:
Whatever the remaining bug is that this fixes, I can't reproduce it. The original bug seems fixed on `master` for me.

3v1n0:
It's not easy to reproduce, but I generally get it in a FullHD setup where non-fractional scaling at 200% is enabled + dock as autohide, and it does fixes #1979096 for good in this case (which previous fix was not enough for).

Issue #2010, opened 2023-05-16
Dash to dock broken, unable to rearrange icons


coreybruce:
Hey there I wanted to report a issue with dash to dock, I am unable to rearrange the icons using dash to dock and the only way  I was able to make it work was to set it to panel mode and also resizing icons doesn't not work unless it's set to panel mode. and have no idea why it's happening

I tried 
```
dconf reset -f /org/gnome/shell/extensions/dash-to-dock/
```

But it is still broken.

[Screencast from 2023-05-17 02-58-17.webm](https://github.com/micheleg/dash-to-dock/assets/8393780/8964f0fd-6324-4719-b746-73dc1d29c645)

I am using Manjaro Gnome using the latest version 81 and tested both the extension package from Manjaros repo and from Gnome extensions and have the same issue.

doppelhelix:
Same Problem here. the workaround is to disable "Move the applications button at the beginning of the dock"
That does the trick for me.

Fascomino:
When I moved the Show Applications icon to the end everything worked as intended.  This error was introduced with the last change to put the button at the edge when in panel mode.

ionutbortis:
Same here. No rearranging and the new opened apps that were not previously pinned to dash as favorites are not displayed properly.
[dash to dock 81 issues.webm](https://github.com/micheleg/dash-to-dock/assets/5013697/fe3e69e3-2412-4d44-ad68-4ede0c25f9c0)

coreybruce:
> Same Problem here. the workaround is to disable "Move the applications button at the beginning of the dock" That does the trick for me.

Hmm yeah that also does it but what is causing this issue I wonder?

coreybruce:
This is only happening on one machine, my laptop for example isn't sufferering from this same issue.

ionutbortis:
> > Same Problem here. the workaround is to disable "Move the applications button at the beginning of the dock" That does the trick for me.
> 
> Hmm yeah that also does it but what is causing this issue I wonder?

There are some recent changes that affect the dock behavior. I downgraded to version 80 and changed the version from the local `metadata.json` file to 81 in order to avoid future update to the problematic version.

coreybruce:
If anyone is using a Arch based system like me which uses Manjaro I have the backup package of the 80-1 package from Manjaro, I know this isn't recommended or trust worthy but I have the sig file to verify it

ionutbortis:
[...]

PR #2021, opened 2023-05-31
dash: Use a different container to hold floating ShowAppsIcon

3v1n0:
When using a floating show apps icon setup, it needs to be at the same level of the dash box, but we can't use the same container because upstream code only uses that for application icons, and this assumption may break things such as DnD and potentially other things.

So move things in a new container.

Closes: #2010


jcsekinger:
How do I « How do I « Use a different container to hold floating ShowAppsIcon»? This is significant for a dev but not for a simple user. Thank you !

PR #2036, opened 2023-06-07
Fixed the lack of app display on Gnome 41/42 

addieStarr:
Reverted 2 lines causing the regression.

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -851,8 +851,8 @@ var DockDash = GObject.registerClass({
         let newIndex = 0;
         let oldIndex = 0;
         while (newIndex < newApps.length || oldIndex < oldApps.length) {
-            const oldApp = oldApps.at(oldIndex);
-            const newApp = newApps.at(newIndex);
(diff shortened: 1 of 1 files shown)


dash.js
```diff
@@ -851,8 +851,8 @@ var DockDash = GObject.registerClass({
         let newIndex = 0;
         let oldIndex = 0;
         while (newIndex < newApps.length || oldIndex < oldApps.length) {
-            const oldApp = oldApps.at(oldIndex);
-            const newApp = newApps.at(newIndex);
+            const oldApp = oldApps.length > oldIndex ? oldApps[oldIndex] : null;
+            const newApp = newApps.length > newIndex ? newApps[newIndex] : null;
```
3v1n0, line 855: This should be fine as well:

```suggestion
            const oldApp = oldApps[oldIndex];
            const newApp = newApps[newIndex];
```

3v1n0: let's ignore this, it may still lead to warning in old gjs versions, not an issue, but let's keep the old verbose code :(

PR #2062, opened 2023-07-24
Update metadata.json to gnome 45

gbPagano:
just adding gnome 45 in metadata.json, tested and working.


3v1n0:
Not enough, 45 breaks all the extensions :)

https://gjs.guide/extensions/upgrading/gnome-shell-45.html#esm

Issue #2066, opened 2023-08-10
GNOME 45 Support


FeyXieXzf:
GNOME 45 is out, however basically every extension is not compatible with this version due to this error:
```
SyntaxError: import declarations may only appear at top level of a module
```

vanvugt:
Thanks for the bug report. We are aware of the problem: https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/6849

PR #2070, opened 2023-08-14
GNOME 45 support - ESM porting

3v1n0:
Use ESM imports as gnome-45 requires, so drop compatibility with all the
old versions so that we could modernize things a bit in future.

Closes: #2066

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -23,3 +23,6 @@ overrides:
         - code: 110
           ignoreUrls: true
           ignoreTemplateLiterals: true
+
+parserOptions:
+  sourceType: module
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ EXTRA_MODULES = \
                 appIcons.js \
                 appIconIndicators.js \
                 fileManager1API.js \
+                imports.js \
                 launcherAPI.js \
                 locations.js \
                 locationsWorker.js \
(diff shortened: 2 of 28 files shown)


3v1n0:
Let's get this in, will iterate new fixes if needed.

Augustiin:
Hi !
I'm trying to install dash to dock but I'm getting this :
`The settings of extension dash-to-dock@micxgx.gmail.com had an error:
```
SyntaxError: import declarations may only appear at top level of a module
```

Stack trace:
```
_init@resource:///org/gnome/Shell/Extensions/js/extensionPrefsDialog.js:23:33
ExtensionPrefsDialog@resource:///org/gnome/Shell/Extensions/js/extensionPrefsDialog.js:10:4
OpenExtensionPrefsAsync/<@resource:///org/gnome/Shell/Extensions/js/extensionsService.js:129:33
asyncCallback@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:115:22
run@resource:///org/gnome/Shell/Extensions/js/dbusService.js:186:20
main@resource:///org/gnome/Shell/Extensions/js/main.js:22:13
run@resource:///org/gnome/gjs/modules/script/package.js:206:19
start@resource:///org/gnome/gjs/modules/script/package.js:190:8
@/usr/share/gnome-shell/org.gnome.Shell.Extensions:1:17
```
`

PR #2091, opened 2023-09-11
appIcons: Use Mtk.Rectangle instead of Meta.Rectangle

stuarthayhurst:
Meta.Rectangle has been deprecated, fixes logs being flooded with:
`Meta.Rectangle is deprecated, use Mtk.Rectangle instead`


3v1n0:
Thanks!

PR #2093, opened 2023-09-24
Fix dashtodock.pot generation (avoid missing strings for translations)

xalt7x:
Some strings are not generated due to redundant option "-k" (leftover from commit 51db80abc3ec5c6c0621a0ba74f04919cb3f0b12)

In result most translations lack:
msgid "Show Details"
msgid "Launch using Integrated Graphics Card"
msgid "Launch using Discrete Graphics Card"
msgid "Failed to launch “%s”"


3v1n0:
Thanks, good spot!

PR #2132, opened 2023-12-12
Remove any dependency on dockmanager in desktopIconsIntegration.js

smedir:
Dockmanager can be null, and this spams the log if desktopIconsIntegration class is initialized and used.

````
Dec 11: JS ERROR: TypeError: DockManager.getDefault() is null
  get extension@file:///home/*/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmai>
  _sendMarginsToExtension@file:///home/*/.local/share/gnome-shell/extensions/dash-to-dock@m>
   _sendMarginsToAll/<@file:///home/*/.local/share/gnome-shell/extensions/dash-to-dock@micxg>
    _sendMarginsToAll@file:///home/*/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.>
   _changedMargins/this._timedMarginsID<@file:///home/*/.local/share/gnome-shell/extensions/>
````

desktopIconsIntegration class is initialized and called from the extension and really should not have any other dependencies other than the shell, it is meant to be used and can be used as is by any extension if called as is independently without modifications.

### DIFF
--- a/desktopIconsIntegration.js
+++ b/desktopIconsIntegration.js
@@ -55,16 +55,17 @@
  *
  *******************************************************************************/
 
-import {GLib} from './dependencies/gi.js';
-import {Main} from './dependencies/shell/ui.js';
-import {ExtensionUtils} from './dependencies/shell/misc.js';
-
-import {DockManager} from './docking.js';
+import GLib from 'gi://GLib';
+import * as Main from 'resource:///org/gnome/shell/ui/main.js';
+import * as ExtensionUtils from 'resource:///org/gnome/shell/misc/extensionUtils.js';
+import {Extension} from 'resource:///org/gnome/shell/extensions/extension.js';
 
 const IDENTIFIER_UUID = '130cbc66-235c-4bd6-8571-98d2d8bba5e2';
 
 export class DesktopIconsUsableAreaClass {
     constructor() {
+        const Me = Extension.lookupByURL(import.meta.url);
+        this._UUID = Me.uuid;
         this._extensionManager = Main.extensionManager;
         this._timedMarginsID = 0;
         this._margins = {};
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: desktopIconsIntegration.js [hunk 1]
```diff
@@ -55,16 +55,17 @@
...
 
-import {GLib} from './dependencies/gi.js';
-import {Main} from './dependencies/shell/ui.js';
-import {ExtensionUtils} from './dependencies/shell/misc.js';
-
-import {DockManager} from './docking.js';
+import GLib from 'gi://GLib';
+import * as Main from 'resource:///org/gnome/shell/ui/main.js';
+import * as ExtensionUtils from 'resource:///org/gnome/shell/misc/extensionUtils.js';
+import {Extension} from 'resource:///org/gnome/shell/extensions/extension.js';
```
3v1n0, line 61: I think you can still use our way to load deps, no?

smedir: Yes. Although that would beat the point of being the same independent file for all extensions. :smile:

In any case, added a commit to import dependencies your way.

3v1n0:
any reason why you closed this?

smedir:
It happened accidentally, I am not sure what happened. And then I could 
not figure out how to reopen it, it did not give a choice to reopen it. 
I am not too familiar with GitHub, on GitLab it is quite easy to 
re-open the MR.

But I did reply to your query, yes the dependencies can be used in your 
way. I did re-submit the MR as a new one with an added commit that does 
just that.

Although I do feel that the file should be quite independent of the 
underlying extension, and perhaps the dependencies and imports should 
not depend on the file structure/paths of the underlying extension.

Sundeep

PR #2140, opened 2024-01-05
Fix Polish translation of "Empty Trash"

pwiecz:
My understanding is that "Empty" here is a verb in imperative, not an adjective. The previous translation was correct only if "Empty" was as adjective.


bboozzoo (approved):
Looking at how https://github.com/micheleg/dash-to-dock/blob/ac7e31749b1755a9d7abf8ff9a4471b2c8fb9f82/locations.js#L736 is called, the change seems fine. LGTM

Issue #1722, opened 2022-04-23
Blurry focusing application


huyhoang8398:
![image](https://user-images.githubusercontent.com/25379882/164911637-204e9ac2-94d2-4f3a-8df2-2066ec7c074c.png)

The focusing application is blurry and have black square

Im using Ubuntu 22.04

zambelly:
I'm having the same issue on Arch
![image](https://user-images.githubusercontent.com/30311816/166680829-39355a73-c595-45d8-8f6c-fd49e7d6901d.png)

It's pretty inconsistent, sometimes it's fine, most of the time it's like this.
When hovering over the icon with the mouse, the issue disappears.
![image](https://user-images.githubusercontent.com/30311816/166680959-722dcb2f-b414-4ee8-aa67-2ee5c02dce52.png)

Anderson0xFF:
It happens to me sometimes, I don't know why.

delthia:
Hello,
If have just installed manjaro gnome, and I noticed the same thing happening. It does not happen on pinned applications, but only on opened applications that get added to the right of the panel.
I've tried in X11 and Wayland, and this does happen on either of them

![image](https://user-images.githubusercontent.com/101663529/175891443-36c2d655-401a-452b-8bf2-22d3f959ea44.png)

I've also just noticed that whenever the mouse cursor is hovering over the dock, it is showing correctly.
The behaviour is as follows for me:

1. When an application is oppened, it might show the glitch.
2. After opening other applications, if switching with `alt+tab` it is solved
3. If it is glitching hovering over the dock will solve the glitching
4. No matter if it was showing correctly or not, when taking the mouse cursor off of the dock, the glitch is shown until I switch applications with alt tab
The above described behaviour was tested with X11, this I haven't tested under wayland.

michaelgosling:
I've recently experienced this in both the latest Ubuntu and Manjaro. Maybe it's related to GNOME 42 and the new theming system? Just a thought.

delthia:
I don't remember if when I commented on the issue I was using gnome 41, in any case it is still happening (and more recently more frequently, I think).
If I might as (to see if it is related), have you experience a graphical glitch similar to this:

https://user-images.githubusercontent.com/101663529/189122616-cc3bd703-cc30-415a-aea0-fd5cfd9c045b.mp4

[...]

PR #2147, opened 2024-02-15
Fix for blurry background when opening a new application

sniirful:
When a new application is opened, the background usually results in being blurry, unless the mouse cursor hovers over the dash, just like in the following screenshots:
image
image

Such a problem was already well documented in the issue #1722.

This pull request fixes this issue by overriding the show() method from DashItemContainer in DockDashItemContainer and simulating a hover over the item once the animation is finished.

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -66,12 +66,11 @@ class DockDashItemContainer extends Dash.DashItemContainer {
         if (this.child == null)
             return;
 
-        let time = animate ? DASH_ANIMATION_TIME : 0;
         this.ease({
             scale_x: 1,
             scale_y: 1,
             opacity: 255,
-            duration: time,
+            duration: animate ? DASH_ANIMATION_TIME : 0,
             mode: Clutter.AnimationMode.EASE_OUT_QUAD,
             onComplete: () => {
                 // when the animation is ended, we simulate
--- a/dash.js
+++ b/dash.js
@@ -27,7 +27,10 @@ import {
     Utils,
 } from './imports.js';
 
-const {DASH_ANIMATION_TIME} = Dash;
+// module "Dash" does not export DASH_ANIMATION_TIME
+// so we just define it like it is defined in Dash;
+// taken from https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/dash.js
+const DASH_ANIMATION_TIME = 200;
 const DASH_VISIBILITY_TIMEOUT = 3;
 
 const Labels = Object.freeze({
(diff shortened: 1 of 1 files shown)


dash.js
```diff
@@ -54,6 +57,30 @@ class DockDashItemContainer extends Dash.DashItemContainer {
...
+    show(animate) {
+        if (this.child == null)
+            return;
+
+        let time = animate ? DASH_ANIMATION_TIME : 0;
+        this.ease({
+            scale_x: 1,
+            scale_y: 1,
+            opacity: 255,
+            duration: time,
```
3v1n0, line 74:
```suggestion
        this.ease({
            scale_x: 1,
            scale_y: 1,
            opacity: 255,
            duration: animate ? DASH_ANIMATION_TIME : 0,
```

3v1n0 (commented):
Thanks for looking into this, it's a long awaited thing!

The workaround looks ok, I'm still wondering if we can ever understand the root reason though :)

3v1n0:
What happens if the application is launched from another launcher or using the keyboard, is the forced-hover causing any side-effect?

sniirful:
> What happens if the application is launched from another launcher or using the keyboard, is the forced-hover causing any side-effect?

I've tested it using a Fedora 39 virtual machine and launching the app both by pressing the Super key and typing the app name, and by pressing Ctrl-Alt-T after setting it as a keyboard shortcut in the settings:
[Screencast from 2024-02-16 15-57-15.webm](https://github.com/micheleg/dash-to-dock/assets/45568396/40f26d85-9111-4076-acff-c242232de24a)
[Screencast from 2024-02-16 15-57-38.webm](https://github.com/micheleg/dash-to-dock/assets/45568396/0ca3ba42-3b2a-47f7-94fc-f9475999147f)

I apologize for the videos appearing with weird coloring (maybe that's a virtio bug?) but you can still clearly see that the animation has the background blurry and it unblurs only when the animation finishes. It's a bit annoying but it's not as annoying as having it always blurry.

> I'm still wondering if we can ever understand the root reason though :)

I'm not sure whether this issue can be attributed to how GNOME renders objects, it almost looks like it always keeps the same background resolution as the first frame it rendered. Since it renders the object by directly animating it, it looks like when it renders the first frame of the animation it doesn't bother to upscale it later.
This might seem just an assumption but it can be verified by two tests:

- Test n. 1:
Change the show function so that it first renders the object without animation, then makes it disappear and then renders the animation. The object will be sharp 100% of the time.
Result:
[Screencast from 2024-02-16 16-14-57.webm](https://github.com/micheleg/dash-to-dock/assets/45568396/74745672-cbf1-4e50-a42f-11cb4e780d7b)
Code:
```js
    show(animate) {
        if (this.child == null)
            return;

        let time = animate ? DASH_ANIMATION_TIME : 0;

        // make it appear first
        this.ease({
            scale_x: 1,
            scale_y: 1,
            opacity: 255,
            duration: 0,
        });
        
        // wait "time" ms and then make it disappear
        setTimeout(() => {
            this.ease({
                scale_x: 0,
                scale_y: 0,
                opacity: 0,
                duration: 0,
            });
        }, time);

        // wait "time * 2" ms and then make it animate
        setTimeout(() => {
            this.ease({
                scale_x: 1,
                scale_y: 1,
                opacity: 255,
                duration: time,
                mode: Clutter.AnimationMode.EASE_OUT_QUAD,
            });
        }, time * 2);
    }
```

- Test n. 2:
To prove it's not a matter of just waiting, let's change it so that instead of animating the object instantly, it waits let's say for 1 second and then it animates. The object will be blurry just like it was before this change.
Result:
[Screencast from 2024-02-16 16-20-09.webm](https://github.com/micheleg/dash-to-dock/assets/45568396/6579c383-f8fc-4b99-a9dd-fc245cbb54bc)
Code:
```js
    show(animate) {
        if (this.child == null)
            return;
        
        // wait 1 second and then animate
        setTimeout(() => {
            this.ease({
                scale_x: 1,
                scale_y: 1,
                opacity: 255,
                duration: animate ? DASH_ANIMATION_TIME : 0,
            });
        }, 1000);
    }
```

stuarthayhurst:
The root cause seems to have been identified here: https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/6567#note_2020532

The blurry issues seem to apply to the shutdown / reboot popup briefly, app icon backgrounds and desktop thumbnails.

Issue #2154, opened 2024-03-01
In multi-monitor setup dash dock could overlap window title tooltips in overview


taoky:
Versions:
- GNOME 45.4
- Dash to Dock 89

This is a bug similar to https://github.com/micheleg/dash-to-dock/issues/1608, but could trigger only in multi-monitor setup with a reasonable dock size:

![image](https://github.com/micheleg/dash-to-dock/assets/2109893/5263b69c-a714-4e02-b7cc-961df673ccf2)

Dock size settings:

![image](https://github.com/micheleg/dash-to-dock/assets/2109893/06c9fe41-f631-4bc9-9bb4-9dee9692242a)

Multi-monitor setup:

![image](https://github.com/micheleg/dash-to-dock/assets/2109893/093dc1dd-cd42-4153-97ee-28dc8d7cd438)

As far as I know, #1608 is resolved by a workaround and is hard to fix: without dash-to-dock, the dash and the `workspacesView` are in the same actor group, helping it to control which one is on the top of another. But the dock in dash-to-dock is outside of `overviewGroup` and has to been put above every widgets in it. Though I'm not sure if a similar workaround could be applied in multi-monitor case.

taoky:
A dirty patch of monkey-patching `WorkspacesView.WorkspacesView._getFirstFitSingleWorkspaceBox` in docking.js (when "Workspaces on all displays") and `WorkspacesView.ExtraWorkspaceView.prototype` (when "Workspaces on primary display only"):

```js
this._methodInjections.addWithLabel(Labels.MAIN_DASH, WorkspacesView.WorkspacesView.prototype,
    '_getFirstFitSingleWorkspaceBox', function (originalFunction, ...args) {
        const box = originalFunction.call(this, ...args);
        if (DockManager.settings.dockFixed)
            return box;
        const dock = DockManager.getDefault().getDockByMonitor(this._monitorIndex);
        if (!dock)
            return box;
        const [, preferredHeight] = dock.get_preferred_height(box.get_width());
        box.y2 -= preferredHeight;
        return box;
    });

this._vfuncInjections.addWithLabel(Labels.MAIN_DASH, WorkspacesView.ExtraWorkspaceView.prototype,
    'allocate', function (box) {
        if (DockManager.settings.dockFixed) {
            this.vfunc_allocate(box);
            return;
        }
        const dock = DockManager.getDefault().getDockByMonitor(this._monitorIndex);
        if (!dock) {
[...]

PR #2158, opened 2024-03-16
docking: fix window picker space for multi-monitor environments

taoky:
Attempted to fix https://github.com/micheleg/dash-to-dock/issues/2154.

This PR:
- Checks whether the dock is in primary monitor in `_computeWorkspacesBoxForState`. It looks like that this function only cares about the workspace on primary monitor, as original GNOME dock is in primary one.
- Adds a hook for `WorkspacesView.SecondaryMonitorDisplay._getWorkspacesBoxForState`. This function is similar to `_computeWorkspacesBoxForState`, and looks like the correct function to hook without glitches.

In `_getWorkspacesBoxForState`:

- `maybeAdjustBoxSize` is not reused as the workspace in secondary monitor does not have things like `searchBox`.
- Only bottom dock is handled, as it looks like nothing is wrong with docks positioned at left/right, and... I'm not sure how dock at top shall look like.
- `preferredHeight` might be too large here:

    image

    But there's no `this.spacing` inside `St.Widget` :(

And this only fixes the window picker part. A similar bug in multi-monitor environments, when "Workspaces on primary display only" is chosen, still exists when overview state is app grid -- but this could be triggered even without any extensions and thus a gnome-shell bug at present, though the space between dock and workspace is still too small.


3v1n0:
Oh, thanks for looking into this!

Let's see if #2166 causes this to break before merging

Issue #2110, opened 2023-10-20
JS ERROR: TypeError: this._prepareStartupAnimation() is undefined


stuarthayhurst:
Running Debian Experimental / GNOME 45
dash-to-dock version 89

Every time I log in, I get the following error message in the logs:
```
gnome-shell[2405]: JS ERROR: TypeError: this._prepareStartupAnimation() is undefined
                   _loadBackground/signalId</id<@resource:///org/gnome/shell/ui/layout.js:688:26
                   @resource:///org/gnome/shell/ui/init.js:21:20
```

No log message when the extension is disabled

dgutov:
Same here.

It doesn't crash the session or anything, but seems like we might be missing some graphical motion?

jtojnar:
For me, login gets stuck on a gray screen whenever this is logged during login.

Fortunately, it does not happen on every login. I see it more frequently on my laptop than my desktop (using the same system SSD) so it might be some kind of race.

dgutov:
>  login gets stuck on a gray screen whenever this is logged during login

This never happened to me so far, after several reboots/relogins (and journalctl has this error logged after my last restart, too, 10 minutes ago).

I've upgraded to GNOME 45 very recently, though, so maybe I'll see the grey screen some other day.

holmanb:
I see this as well. Ubuntu Noble (24.04) / GNOME Shell 45.3

rmeissn:
Seeing the issue as well on Gnome 45.4, dash to dock 89.
Journal entries which might be related, including the JS ERROR:
```
Mär 01 11:40:24 framework gnome-shell[57796]: Failed to open sliced image: Vorgang wurde abgebrochen
Mär 01 11:40:24 framework gnome-shell[57796]: Failed to open sliced image: Vorgang wurde abgebrochen
Mär 01 11:40:24 framework gnome-shell[57796]: Failed to open sliced image: Vorgang wurde abgebrochen
Mär 01 11:40:24 framework gnome-shell[57796]: Failed to open sliced image: Vorgang wurde abgebrochen
Mär 01 11:40:24 framework gnome-shell[57796]: Failed to open sliced image: Vorgang wurde abgebrochen
Mär 01 11:40:24 framework gnome-shell[57796]: Failed to open sliced image: Vorgang wurde abgebrochen
Mär 01 11:40:24 framework gnome-shell[57796]: cogl_framebuffer_set_viewport: assertion 'width > 0 && height > 0' failed

[...]

PR #2163, opened 2024-03-20
utils: Override async functions with async functions

vanvugt:
Otherwise gnome-shell (`LayoutManager._loadBackground`) can't find the override when it was expecting an async function.

Closes: https://github.com/micheleg/dash-to-dock/issues/2110

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -322,9 +322,16 @@ export class InjectionsHandler extends BasicHandler {
         if (!(original instanceof Function))
             throw new Error(`Virtual function ${name}() is not available for ${object}`);
 
-        object[name] = function (...args) {
-            return injectedFunction.call(this, original, ...args);
-        };
+        if (original.constructor.name === 'AsyncFunction') {
+            // eslint-disable-next-line require-await
+            object[name] = async function (...args) {
+                return injectedFunction.call(this, original, ...args);
+            };
+        } else {
+            object[name] = function (...args) {
+                return injectedFunction.call(this, original, ...args);
+            };
+        }
         return [object, name, original];
     }
 


utils.js
```diff
@@ -322,9 +322,16 @@ export class InjectionsHandler extends BasicHandler {
         if (!(original instanceof Function))
             throw new Error(`Virtual function ${name}() is not available for ${object}`);
 
-        object[name] = function (...args) {
-            return injectedFunction.call(this, original, ...args);
-        };
+        if (original.constructor.name === 'AsyncFunction') {
```
3v1n0, line 325: Oh, funny thing... I had this changed months ago and I forgot to commit...

However, define something like `const AsyncFunctionType = (async () => {}).constructor;` and then check for `original instanceof AsyncFunctionType`.

Also I feel we should check if the two things match, so if both the original ones are async or not and depending on that behave differently.

utils.js
```diff
@@ -322,9 +322,16 @@ export class InjectionsHandler extends BasicHandler {
...
+            // eslint-disable-next-line require-await
+            object[name] = async function (...args) {
+                return injectedFunction.call(this, original, ...args);
+            };
```
3v1n0, line 329: So, if you're declaring this `async` and `injectedFunction` that would be meant to await, so I think you can just drop the async and return the promise as it is no?

3v1n0 (changes requested):

utils.js
vanvugt: I saw that suggestion when googling but it felt like a hack so I intentionally didn't do it that way. It's uglier, more code, and not obviously correct to me.

Good point about checking two things match though...

Since it turns out the bug #2110 actually isn't important, and it's fixed by #2166, and you have another fix in mind, I suggest just merging your own fix or wait for #2166.

utils.js
vanvugt: I expect #2110 will be solved either by #2166 or by your own commit you said you had, or both. So we don't need to continue here.

PR #2164, opened 2024-03-20
Draft: Restore the missing login animation

vanvugt:
Restore the missing login animation, primarily for Ubuntu Dock which has disable-overview-on-startup.

https://bugs.launchpad.net/bugs/2058468

- [x] Use non-zero/undefined animation time.
- [x] Stop showing the dock before the animation starts.
- [ ] Prevent the panel from showing before the animation starts (or don't animate the panel at all).
- [ ] Disabling disable-overview-on-startup to re-enable the overview on startup has a broken animation.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -2269,7 +2269,6 @@ export class DockManager {
                     injections.destroy();
 
                     const onComplete = () => callback(...callbackArgs);
-                    dockManager._prepareStartupAnimation();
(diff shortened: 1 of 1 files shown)


vanvugt:
I think I can do better and rewrite for the new startup animations in GNOME 45/46, rather than continuously adding on...

vanvugt:
Moved to #2166

PR #2148, opened 2024-02-15
Fix dockmanager null

smedir:
Dockmanager can be null, and this spams the log if desktopIconsIntegration class is initialized and used.

Dec 11: JS ERROR: TypeError: DockManager.getDefault() is null
  get extension@file:///home/*/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmai>
  _sendMarginsToExtension@file:///home/*/.local/share/gnome-shell/extensions/dash-to-dock@m>
   _sendMarginsToAll/<@file:///home/*/.local/share/gnome-shell/extensions/dash-to-dock@micxg>
    _sendMarginsToAll@file:///home/*/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.>
   _changedMargins/this._timedMarginsID<@file:///home/*/.local/share/gnome-shell/extensions/>
   @resource:///org/gnome/shell/ui/init.js:21:20

desktopIconsIntegration class is initialized and called from the extension and really should not have any other dependencies other than the shell, it is meant to be used and can be used as is by any extension if called as is independently without modifications.

PR #2165, opened 2024-03-20
Remove any dependency on dockmanager in desktopIconsIntegration.js

smedir:
Dockmanager can be null, and this spams the log if desktopIconsIntegration class is initialized and used.

desktopIconsIntegration class is initialized and called from the extension and really should not have any other dependencies other than the shell, it is meant to be used and can be used as is by any extension if called as is independently without modifications.

Fix resource path for Extension

### DIFF
--- a/desktopIconsIntegration.js
+++ b/desktopIconsIntegration.js
@@ -55,16 +55,17 @@
  *
  *******************************************************************************/
 
-import {GLib} from './dependencies/gi.js';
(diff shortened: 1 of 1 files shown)


smedir:
Apologies, any force pushing a branch, does not let you re-open the pull request. Here is it again, please see #2148 and #2132.

I really need to learn how to navigate GitHub! More used to GitLab.

3v1n0:
Thanks!

PR #2166, opened 2024-03-21
Fix startup animation

vanvugt:
1. Make `STARTUP_ANIMATION_TIME` non-zero again.
2. Stop trying to fully replace the startup animation when we can just use the existing one upstream gnome-shell provides.
3. Slide the dock in when upstream's startup animation is done.

Closes: #2110 and https://bugs.launchpad.net/bugs/2058468

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -65,7 +65,6 @@ const Labels = Object.freeze({
     MAIN_DASH: Symbol('main-dash'),
     OLD_DASH_CHANGES: Symbol('old-dash-changes'),
     SETTINGS: Symbol('settings'),
-    STARTUP_ANIMATION: Symbol('startup-animation'),
     WORKSPACE_SWITCH_SCROLL: Symbol('workspace-switch-scroll'),
 });
 
@@ -2065,7 +2064,7 @@ export class DockManager {
         this.emit('docks-ready');
     }
 
-    _prepareStartupAnimation(callback) {
+    _prepareStartupAnimation() {
         DockManager.allDocks.forEach(dock => {
             const {dash} = dock;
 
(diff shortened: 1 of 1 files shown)


vanvugt:
Tested on both GNOME 46 and 45, which conveniently are the only versions we have to support now.

docking.js
```diff
@@ -2198,112 +2161,10 @@ export class DockManager {
...
-                            });
-
-                            connections.add(Utils.getMonitorManager(), 'monitors-changed', () => {
-                                connections.destroy();
-
-                                reject(new GLib.Error(Gio.IOErrorEnum,
-                                    Gio.IOErrorEnum.CANCELLED, 'Loading was cancelled'));
-                            });
-                        });
-                    };
```
3v1n0, line 2243: Ideally I'd be very happy to drop all this code, but one reason for that was that https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2763 was still a thing at the time.

I've not tested anymore the script, but that was triggering the same root issue we had that was leading to a black/non-responsive screen input area.

So, have you tested with tricks like those and see if we're strong enough now?

3v1n0 (commented):
Thanks, I'm happy to drop this code that we needed to address some issues, I'm wondering if upstream code is good enough to handle the various races we had before though...

docking.js
vanvugt: I don't remember all the details of those clipping bugs, only that it happened twice in the past. So we should investigate, but I am keen to build on a much smaller code base if any new fix is required. I vaguely recall finding upstream had the same bug, so that plus the fact that we no longer modify the upstream animation, should mean any fix required for the clipping issue could go upstream.

vanvugt: Sorry for the delay...

The clipping bug does still happen upstream (no extensions), but only for logging into `Main.sessionMode.hasOverview == false` which is not the default. And the clipping gets corrected at the end of the login animation.

dash-to-dock master does not have the bug, but to test that you also need to fix the undefined `STARTUP_ANIMATION_TIME` issue or else there's no way to slow down animations.

dash-to-dock #2166 does have the bug, but it now corrects itself at the end of the login animation just like upstream does.

So we no longer need to carry the clipping fix here. The bug is now the same with or without extensions after this PR. It should be treated as [an upstream bug](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/6473) from now on.

3v1n0: So I checked the script and the bug is back... mhmh

vanvugt: Yeah that's kind of what I said in the last two paragraphs. I think it's OK. It recovers at the end of the animation and [the same bug](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/6473) occurs without extensions so any fix should go in upstream gnome-shell.

vanvugt: Since it does recover at the end of the animation the bug shouldn't be visible at all when animations are disabled (like in the original VM case). But I forgot to test with animations disabled. I will need to do that.

Issue #2160, opened 2024-03-20
Dock indicators appear on top of icons in GNOME 46


vanvugt:
Dot indicators appear on top of icons in GNOME 46:

![Screenshot from 2024-03-19 15-42-55](https://github.com/micheleg/dash-to-dock/assets/11951986/722c7879-d11f-4edd-a398-de34a063175a)

a1ex-ak:
if you set "Use built-in theme" in the settings, then the point returns to its place.
![Снимок экрана от 2024-03-20 13-19-08](https://github.com/micheleg/dash-to-dock/assets/121604147/d79728c0-ff49-4d23-ad33-354eaf7ad6ba)

garrett:
If you don't choose "built-in theme", but change from "Default" to "Dots", it gets less wrong.

Here's built-in off with "default" on:

![image](https://github.com/micheleg/dash-to-dock/assets/10246/8811dfc1-dd56-48b1-9ca2-8bf0a544153b)


![image](https://github.com/micheleg/dash-to-dock/assets/10246/57ee0f8c-8e40-4945-94b1-935bff8a420c)

And with "dots on:

![image](https://github.com/micheleg/dash-to-dock/assets/10246/269feecf-b29c-458e-823b-bacb27a192df)

![image](https://github.com/micheleg/dash-to-dock/assets/10246/d59e1edf-0a8e-468f-84a3-da5353796f6e)

Dots is basically the same as built-in theme. Here's built-in theme as a comparison:

![image](https://github.com/micheleg/dash-to-dock/assets/10246/40897595-b7e7-487e-9c2c-37c20585ecdb)

The dots look fine on most icons, but are incorrect on chunkier icons like all the chat apps. In other words, I'd say it's always wrong (as the indicator shouldn't touch the icons), but it's less wrong with dots or built-in.

vanvugt:
Judging from those screenshots, I think my first one at the top of the page looks worst (most overlapped) because I'm using display scale 200%. So https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/7488 is a little bit of a factor here. I guess we can avoid getting confused with that bug by testing with display scale 100% first.

Calosis:
[...]

PR #2168, opened 2024-03-22
Fix dot height in default mode

sergio-costas:
Fix https://github.com/micheleg/dash-to-dock/issues/2160

### DIFF
--- a/_stylesheet.scss
+++ b/_stylesheet.scss
@@ -20,6 +20,7 @@ $dash_background_color: lighten($system_bg_color, 5%);
 $dash_placeholder_size: 32px;
 $dash_padding: $base_padding*2; // 12px
 $dash_border_radius: $modal_radius + $dash_padding;
+$dash_padding_gnome: $dash_padding;
 
 // Dash to dock specifics
 $dash_padding: $base_padding + 4px; // 10px
@@ -259,6 +260,10 @@ $dock_style_modes: [null, shrink, extended, extended-shrink];
 .dash-item-container .overview-tile.focused .overview-icon,
 .dash-item-container .show-apps:hover .overview-icon {
     background-color: $remark_color;
+}
+
+.dash-item-container .overview-tile .app-grid-running-dot {
+    margin-bottom: $dash_padding_gnome - 4px;
 }
 
 @each $side in $dock_sides {


sergio-costas:
@vanvugt Fixed the dot height in Gnome 46, and confirmed that it doesn't break in Gnome 45.

vanvugt:
Thanks. Testing it I see the dot is still too close to the icon. And reverting to scale 100% doesn't totally fix that.

Can you check to see how upstream does their spacing?

vanvugt:
Maybe fixing https://github.com/micheleg/dash-to-dock/issues/2173 would change things? Or maybe it's not related.

sergio-costas:
I see the dot exactly in the same place than in gnome 45...

Gnome45:
Captura desde 2024-03-25 16-55-50

Gnome46:
imagen

sergio-costas:
But, of course, I don't use scaling... At which scale are you using your desktop? How can I take into account scaling in CSS?

3v1n0:
> How can I take into account scaling in CSS?

Css should handle it automatically.

You can simulate scaling in a nested GNOME shell easily though. Launches it with something like `MUTTER_DEBUG_DUMMY_MONITOR_SCALES=1` and  `MUTTER_DEBUG_DUMMY_MODE_SPECS="1920x1080:3840x2160:4096x2160:7680x4320:3840x2400:3200x1800:2650x1440:1024x768:800x600:1440x900:1680x1050:1680x1012"`

vanvugt:
> I see the dot exactly in the same place than in gnome 45...
> 
> Gnome45: Captura desde 2024-03-25 16-55-50
> 
> Gnome46: imagen

That's not a good example because they are different icons (one's not square). Also the dot is touching the bottom in the second screenshot, but not the first. But all this may be off-topic so I'll take new screenshots...

vanvugt:
OK my main concern is only visible at 200% today:

| gnome-shell | #2168 |
| --- | --- |  
| scale200-gnome-shell | scale200-pr2168 |

It's minor and maybe can be ignored while we wait for https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3245

_stylesheet.scss
```diff
@@ -261,6 +262,10 @@ $dock_style_modes: [null, shrink, extended, extended-shrink];
     background-color: $remark_color;
 }
 
+.dash-item-container .overview-tile .app-grid-running-dot {
+    margin-bottom: $dash_padding_gnome - 3px;
```
vanvugt, line 266: Changing 3px to 4px resolves my concern about the slightly insufficient spacing. With 4px it now looks the same as upstream gnome-shell. Would 4px work for you?

sergio-costas: No problem. I'll do the change.

sergio-costas: Done.

vanvugt: Thanks. The change should have been squashed into the first commit so I force-pushed the squash into your branch. I will now merge it...

Issue #1608, opened 2021-12-04
Gnome 41: The dash (dock) sits above the overview, meaning it obscures things like window titles.


Bananaman:
This issue happens in the Gnome Overview / Activities view.

- Without Dash to Dock, all window titles show above the dock.
- With Dash to Dock, the titles are hidden behind the dock.

Here's an image of the issue:

![image](https://user-images.githubusercontent.com/38923130/144714671-3f7d8e71-d63b-4e02-9450-b4121b294e75.png)

dnut:
Ideally there would be a margin between the dock and the windows so nothing has to overlap.

Bananaman:
@dnut I guess so but that would be more invasive. I just hope the core functionality can be restored.

I suspect that the cause is something like "Dash to Dock" modifies the Dash CSS to "always on top" via a high z-index.

The fix would be something like modifying the window label CSS to "always on top + 1". :)

somepaulo:
> Ideally there would be a margin between the dock and the windows so nothing has to overlap.

In vanilla Gnome there's no margin, but whatever window you hover over is brought to the top and things appear above the dash, overlapping it. When you hover over the dash again, you're no longer hovering over a window, so the window name disappears, and there's no ugly overlapping.

Bananaman:
I don't understand the fixes (it's adding padding between the window selector and the dock?), but I can confirm that the latest Dash 2 Dock master build as of today + GNOME 42.1 = Looks Perfect. Thank you! @3v1n0 :heart:

subnut:
Nope. Still happens.

![Image](https://github.com/user-attachments/assets/faa63b58-9aae-494b-bcee-1f02fffa3ccf)

 
GNOME 47  
Wayland

PR #2172, opened 2024-03-25
docking: adapt ControlsManagerLayout spacing for GNOME 46

taoky:
In https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3104, the base class of `ControlsManagerLayout` is changed from `Clutter.BoxLayout` to `Clutter.LayoutManager`, and `spacing` is replaced by `_spacing`. This makes the fix in https://github.com/micheleg/dash-to-dock/issues/1608 no longer works in GNOME 46.


3v1n0:
Thanks!

Issue #1381, opened 2021-02-01
Bug: Dock flickers on top of Chrome window


balogandras:
Hi! I'm experiencing some flickering:

https://photos.app.goo.gl/pFcSkBp2fAXtKDL16

Only
- on the second monitor,
- when dock auto hide is enabled,
- when Chrome window is maximized
- and active.
- on OpenSUSE Tumbleweed, GNOME 3.38.3, Wayland.

It worked well with OpenSUSE Leap 15.2, GNOME 3.34

Thanks for the extension. I like it this way since Unity.

HanyuCheng:
I met the same issue on arch with Kernel: x86_64 Linux 5.10.13-arch1-2, GNOME 3.38.3, Wayland, Dash to Dock v69.
 -  when auto hide is enabled,
- when Chrome window is maximized
- and active.

eggheroesgrave:
I had this same pedicament and upon searching for a solution, I found a temporary fix by turning off hardware accelaration for Chrome.
This can be done by simply heading over to "Settings> Advanced: System> Hardware Accelaration" and toggling it off.

Unfortunately this is probably a Chrome specific fix and will also result in a slightly lower UX.

I'm on Ubuntu 21.04 with Wayland, GNOME Shell 3.38.4 and Dash to Dock v69.

speakradmin:
Same issue here with Chrome but also getting it with Microsoft Teams. I can confirm disabling hardware acceleration on Teams works same as Chrome. Could be Electron related?

Dorro101:
Same issue here with Firefox 102.0, Fedora 36, Gnome 42.2 on Wayland, Lenovo ThinkPad X1 Tablet Gen 3, Intel® Core™ i5-8250U, Mesa Intel® UHD Graphics 620...

x3rAx:
Same issue happens to me in the following apps:

- Firefox 106.0.1
- VScode 1.69.2
- MS Teams 1.5.00.23861
- Obsidian 1.0.3

I'm on:

- OS: NixOS 22.05
- Wayland
- [Framework Laptop (12th Gen Intel® Core™)](https://frame.work/de/en/products/laptop-diy-12-gen-intel)
- no dedicated graphics card
- display connected through HDMI

Disabling hardware acceleration in all of these apps did nothing (yes I restarted them :wink:). Pushing the cursor against the screen edge as @Dorro101 suggests does not make a difference for me.

[...]

PR #2184, opened 2024-03-28
intellihide: treat fullscreen windows as maximized

JimmyJichi:
I can't think of any reason to NOT hide the dock when the window is in fullscreen mode. In my opinion that defeats the purpose of "dodge maximized windows".

### DIFF
--- a/intellihide.js
+++ b/intellihide.js
@@ -283,7 +283,7 @@ export class Intellihide {
 
         case IntellihideMode.MAXIMIZED_WINDOWS:
             // Skip unmaximized windows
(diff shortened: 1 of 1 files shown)


vanvugt:
@taoky, do you think this could be related to your recent work?

taoky:
> @taoky, do you think this could be related to your recent work?

I don't think so, as I could reproduce the related problem (https://github.com/micheleg/dash-to-dock/issues/1381#issuecomment-2025546513) with 554f7c5541bc4ff05d89a9419d2580d3780595e3.

3v1n0:
Looks reasonable though.

Issue #2159, opened 2024-03-20
Gjs-WARNING: The GObject property St.ScrollView.[hv]scroll is deprecated (in GNOME 46)


vanvugt:
In GNOME 46:

```
(gnome-shell:18994): Gjs-WARNING **: 13:53:18.466: The GObject property St.ScrollView.vscroll is deprecated.
0 ensureActorVisibleInScrollView() ["file:///home/dan/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/dash.js":1143:38]
1 _createAppItem/<() ["file:///home/dan/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/dash.js":569:46]
2 _generateAccessors/propdesc.set() ["resource:///org/gnome/gjs/modules/core/_common.js":38:25]
3 _updateFocusState() ["file:///home/dan/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js":328:8]
4 _init/<() ["file:///home/dan/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIcons.js":921:74]
5 popModal() ["resource:///org/gnome/shell/ui/main.js":739:16]
6 _syncGrab() ["resource:///org/gnome/shell/ui/overview.js":526:21]
7 _hideDone() ["resource:///org/gnome/shell/ui/overview.js":647:13]
8 _animateNotVisible/<() ["resource:///org/gnome/shell/ui/overview.js":627:54]
9 onStopped() ["resource:///org/gnome/shell/ui/overviewControls.js":757:20]
10 _makeEaseCallback/<() ["resource:///org/gnome/shell/ui/environment.js":65:21]
11 _easeActorProperty/<() ["resource:///org/gnome/shell/ui/environment.js":232:59]
12 anonymous() ["resource:///org/gnome/shell/ui/init.js":21:19]


(gnome-shell:18994): Gjs-WARNING **: 13:53:18.466: The GObject property St.ScrollView.hscroll is deprecated.
0 ensureActorVisibleInScrollView() ["file:///home/dan/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/dash.js":1144:38]
1 _createAppItem/<() ["file:///home/dan/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/dash.js":569:46]
[...]

PR #2185, opened 2024-03-28
fix deprecated "St.ScrollView.[hv]scroll" property

K1ngjulien:
no more deprecation warnings on my machine :)

see: #2159

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -1140,10 +1140,8 @@ export const DockDash = GObject.registerClass({
  * @param actor
  */
 function ensureActorVisibleInScrollView(scrollView, actor) {
-    const {adjustment: vAdjustment} = scrollView.vscroll;
-    const {adjustment: hAdjustment} = scrollView.hscroll;
-    const {value: vValue0, pageSize: vPageSize, upper: vUpper} = vAdjustment;
-    const {value: hValue0, pageSize: hPageSize, upper: hUpper} = hAdjustment;
+    const {value: vValue0, pageSize: vPageSize, upper: vUpper} = scrollView.vadjustment;
+    const {value: hValue0, pageSize: hPageSize, upper: hUpper} = scrollView.hadjustment;
     let [hValue, vValue] = [hValue0, vValue0];
     let vOffset = 0;
     let hOffset = 0;


K1ngjulien:
no clue what the docstring above the function is supposed to mean tho. I couldn't find anything similar in utils.js.
any suggestions or should i just leave it?

K1ngjulien:
ah damn i'm not sure this is backwards compatible with gnome 45. no clue how to do that :sweat_smile:

K1ngjulien:
its just JavaScript right? can i just check if `scrollView.hadjustment === undefined` and then fall back to the old way of accessing it?

or is there a better way of doing things?
(sorry, I'm pretty new to writing gnome extensions)

dash.js
```diff
@@ -1140,8 +1140,15 @@ export const DockDash = GObject.registerClass({
  * @param actor
  */
 function ensureActorVisibleInScrollView(scrollView, actor) {
-    const {adjustment: vAdjustment} = scrollView.vscroll;
-    const {adjustment: hAdjustment} = scrollView.hscroll;
+    let vAdjustment = scrollView.vadjustment
```
sergio-costas, line 1143: What about something like...?

`const vAdjustment = scrollView.vadjustment ?? scrollView.vscroll.adjustment`

sergio-costas: (and maybe add a comment before explaining that scrollView.vscroll/hscroll has been deprecated in Gnome Shell 46)

K1ngjulien: ooh good point! i forgot that this operator exist in js :)

sergio-costas: I discovered it some months ago, and I really love it :-)

K1ngjulien: updated it :+1:

K1ngjulien:
should be fixed now

sergio-costas:
Thanks!!!!

K1ngjulien:
you're welcome :)
sooo ready for gnome 46 release now? :P

Aqua1ung:
> sooo ready for gnome 46 release now? :P

C'mon guys, put it up there already!

PR #2207, opened 2024-04-24
Fix window preview in Gnome 45

### LINKED ISSUE #2202: Window preview doesn't work anymore
Hi, I have just updated to version 91 and I had configured the middle button of the mouse to cause an application to pop up a preview of its content but it does no longer work (neither with the window focused, minimized nor any other window status)
previewNotWorking

sergio-costas:
When setting the middle button click as "show window preview", the previews aren't shown. Instead, an empty rectangle appears and a lot of errors are shown in the log.

This patch fixes this.

Fix #2202

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -197,10 +197,7 @@ export const DockDash = GObject.registerClass({
         });
         this._box._delegate = this;
         this._boxContainer.add_child(this._box);
-        if (this._scrollView.add_actor)
-            this._scrollView.add_actor(this._boxContainer);
-        else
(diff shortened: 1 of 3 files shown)


3v1n0 (approved):
Looks good to me, @JoseExposito maybe you can check again the new version still works for you?

JoseExposito:
@3v1n0 tested after the refactor, keeps working.

Thank you both of you for your work on the extension!

3v1n0:
Thanks both!

PR #2217, opened 2024-05-05
locations: Ask again for password on encryption failures

3v1n0:
This sadly requires parsing the error strings since there are not good APIs yet, but that's how upstream handles this too, so let's accept it for now.

See: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3300


vanvugt:
Seems to run without error (providing you use GNOME 46 and not 47 (https://github.com/micheleg/dash-to-dock/pull/2223)).

So long as it's given time to mature on master and **isn't** released soon, I'm happy to merge it.

Issue #2216, opened 2024-05-05
Desktop crash if you try to lock your screen from overview


J0hn-C77n:
Hello!

I noticed that I have some issues with my system when I use dash to dock. 

first thing is that on my 2k screen there's a broken default color scheme
![image](https://github.com/micheleg/dash-to-dock/assets/154112993/39dba329-1551-4acc-ad81-92aef0028eeb)

while on my 1080p everything is OK
![image](https://github.com/micheleg/dash-to-dock/assets/154112993/9c7ea565-b9da-4bce-8cb1-6bdd452336e2)

also if you press super + L from this view, you'll get a gnome crash screen and you'll be forced to logout 
![image](https://github.com/micheleg/dash-to-dock/assets/154112993/2b0b1901-1c71-4966-8da3-90e496beaacd)

Here's my system specs:

## Hardware Information:
- **Hardware Model:**                              Lenovo ThinkBook 16 G4+ IAP
- **Memory:**                                      16.0 GiB
- **Processor:**                                   12th Gen Intel® Core™ i9-12900H × 20
- **Graphics:**                                    Intel® Graphics (ADL GT2)
- **Graphics 1:**                                  NVIDIA GeForce RTX™ 2050
- **Disk Capacity:**                               1.5 TB

## Software Information:
- **Firmware Version:**                            HYCN48WW
- **OS Name:**                                     Debian GNU/Linux trixie/sid
- **OS Build:**                                    (null)
- **OS Type:**                                     64-bit
- **GNOME Version:**                               46
- **Windowing System:**                            X11
- **Kernel Version:**                              Linux 6.7.12-amd64

And also extensions that I use
![image](https://github.com/micheleg/dash-to-dock/assets/154112993/70997ab5-6332-4384-bab8-218a7bb2300b)

if you need it

MarvvanPal:
Regarding the crash see #2214

sergio-costas:
I've been able to reproduce the crash when locking the system while in the overview mode. I wrote a workaround.

vanvugt:
Duplicate of #2179

Issue #2179, opened 2024-03-27
GNOME shell crashes when dash-to-dock extension is disabled in overview


taoky:
With current code from master (https://github.com/micheleg/dash-to-dock/commit/1053b3e336780d0c9a51cca46034d48463b5c2bf), dash-to-dock would make shell crash in both GNOME 45 and 46. This could be reproduced by:

1. Open Overview (both window picker and app grid could trigger this)
2. Lock the screen, or run `gnome-extensions disable dash-to-dock@micxgx.gmail.com` if `ssh`ing.

My local bisecting result shows that https://github.com/micheleg/dash-to-dock/commit/de35ab608daa8368d73397d52783507fa7edd15b is the first bad commit (weird) when testing with GNOME 45. And this diff patch "workarounds" the crash in both GNOME 45 and 46:

```diff
diff --git a/_stylesheet.scss b/_stylesheet.scss
index b23cc43..adc45af 100644
--- a/_stylesheet.scss
+++ b/_stylesheet.scss
@@ -251,10 +251,6 @@ $dock_style_modes: [null, shrink, extended, extended-shrink];
     }
 }
 
-.dash-item-container .overview-tile .overview-icon,
-.dash-item-container .show-apps .overview-icon {
-    background-color: rgba(255,255,255,0);
-}
 
 .dash-item-container .overview-tile:hover .overview-icon,
 .dash-item-container .overview-tile.focused .overview-icon,
```

This issue looks like https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/7339 but it's slightly different (crashing widget is `Gjs_ui_dash_ShowAppsIcon`), and I have tried to find the cause but no clue. And this bug, if exists in shell, could only be triggered by an extension. I have also tried to build a simpler extension example, but it does not crash with a CSS-only extension.

bt full (GNOME 46): 
[bt-full-fedora-gnome46.txt](https://github.com/micheleg/dash-to-dock/files/14772277/bt-full-fedora-gnome46.txt)

vanvugt:
Thanks for the bug report. dash-to-dock shouldn't be able to cause a segfault like that, and particularly not with style changes. It looks like we should track this [upstream](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/7339) just keeping in mind the only way to reproduce it is with dash-to-dock.

vanvugt:
[...]

PR #2219, opened 2024-05-07
Hide overview before locking screen

sergio-costas:
There seems to be a bug in St/Clutter that crashes Gnome Shell if Dash to Dock is enabled and the user locks the screen from the overview mode.

This patch is a workaround in the meantime, while the true bug is found, to avoid this. It exits the overview mode before locking the screen.

Fix https://github.com/micheleg/dash-to-dock/issues/2216

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -1662,10 +1662,42 @@ const WorkspaceIsolation = class DashToDockWorkspaceIsolation {
 
 
 export class DockManager {
+    _restoreLock() {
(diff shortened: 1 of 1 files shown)


sergio-costas:
@3v1n0 Can you review this when you have some spare time, please?

vanvugt:
Sounds like it was https://github.com/micheleg/dash-to-dock/issues/2179, for which the upstream fix is https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3252

PR #2221, opened 2024-05-13
Fix communication with DING (same patch for "ubuntu-dock" branch)

sergio-costas:
This patch is the same patch from master, but for merging in "ubuntu-dock"

The extension state naming has changed from gnome shell 45 to gnome shell 46, so the code to notify margins to DING wasn't being able to detect when an extension was active, and so it didn't prevent to put icons below the dock.

This patch fixes it.

### DIFF
--- a/desktopIconsIntegration.js
+++ b/desktopIconsIntegration.js
@@ -63,6 +63,11 @@ import {Extension} from 'resource:///org/gnome/shell/extensions/extension.js';
 const IDENTIFIER_UUID = '130cbc66-235c-4bd6-8571-98d2d8bba5e2';
 
 export class DesktopIconsUsableAreaClass {
+    _checkIfExtensionIsEnabled(extension) {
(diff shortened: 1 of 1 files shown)


sergio-costas:
Coffee hasn't kicked in yet, sorry...

sergio-costas:
@3v1n0 @vanvugt Can you review this and https://github.com/micheleg/dash-to-dock/pull/2220 when you have some spare time, please? In my opinion, this should be added for 24.04.1

vanvugt (approved):

vanvugt:
Merged, thanks.

> In my opinion, this should be added for 24.04.1

It will need an LP bug for that.

sergio-costas:
@vanvugt Here it is the LP https://bugs.launchpad.net/ubuntu/+source/gnome-shell-extension-desktop-icons-ng/+bug/2068882

PR #2236, opened 2024-06-24
Merge last changes in master

sergio-costas:
This patch adds the last changes currently at upstream/master.


sergio-costas:
@3v1n0 Marco, can you review this when you have some spare time, please?

3v1n0:
Better to use a merge commit for this, not to miss the history, but I'll do that when we've to go through SRU or oracular releases

PR #2237, opened 2024-06-24
Don't show error messages when disabling

sergio-costas:
Since the destroy() function can be called several times, it is paramount to don't call objects that have been freed in previous calls.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -2551,7 +2551,7 @@ export class DockManager {
         this._appSwitcherSettings = null;
         this._oldDash = null;
 
-        this._desktopIconsUsableArea.destroy();
+        this._desktopIconsUsableArea?.destroy();
         this._desktopIconsUsableArea = null;
         this._extension = null;
         DockManager._singleton = null;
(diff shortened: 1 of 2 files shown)


sergio-costas:
@3v1n0 And this one too, please?

vanvugt:
What are the error messages? Have they been reported as bugs in Launchpad or here?

sergio-costas:
Mmm... odd... now I can't reproduce them... But they were messages that said that `undefined` hasn't a `destroy()` method.

sergio-costas:
Anyway, if you check the lines 2535 to 2540, before calling `destroy()` it checks if the variable is NULL or not, because it is being set to NULL right after calling the method.

sergio-costas:
And it does my way in other lines of the same method too:
imagen

vanvugt (approved):

3v1n0:
Thanks, in theory we should not call it multiple times, but it's not a problem to be protected by that.

PR #2240, opened 2024-06-26
UDENG-3111: allow to disable an icon in the dock during updates

sergio-costas:
For the new RAA we need to be able to mark an icon as "disabled" while it is being updated. The current Unity LauncherAPI (https://wiki.ubuntu.com/Unity/LauncherAPI) doesn't support this, so this patch adds an extra option for this.

https://docs.google.com/document/d/1--DgBRl6AqNiyjW_luOjl1dDzsIZPlF0Ukc7INW9_XQ

### DIFF
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -740,6 +740,10 @@ class UnityIndicator extends IndicatorBase {
             'urgent-changed',
             (sender, {urgent}) => this.setUrgent(urgent),
         ], [
+            remoteEntry,
+            'updating-changed',
+            (sender, {updating}) => this.setUpdating(updating),
+        ], [
             notificationsMonitor,
             'changed',
             () => this._updateNotificationsCount(),
@@ -759,6 +763,7 @@ class UnityIndicator extends IndicatorBase {
         this._notificationBadgeBin = null;
         this._hideProgressOverlay();
         this.setUrgent(false);
+        this.setUpdating(false);
         this._remoteEntry = null;
 
         super.destroy();
@@ -1017,6 +1022,16 @@ class UnityIndicator extends IndicatorBase {
             this._isUrgent = urgent;
         else
             delete this._isUrgent;
+    }
+
+    setUpdating(updating) {
+        if (updating || this._isUpdating !== undefined)
+            this._source.updating = updating;
+
+        if (updating)
+            this._isUpdating = updating;
+        else
+            delete this._isUpdating;
     }
 }
 
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -1025,13 +1025,7 @@ class UnityIndicator extends IndicatorBase {
     }
 
     setUpdating(updating) {
-        if (updating || this._isUpdating !== undefined)
-            this._source.updating = updating;
-
-        if (updating)
-            this._isUpdating = updating;
-        else
-            delete this._isUpdating;
+        this._source.updating = updating;
     }
 }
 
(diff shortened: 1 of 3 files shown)


sergio-costas:
@3v1n0 @vanvugt Can you review this when you have some spare time, please?

appIconIndicators.js
```diff
@@ -1018,6 +1023,16 @@ class UnityIndicator extends IndicatorBase {
         else
             delete this._isUrgent;
     }
+
+    setUpdating(updating) {
+        if (updating || this._isUpdating !== undefined)
+            this._source.updating = updating;
```
vanvugt, line 1029: This feels like an unclear way of ever setting `.updating` to false. Can you rewrite the statement to make it more clear that it works the same for both true and false?

sergio-costas: I fully agree in that this piece of code is very convoluted, but I just copied the code from `setUrgent(urgent)`, which is the method placed just before this one. I presumed that if the original author used that way, was for some reason. It seems that it allows to set the source variable to TRUE as many times as a DBus message is received, but set it to FALSE only once.

What do you think about something like:

```
    setUpdating(updating) {
        if (updating || this._source.updating)
            this._source.updating = updating;
    }
```

Another possibility is to define `this._isUpdating` as FALSE in the constructor, compare here with it, and update its value after the comparison...

vanvugt: Technically you don't need to store any member variables at all because the boolean you need is already a parameter in the notification. So you could remove the whole `setUpdating` function and the member variables :)

sergio-costas: Ok, I think that I now why it was made this way with `setUrgent()`: when it is enabled, it sets a property in a list of "interesting windows", so every time it is enabled it must rebuild that window list; but if it is disabled it only has to remove the animation. And since it doesn't check in `appIcons.js` if it was already disabled, it does that here.

Since nothing of that is needed with `setUpdating()`, I'll change it to the simplest one.

appIcons.js
```diff
@@ -187,6 +191,13 @@ const DockAbstractAppIcon = GObject.registerClass({
             }
         });
 
+        this.connect('notify::updating', () => {
+            const icon = this.icon._iconBin;
+            if (this.updating)
+                icon.set_opacity(128);
+            else
+                icon.set_opacity(255);
```
vanvugt, line 199: I'm wondering if these would be better as `set_opacity_override`, so as to prevent losing any previous custom opacity value. But I won't block on that -- it would only hurt the dock itself if such a loss ever happens so we'll deal with that later.

appIcons.js
```diff
@@ -187,6 +191,13 @@ const DockAbstractAppIcon = GObject.registerClass({
             }
         });
 
+        this.connect('notify::updating', () => {
+            const icon = this.icon._iconBin;
```
3v1n0, line 195: For next time, I think it would have been nicer to use `this.bind_property_full`.

However, I feel this code should be better to be as part of `AppIconIndicators.IndicatorBase`, no?

Since the `appIcon` is only the logical representation of an icon, shouldn't touch its look.

sergio-costas: Now that I see the code with more detail, I don't think so: this code is in DockAbstractAppIcon, from which DockAppIcon and DockLocationAppIcon. Maybe it can be moved to DockAppIcon, but surely it has to be there, because it is what manages the appearance of the icon itself. IndicatorBase only manages the things that are over the icon. In fact, IndicatorBase manages the "updating" notification, and relays it to the corresponding DockAbstractAppIcon, which, I think, is the right way.

PR #2241, opened 2024-06-26
UDENG-3111: allow to disable an icon in the dock during updates

sergio-costas:
This is the same PR than #2240, but to merge in ubuntu-dock branch.


vanvugt:
It would avoid confusion and duplication of effort to only propose one branch at a time. So I'll close this temporarily and please see #2240 for additional comments.

sergio-costas:
Good point. Agree.

Issue #2193, opened 2024-04-11
Short-lived programs cause the dock to bounce


starlightsys:
Every time Neovim interacts with the clipboard, the dock is animated for a brief moment as it responds to `wl-copy`/`wl-paste` being launched and closed.

[Screencast from 2024-04-11 22-44-22.webm](https://github.com/micheleg/dash-to-dock/assets/2768754/c3f12387-a31b-4906-a167-ed6c15a42950)

It is agonizingly distracting to have those movements happen all the time while editing files, and I can't use Dash-to-Dock as it is currently together with Neovim or other programs that similarly run short-lived programs like that.

I would imagine this bug could be fixed by implementing #988 or #2064, or by having a small delay between program launch and the program appearing in the dock.

bedohave:
I notice that this jiggle in the dock coincides with a jiggle in the Activities window (when I have at least 3 open windows). So this may or may not be a fault in Dash-to-Dock. 

I am running Fedora 40 on AMD CPU/GPU.

https://github.com/micheleg/dash-to-dock/assets/4650839/92f989e7-6fed-4806-9dba-b1d5645bb51a

Issue #2064, opened 2023-08-03
Feature Request: Hiding Programs From The Dock


Joey451-OG:
Hello, I wonder if it would be possible to make a setting which stops specified programs from showing on the dock. I have some programs that I want running in the background at all times. I would like to hide these programs so they do not clutter my limited screen space on my laptop.

Issue #988, opened 2019-09-20
Allow hiding of specific apps in dock


rschmied:
Would be cool if I could specify that certain apps should not show in the dock. I'm running the "desktop icons" Gnome extension and that launches an app / process 'gjs' that can not be quit from the dock and only shows the generic question mark icon. It just uses space and should be hidden.

PR #2244, opened 2024-06-29
Fix wl-clipboard messing with the dock

ndr3www:
A temporary fix of `wl-clipboard` appearing in the running apps section (https://github.com/micheleg/dash-to-dock/issues/2193) and causing a hidden dock to show up for a brief moment when launched.

The problem is caused by not only the `wl-clipboard` but also other short-lived programs stealing focus so, as proposed in https://github.com/micheleg/dash-to-dock/issues/988 and https://github.com/micheleg/dash-to-dock/issues/2064, it would be nice to take it further and implement an option to ignore apps specified by the user.

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -824,7 +824,7 @@ export const DockDash = GObject.registerClass({
 
             // Second: add the new apps
             running.forEach(app => {
-                if (!showFavorites || !(app.get_id() in favorites))
+                if ((!showFavorites || !(app.get_id() in favorites)) && !(app.get_windows()[0].get_title() === 'wl-clipboard'))
                     newApps.push(app);
             });
         }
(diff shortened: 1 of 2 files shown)


3v1n0:
Sorry, custom code like this can't be accepted upstream.

I wonder if you can find a window type or class the application matches. Or if the app needs to be fixed

ndr3www:
Well, that's the main problem. `wl-clipboard` has no window class or anything specific assiociated with it, I'm aware that the title matching is not a good solution, but couldn't think of anything better unfortunately.

sergio-costas:
@ndr3www @3v1n0 Maybe delaying for half a second adding a new icon to the dock?

3v1n0:
Bah, it could be a posisbility, but normally I'm not in love with delays since they're not fully predictable, so they may work by luck and depending on the environment.

sergio-costas:
I fully agree with delays; unfortunately, in this case is the only solution, because the point is, precisely, that the window is created and deleted very quickly, so we should only show icons for windows that have been in the screen for a minimum amount of time, and don't add icons to those windows that are there for less time than that minimum.

Issue #1972, opened 2023-03-14
Selecting window from preview does not change focus


Spirrwell:
Hello!

I'm running Dash to Dock v79. My preferred "Click action" behavior is to minimize or show previews. When I click on an application that has multiple windows and I click on the preview for the window I want, the focus does not change to that window. It brings the window to the foreground, but that window is not focused. I have to click on the window itself for it to be focused.

I don't recall having this issue before. I've been doing a lot of distro hopping, so I can't say if it's THIS version of the extension that broke. I'm on a pretty fresh install of Garuda GNOME edition at the moment with GNOME 43.3. I tried disabling all other extensions to see if perhaps there was some weird conflict and there was no change.

If there is any other information I should provide, please let me know.

Thanks!

EDIT: For what it's worth, focus DOES change if I use the "Cycle through windows" click action.

TL;DR Info:
Extension: Dash to Dock v79
DE: GNOME 43.3
OS: Garuda Linux

Striffly:
Same issue for me. I’ve noticed several points:
- If my focus is on app A and I click on the preview of app B, the focus switches correctly.
- If app A has two windows (1 and 2), and I’m already focused on window 1, clicking on the preview of window 2 doesn’t change the focus and it stays on window 1. The focus only switches if I click a second time on the preview of window 2.

thvlio:
Also having this issue, same behavior as the others mentioned. I'll add the disruption that made me notice it: I was trying to use multiple terminals and multiple open folders at once. After clicking on any of the previews on any app that has more than one window opened, clicking on the desired window doesn't change focus, it just brings it to the front. I either have to click on some other window then back on the window I want focus on or click on the desired window preview a second time.

Using Ubuntu 24.04 (GNOME 46), extension version 92. When I was using Ubuntu 22.04 (GNOME 42), extension version 84, I don't think this was happening, if it helps at all.

e-rakavest:
I guess the related issue  https://github.com/micheleg/dash-to-dock/issues/2235

vanvugt:
Ubuntu bug: https://bugs.launchpad.net/bugs/1997550

PR #2245, opened 2024-07-01
Fix window selected from preview not getting focus

Voltara:
This is a fix for an issue where activating a window from the window preview list would not reliably give focus to that window (#1972)

I don't know why reordering these two lines of code fixes it, but there's [confirmation](https://github.com/micheleg/dash-to-dock/issues/1972#issuecomment-2200097373) from another user that the change also works for them.


3v1n0:
Thanks

PR #2258, opened 2024-08-05
Icons: update state cleanups

3v1n0:
Make possible to theme updating icons and avoid actions to be available during that phase.


3v1n0:
Oh, I see Sergio is going to be out for some days, so @vanvugt can you maybe give it a quick check?

Or I'll just merge it so that I can unblock GNOME shell 47 landing.

vanvugt (approved):
Looks reasonable, but I'm no expert on this code.

Issue #2222, opened 2024-05-27
Indicator dots overlap app icons in gnome-shell 46.2


shelvaldes:
Fedora 40, the white dots that indicate an active app are overlapping the actual icon in any case. Here a screenshot.

![white points dock](https://github.com/micheleg/dash-to-dock/assets/35204421/10a15b56-6e7b-4ea2-b285-73d97cb3f999)

andreglatz:
I have the same problem on Arch Linux with Gnome 46.2

![image](https://github.com/micheleg/dash-to-dock/assets/54822067/25e9dedd-d987-494f-ad9a-18147dc175ab)

Calosis:
Same issue here on 46.2

kenny-kelley:
May be related to previous a issue and fix: #2160 and #2168

I had this problem pre-`v90` on any GNOME 46. It's back in GNOME 46.2 + Dash to Dock 92.

I'm also on Fedora 40 using the `dnf` package for Dash to Dock. 100% scaling on a 2560x1440, 27 inch monitor. Seeing similar behavior on my laptop's much smaller monitor too.

These are my settings:
![Screenshot from 2024-05-27 14-57-13](https://github.com/micheleg/dash-to-dock/assets/16879853/8eaded6b-2693-4b9c-a08b-e6b4d0d0f03b)

Toggling on "Use built-in theme" definitely improves it.

akashdiphazra:
This thing appears after recent gnome-shell updates in fedora 40, toggling back indicators from square to dots fixes the issue.

dm1cael:
Same issue with Fedora 40. As @akashdiphazra said, changing from default (or, in his case, square) to dots fixed the issue.

**You can find this option on the "appearence" page.**
![image](https://github.com/micheleg/dash-to-dock/assets/167719716/69d7a6c5-872f-4041-89eb-feff1180047a)

kenny-kelley:
The "Dots" option is slightly different from "Default".

For example, "Default" does a single dot for applications that you have multiple windows of while "Dots" does multiple dots.

GazimSoliev:
I have the same ui issue

ArchUser123:
Changing this option to **"Dots"** instead of Default fixes it for me!
![image](https://github.com/micheleg/dash-to-dock/assets/147475208/5af12912-f2b4-4fd6-a729-5d802a5fd11b)

v9lu:
[...]

PR #2265, opened 2024-08-21
Add `Dot` running indicator style

proninyaroslav:
Dot indicator

It is known that in GNOME 46.2 the default indicator has undergone breaking changes, and now its positioning is faulty: #2222. So I decided it would be nice to make a custom indicator like `Dots`, but with just one dot and as close in size and style as possible to the default indicator.

P.S: It doesn't replace the default indicator for backwards compatibility reasons, but it does provide an alternative in newer versions of GNOME. In the future, this could be made the default, instead of the current `Dots` option.

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -2071,6 +2071,7 @@
                                           <item translatable="yes">Ciliora</item>
                                           <item translatable="yes">Metro</item>
                                           <item translatable="yes">Binary</item>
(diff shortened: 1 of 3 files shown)


3v1n0:
Looks fine, but please squash the fixup commits (like lint errors) and drop the changes to the translation file (is that for another PR?)

proninyaroslav:
@3v1n0 
Done.

Issue #1351, opened 2020-11-20
JS ERROR: TypeError: null has no properties


rekaerst:
# journald spammed with JS ERROR: TypeError: null has no properties
Causing high CPU usage and input lagging
## Log

```log

 JS ERROR: TypeError: null has no properties
                                             _onFocusChanged@resource:///org/gnome/shell/ui/closeDialog.js:135:9
                                             actionMoveWorkspace@resource:///org/gnome/shell/ui/windowManager.js:2200:23
                                             onScrollEvent@/home/arthur/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/docking.js:1304:29

```

repeating at the speed of about 4000 times a minute
## OS information
OS: Ubuntu 20.10 x86_64 
Kernel: 5.8.0-28-generic 
Uptime: 1 day, 10 hours, 12 mins 
Packages: 4103 (dpkg) 
Shell: zsh 5.8 
Resolution: 1920x1080 
DE: GNOME 3.38.1 
WM: Mutter 
WM Theme: Materia-dark-compact 
Theme: Materia-dark-compact [GTK2/3] 
Icons: Papirus-Dark [GTK2/3] 
Terminal: gnome-terminal 
CPU: AMD Ryzen 5 4600H with Radeon Graphics (12) @ 3.000GHz 
GPU: NVIDIA 01:00.0 NVIDIA Corporation TU117M 
GPU: AMD ATI 05:00.0 Renoir 
Memory: 7808MiB / 15431MiB

cgodgers:
I suffer the same problem.

Apparently `killall -3 gnome-shell` partly solves the problem, but I'd expect a better solution than this.

maffelbaffel:
Got the same problem.

roblatham00:
I don't know if "dash-to-dock" is the right project for this issue but I see it too

tonymartino:
Still seeing this:
Aug 25 13:08:37 chainsaw.inside.msnomer.com gnome-shell[4758]: JS ERROR: TypeError: null has no properties
                                                               get settings@file:///home/tony/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/docking.js:1752:28
                                                               updateNotificationBadgeStyle@file:///home/tony/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIconIndicators.js:788:33
[...]

PR #2271, opened 2024-08-26
Fix null errors

sergio-costas:
Several errors are shown when the extension is disabled and re-enabled during desktop lock.

Fix https://github.com/micheleg/dash-to-dock/issues/1351

### DIFF
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -785,8 +785,8 @@ export class UnityIndicator extends IndicatorBase {
         const defaultFontSize = fontDesc.get_size() / 1024;
         let fontSize = defaultFontSize * 0.9;
         const {iconSize} = Main.overview.dash;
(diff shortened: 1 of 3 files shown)


3v1n0:
Mhmh, I feel that we should go more to the root problem, I mean if we're accessing those values after unlock or while re-enabling the extension, it means we're forgetting to disconnect some signal or anything?

sergio-costas:
@3v1n0 That's why I added the "destroy" call, but seems to be not enough. So, in the meantime...

sergio-costas:
Mmm... maybe the problem is with the changes to show notification icons in all icons, even in the search and other places of Gnome Shell? If in any place it tries to read the settings, but the extension is disabled...

3v1n0:
Closing this in favor of https://github.com/micheleg/dash-to-dock/pull/2273

Issue #2272, opened 2024-08-26
memory leak... once again


Nekothegamer:
I don't know why, but as of a few days at this point the extension started leaking on my system, making the shell go from it's usual ~100 MB to over 5 GB, which slows down the entirety of the shell down to a crawl. Again, i have no idea how it started, what i remember before it started leaking is that i was messing with the [GKT4 port of Desktop Icons NG](https://extensions.gnome.org/extension/5263/gtk4-desktop-icons-ng-ding/) and out of nowhere the whole gnome shell interface started to get really slow, and i only discovered today that it reached 6 GB in memory, and only a few minutes ago i found out it was dash to dock's fault, and now I'm here. 

if you want any info there it is:
OS: Fedora 40
Gnome version: 46.4
Extension version: the most recent one, v94
RAM: 16 GB
Display server/protocol: Wayland

UPDATE: this seems to be a bug within v94, since when downgrading to v93 the leak doesn't happen

3v1n0:
Please, check https://github.com/micheleg/dash-to-dock/pull/2273 it should fix it

Nekothegamer:
yeah but it still needs to be merged...

Nekothegamer:
Yes!, thank you!

Issue #2270, opened 2024-08-25
v94 crashes gnome-shell when invoking lock screen.


bennyb0i:
After updating to the latest v94, gnome-shell crashes when the lock-screen is invoked, either manually (i.e., Super+L) or by system timer (e.g., set by gnome-control-center). Journal is repeatedly flooded with the following errors:

```
Aug 25 10:49:08 gnome-shell[1932]: ../glib/gobject/gsignal.c:2685: instance '0x64a82ef43800' has no handler with id '36764'
Aug 25 10:49:08 gnome-shell[1932]: Object .Gjs_ui_appDisplay_FolderView (0x64a82ef43800), has been already disposed — impossible to access it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
    #1   7ffd4de6f200 I   file:///home/user/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/utils.js:184 (3e4f5c39a6f0 @ 214)
    #2   7ffd4de6f230 I   file:///home/user/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/utils.js:101 (3e4f5c39a380 @ 15)
    #4   7ffd4de6f3e0 b   file:///home/user/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/utils.js:101 (3e4f5c39a330 @ 54)
    #5   7ffd4de6f490 b   file:///home/user/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/utils.js:56 (3e4f5c39a0b0 @ 15)
    #7   7ffd4de6f5b0 b   file:///home/user/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/utils.js:55 (3e4f5c39a060 @ 44)
    #8   64a82cafb428 i   file:///home/user/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIconsDecorator.js:61 (3e4f5c3b6b00 @ 43)
    #9   64a82cafb3a8 i   file:///home/user/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIconsDecorator.js:78 (3e4f5c3b6c40 @ 11)
[...]

PR #2273, opened 2024-08-26
appIconsDecorator: Handle signal disconnections on destruction

3v1n0:
Ensure that we destroy the created objects and that we disconnect from all their signals on destruction.

Also, stop monitoring objects destruction in signal handler after an object has been destroyed.

Closes: #2272, #2270


vanvugt (approved):
If it closes the bugs, sure!

PR #2275, opened 2024-08-28
Remove Animate Show Applications from the desktop (Fix "Show Applications" button)

sergio-costas:
The "Show Applications" button fails because the Overview.ANIMATION_TIME property is now read-only (it has a "const" prefix), so it's not possible to overwrite it to show the applications list "instantaneously". It "does nothing" and shows this error in the log:

```
ago 28 12:20:46 raster3-PROX14-AMD gnome-shell[534248]: JS ERROR: TypeError: "ANIMATION_TIME" is read-only
                                                        _onShowAppsButtonToggled/<@file:///home/raster/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/docking.js:2491:21
```

This patch removes the option of disabling the animation, thus fixing the problem.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -15,7 +15,6 @@ import {
     AppDisplay,
     Layout,
     Main,
-    Overview,
     OverviewControls,
     PointerWatcher,
     Workspace,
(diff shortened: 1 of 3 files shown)


3v1n0:
I think it's fine to drop this option, since it was mostly something problematic in the past, HOWEVER... Not sure if some user rely on that (but I guess no because we had no many complaints)

In case we want to keep the option, we could technically just patch `Overview.animateToOverview` and `panel.fadeIn` on click to not animate, but this looks still prone to breakage, so I think we're good dropping it.

3v1n0:
Please, just squash your commits in one

sergio-costas:
@3v1n0 Squashed.

PR #2276, opened 2024-08-29
Progress bar theming tuning

3v1n0:
Move progress bar on icon baseline on overview and permit tuning their positioning and size better via style variables.

This is following the upstream theme, while for yaru some further customization will be done


sergio-costas (approved):
Looks fine. Don't see anything odd.

PR #2277, opened 2024-08-30
Update Russian translations

proninyaroslav:


sergio-costas (approved):
Looks good. Nothing seems out of place.

PR #2286, opened 2024-09-16
appIcons: Show app details menu for snap applications

3v1n0:
If an application is a snap package, and the snap store is installed we can show the snap packages details as we do for other non-snap applications if GNOME Software is installed

UDENG-4316
UDENG-4700

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -22,6 +22,7 @@ import {
 
 import {
     ParentalControlsManager,
+    Util,
 } from './dependencies/shell/misc.js';
 
 import {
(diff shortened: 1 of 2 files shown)


appIcons.js
```diff
@@ -1184,7 +1189,8 @@ const DockAppIconMenu = class DockAppIconMenu extends PopupMenu.PopupMenu {
             }
 
             if (Shell.AppSystem.get_default().lookup_app('org.gnome.Software.desktop') &&
```
sergio-costas, line 1191: Does this still apply? I mean: with the new Snap Store, there is nothing with `org.gnome.Software` at/in `/var/lib/snapd/desktop/applications`...

sergio-costas (commented):
Just a question...

appIcons.js
3v1n0: That's the code for non-snaps, dash-to-dock it's not an ubuntu project so we definitely support also default GNOME installations where GNOME software is installed.

sergio-costas: Oh, ok. Understood.

PR #2285, opened 2024-09-16
theming: Add support for Cogl.Color since Clutter.Color is deprecated in 47

vanvugt:
Closes: https://github.com/micheleg/dash-to-dock/issues/2284

Issue #2284, opened 2024-09-13
JS ERROR: TypeError: (intermediate value).Color is undefined ( GNOME Shell 47.rc )


sidt4:
```
JS ERROR: TypeError: (intermediate value).Color is undefined
_updateDashColor@file:///home/sid/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/theming.js:171:34
updateCustomTheme@file:///home/sid/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/theming.js:229:14
_unloadExtensionStylesheet@resource:///org/gnome/shell/ui/extensionSystem.js:155:15
_callExtensionDisable@resource:///org/gnome/shell/ui/extensionSystem.js:220:14
_onEnabledExtensionsChanged@resource:///org/gnome/shell/ui/extensionSystem.js:637:24
_sessionUpdated@resource:///org/gnome/shell/ui/extensionSystem.js:824:20
ExtensionManager/<@resource:///org/gnome/shell/ui/extensionSystem.js:46:18
_callHandlers@resource:///org/gnome/gjs/modules/core/_signals.js:130:42
_emit@resource:///org/gnome/gjs/modules/core/_signals.js:119:10
_sync@resource:///org/gnome/shell/ui/sessionMode.js:205:14
pushMode@resource:///org/gnome/shell/ui/sessionMode.js:167:14
activate@resource:///org/gnome/shell/ui/screenShield.js:614:34
lock@resource:///org/gnome/shell/ui/screenShield.js:681:14
_prepareForSleep@resource:///org/gnome/shell/ui/screenShield.js:248:22
_callHandlers@resource:///org/gnome/gjs/modules/core/_signals.js:130:42
_emit@resource:///org/gnome/gjs/modules/core/_signals.js:119:10
_prepareForSleep@resource:///org/gnome/shell/misc/loginManager.js:212:14
_callHandlers@resource:///org/gnome/gjs/modules/core/_signals.js:130:42
_emit@resource:///org/gnome/gjs/modules/core/_signals.js:119:10
_convertToNativeSignal@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:152:19
```

Refer [Port Extensions to GNOME Shell 47 ](https://gjs.guide/extensions/upgrading/gnome-shell-47.html).

### For fix refer:

- https://github.com/mgalgs/gnome-shell-system-monitor-applet/issues/90
[...]

PR #2287, opened 2024-09-16
Use Cogl.Color when defined everywhere

3v1n0:
Similar to #2285 to handle more cases of #2284.

/cc @vanvugt

### DIFF
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -1,5 +1,6 @@
 import {
     Clutter,
+    Cogl,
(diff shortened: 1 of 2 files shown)


vanvugt (changes requested):
Two of these commits have nothing to do with "Color", so probably need separate MRs.

3v1n0:
Oh, they were already part of #2286, I just used the wrong base branch.

vanvugt (approved):

PR #2315, opened 2024-10-07
Fix dot position on Gnome 46.2 and greater

sergio-costas:
This patch fixes the dot in Default mode when not enabling "Few customizations".

Tested in Ubuntu Noble, Fedora 46 (100% and 200% scale), and Ubuntu Oracular (both in Ubuntu Desktop and Classic Gnome sessions).

Fix: #2222

### DIFF
--- a/_stylesheet.scss
+++ b/_stylesheet.scss
@@ -272,7 +272,7 @@ $dock_style_modes: [null, shrink, extended, extended-shrink];
 
 .dash-item-container .overview-tile .app-grid-running-dot {
     margin-bottom: $dash_padding_gnome - 4px;
-    //offset-y: -$dash_padding + $dash_edge_offset;
+    offset-y: -$dash_padding + $dash_edge_offset;
 }
 
 @each $side in $dock_sides {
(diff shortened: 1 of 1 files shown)


_stylesheet.scss
```diff
@@ -270,6 +272,7 @@ $dock_style_modes: [null, shrink, extended, extended-shrink];
 
 .dash-item-container .overview-tile .app-grid-running-dot {
     margin-bottom: $dash_padding_gnome - 4px;
+    //offset-y: -$dash_padding + $dash_edge_offset;
```
vanvugt, line 275: It doesn't seem to work in this form. But removing the `//` appears to work.

vanvugt: Also `dash_edge_offset` isn't reused so probably doesn't need to be a variable.

vanvugt (changes requested):

_stylesheet.scss
sergio-costas: Agh! I was doing some tests with and without to ensure that it worked fine, and clearly forgot to remove the comment. Where's the emoji for "banging the head against a wall"?

sergio-costas: Fixed.

PR #2331, opened 2024-11-12
Make the 'Pin to Dock' label translatable

### LINKED ISSUE #2330: "Pin To Dock" menu item is untranslatable
"Pin To Dock" menu label uses the default translation of the Gnome Shell. But the text of the Gnome Shell menu item is "Pin To Dash", therefore this label is always displayed in English.

xalt7x:
With the single underscore (_) we can re-use original translation.   
However, for the 'Pin to Dock' there's no match since GNOME 42-47 original label is 'Pin to Dash'.   
Using two underscores (__) as a translation function resolves #2330

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -1180,7 +1180,7 @@ const DockAppIconMenu = class DockAppIconMenu extends PopupMenu.PopupMenu {
                         favs.removeFavorite(app.get_id());
                     });
                 } else {
-                    const item = this._appendMenuItem(_('Pin to Dock'));
+                    const item = this._appendMenuItem(__('Pin to Dock'));
                     item.connect('activate', () => {
                         const favs = AppFavorites.getAppFavorites();
                         favs.addFavorite(app.get_id());


xalt7x:
fix_pin-to-dock_label


See also:
- #2117 
- [GNOME commit "dash: Use pin instead of favorites"](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/a67877f5201c7f5b168329c3758adee3f90ba575/js/ui/appMenu.js#L164)
- [commit "appIcons: Drop code that is not used anymore by any supported version"](https://github.com/micheleg/dash-to-dock/commit/87d64dadff81da35cdef452a3057a1769d2d9999)
- [commit "Fix translation of popupmenu strings (e.g. "All Windows")."](https://github.com/micheleg/dash-to-dock/commit/51db80abc3ec5c6c0621a0ba74f04919cb3f0b12)

sergio-costas:
In my opinion, this should be fixed by updating the translations, not by reusing the translation of a different text.

xalt7x:
@sergio-costas 
> In my opinion, this should be fixed by updating the translations, not by reusing the translation of a different text.
I guess, there's some misunderstanding. Translations files are already updated but translated string is not used.
I'll try to add more details.
With current
`const item = this._appendMenuItem(_('Pin to Dock'));`
This extension tries to re-use GNOME string. It doesn't exist so we receive English/untranslated string.
One can simply change to GNOME string 'Pin to Dash'.
`const item = this._appendMenuItem(_('Pin to Dash'));`
Now an item uses GNOME translation but this hacky solution is bad. Dash to Dock pot file will contain `Pin to Dash instead of `Pin to Dock`. Also, translators would need to update .po files and translate "Dash" as "Dock". For that, they would need to have knowledge about the meaning of this hack, existence of issue #2330 and discussion in this PR.
---
With the simple fix in this PR
`const item = this._appendMenuItem(__('Pin to Dock'));`
If locale .po file contains translated `Pin to Dock` string, it will be used.
I believe it's correct way.

sergio-costas:
@xalt7x Oh!!!! You are right, my fault.

sergio-costas (approved):
As commented in the source code https://github.com/micheleg/dash-to-dock/blob/846f1fdb785e0548fc273cea406807c943d733f4/appIcons.js#L40-L41 and also by the PR author in the comments, this is the right solution.

Issue #2347, opened 2025-01-14
Gnome 48 Alpha Error {TypeError: (intermediate value).disable_unredirect_for_display is not a function}


ryzendew:
Well i am on that early fun and updating or trying to update extensions early to see what comes and figured i would let you know

![image](https://github.com/user-attachments/assets/70f90ee4-34b9-4747-9805-f2e4d188e843)

vanvugt:
Yes it seems to have changed in Mutter 48.alpha:
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4131/diffs#23374bcf081adfa0475bce05d31f7bc2871f4833_1269_1229

PR #2350, opened 2025-01-15
Add GNOME 48 support

vanvugt:
Closes: https://github.com/micheleg/dash-to-dock/issues/2347


ryzendew:
can conform it works and not found any issues as of yet 
image

PR #2354, opened 2025-02-06
CSS: fix syntax style

sergio-costas:
The SASS documentation specifies that a list of elements for the @each statement should be defined without square brackets, but there are two definitions in the .scss file that uses them. This patch fixes it.

source: https://sass-lang.com/documentation/at-rules/control/each/

### DIFF
--- a/_stylesheet.scss
+++ b/_stylesheet.scss
@@ -139,8 +139,8 @@ $dock_icons_distance: 4px;
     }
 }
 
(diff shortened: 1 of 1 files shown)


3v1n0:
True, but this is more a list of items despite being used in each, no?

However I'm fine if it works the same way.

sergio-costas:
Yes, but Code syntax highlight complains... 😇

PR #2355, opened 2025-02-06
Quick'n'dirty fix for gnome shell crash

sergio-costas:


3v1n0:
This got fixed upstream via https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3809 so we can close the PR.

Issue #2174, opened 2024-03-25
Intelligent autohide not working after drag event 


cesar19004:
As the title says, if you drag an icon, the Dock will no longer hide automatically. A workaround is to press the super key and the dock will automatically hide again.

akashdiphazra:
Same  for me also.

https://github.com/micheleg/dash-to-dock/assets/61003681/d13c7781-9ebe-4b41-94ef-1a4044844fe6

PR #2356, opened 2025-02-10
Fix autohide after drag event

cesar19004:
Issue: When starting a drag event while the dock is overlapping (e.g., when a window is maximized) and intelligent autohide is enabled, the dock fails to hide after the drag event ends.
This pull request invokes `_updateDashVisibility` after the drag event concludes to ensure the dock's state is correctly updated.

Fix: #2174


vanvugt:
Thanks

PR #2362, opened 2025-02-23
Changed Pin to Dock back to Pin to Dash and updated the Chinese translation

bashensi:
I'm sorry, but you shouldn't change Pin to Dash to Pin to Dock, but even if you do, make it a translatable version, as the text is inconsistent and can't be translated at all.

I updated the new Chinese translation.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -1180,7 +1180,7 @@ const DockAppIconMenu = class DockAppIconMenu extends PopupMenu.PopupMenu {
                         favs.removeFavorite(app.get_id());
                     });
                 } else {
-                    const item = this._appendMenuItem(__('Pin to Dock'));
+                    const item = this._appendMenuItem(__('Pin to Dash'));
                     item.connect('activate', () => {
                         const favs = AppFavorites.getAppFavorites();
                         favs.addFavorite(app.get_id());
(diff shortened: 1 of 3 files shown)


bashensi (commented):
Fixed Chinese Trans ( ;

xalt7x:
@bashensi 
I don't have a strong opinion on this but as the extension called "Dash to Dock", it seem consistently to refer to extension panel as a "Dock".   
Also, if we decide to use original label, we should drop translatable string and reuse GNOME Shell string. Please see #2331 for details.

bashensi:
I know.

bashensi:
@xalt7x

看看这个吧。
不能翻译上去。

截图 2025-02-24 04-23-00
截图 2025-02-24 04-23-10

但是 Pin to Dash 可以正常调用系统翻译。



而且我还写了一个 Pin to Dash 版本的翻译，如果后续决定改项目了，我就发上来。

bashensi:
H52H94oYcMTZW8Um

这就是 Pin to Dash 的效果。

xalt7x:
@bashensi ,
Thanks for the report. Somehow, I've missed this label in the App Grid.   
See PR above for the fix.

PR #2363, opened 2025-02-23
Make App Grid 'Pin to Dock' label translatable

xalt7x:
With the single underscore (_) we can re-use original GNOME translation.  
However, since the original GNOME string is 'Pin to Dash' (rather than "Pin to Dock"), there's no match.   
Using two underscores (__) as a translation function makes this extension string translatable.

Complements https://github.com/micheleg/dash-to-dock/pull/2331   
Resolves issue mentioned on https://github.com/micheleg/dash-to-dock/pull/2362#issuecomment-2677103494


bashensi:
谢谢你，之后我上传一下中文翻译

3v1n0:
This was fixed via 889114c207f3764043f6254b23403e66782690c1 thanks.

PR #2369, opened 2025-03-04
translation: also generate translation templates for gschema options

medicalwei:
We received PR #2364 that also translated gschema options, however current Makefile does not handle generating templates from gschema. This changeset adds the translation templates for the file.


3v1n0:
Well, I wouldn't ask our kind translators to go too far, these are very dev-specific things so I would just translate what we have in settings UI

ghost:
Please avoid this suggestion. There are too many strings and has almost no value to users.

PR #2387, opened 2025-04-12
Fixing 'Pin to Dock' translation in GNOME list of applications menu

Tommimon:
In pull request #2331 you fixed the 'Pin to Dock' translation in [appIcons.js](https://github.com/micheleg/dash-to-dock/blob/master/appIcons.js) but you forgot that the same translation is used also in [docking.js](https://github.com/micheleg/dash-to-dock/blob/master/docking.js).

I changed the `_` shell translation function to the `__` extension function and added the needed imports.

Now in the GNOME list of applications, the correct translation for 'Pin to Dock' is picked up for the languages that provide it. Previously it was working only in the Dock context menu.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -40,6 +40,12 @@ import {
     Utils,
 } from './imports.js';
 
+import {Extension} from './dependencies/shell/extensions/extension.js';
+
+// Use __ () and N__() for the extension gettext domain, and reuse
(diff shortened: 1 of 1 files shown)


3v1n0:
Lovely, thanks for spotting it! :)

PR #2389, opened 2025-04-13
Update Swedish translation

MorganAntonsson:


po/sv.po
```diff
@@ -8,129 +8,181 @@ msgid ""
...
 
-#: locations.js:635
+#: locations.js:779
 msgid "Empty Trash"
 msgstr "Töm papperskorgen"
 
+#: locations.js:1049
+#, javascript-format
+msgid "Failed to launch “%s”"
+msgstr "Det gick inte att montera “%s”"
```
sergio-costas, line 184: Is this translation correct? It seems to talk about "mounting a drive", not "launching a program"...

sergio-costas (commented):
Looks good, but have a little doubt.

po/sv.po
MorganAntonsson: You are right. Thanks for spotting it. I will fix it.

MorganAntonsson:
Fixed.

PR #2390, opened 2025-04-25
Add dock margin size customization feature

mojoro:
### Description

I added a feature to customize the dock margin size when panel mode is enabled. This feature introduces a slider in the settings, allowing users to adjust the margin from the edge of the screen (default: 0px) up to 300px.

This is particularly useful for setups with differently sized monitors, where getting a consistent amount of space from the edge of the screen can be difficult without access to dynamic units (e.g. vw, %, etc.)
dock-margin-demonstration

### Implementation Details

Added a new GSettings key ([dock-margin-size](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)) to store the margin value.
Updated the settings UI ([Settings.ui](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)) to include a slider for margin adjustment.
Modified the dock positioning logic in [docking.js](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) to respect the margin value.
Bound the slider to the GSettings key in [prefs.js](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html).

### Testing

Tested on GNOME Shell 46 with a dual-monitor setup. (1920x1080 and 3440x1440)
Verified that the margin applies correctly in panel mode, both vertically and horizontally.
Confirmed that the default behavior (0px margin) remains unchanged.

#### Sidenote
This is my first contribution to anything open-source, so feedback is very welcome. If you'd like me to change anything, please let me know.

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -285,6 +285,14 @@ const DockSettings = GObject.registerClass({
             });
     }
 
+    dock_margin_size_value_changed_cb(scale) {
+        // Get the current value of the scale
+        const value = scale.get_value();
+  
+        // Update the GSettings key
+        this._settings.set_int("dock-margin-size", value);
+      }
+
     icon_size_scale_value_changed_cb(scale) {
         // Avoid settings the size consinuosly
         if (this._icon_size_timeout > 0)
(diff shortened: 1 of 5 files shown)


3v1n0 (commented):
Thanks looks good and code changes are small enough not to be a problem to maintain, but... Shouldn't the dock borders also be rounded in such case?

I feel the squared borders don't fit well with this case... So I guess we need some new style classes

mojoro:
Good observation. I didn't notice that. I updated it to use rounded margins when the margin length is greater than 0, and I removed that stray new-line you called out. Sorry it took me so long to get to!

dock-margin-rounded-borders

prefs.js
```diff
@@ -285,6 +285,14 @@ const DockSettings = GObject.registerClass({
             });
     }
 
+    dock_margin_size_value_changed_cb(scale) {
+        // Get the current value of the scale
+        const value = scale.get_value();
+  
+        // Update the GSettings key
+        this._settings.set_int("dock-margin-size", value);
+      }
```
3v1n0: This is not needed if there is already the settings binding, no?

prefs.js
```diff
@@ -587,18 +595,44 @@ const DockSettings = GObject.registerClass({
...
         this._builder.get_object('preview_size_scale').set_value(
             this._settings.get_double('preview-size-scale'));
+        
+
+        // Dock Margin Size
+        const dockMarginSizeScale = this._builder.get_object(
+            "dock_margin_size_scale"
+        );
+        dockMarginSizeScale.set_range(0, 300); // Set the range for the margin size
+        dockMarginSizeScale.set_value(this._settings.get_int("dock-margin-size")); // Initialize with the current value
```
3v1n0: IIRc this is not needed if the binding is used later.

prefs.js
```diff
@@ -587,18 +595,44 @@ const DockSettings = GObject.registerClass({
...
+        dockMarginSizeScale.set_format_value_func((_, value) => {
+            return `${value} px`; // Display the value in pixels
+        });
+
+        dockMarginSizeScale.connect("value-changed", (scale) => {
+            // Update the GSettings key when the value changes
+            this._settings.set_int("dock-margin-size", scale.get_value());
+        });
```
3v1n0: again, this should be managed by the binding

mojoro:
Thank you for the swift feedback!

The previous approach was based on what I could surmise from the GNOME documentation and other size options' implementations and limited by my understanding of the Gtk. 

You are right that managing all of that with the binding is much cleaner. This is my first time working with the Gtk settings, so please let me know if the current implementation matches your expectations, or should I have kept the `dock_margin_size_value_changed_cb` and updated it to match more closely the debounce behavior in `dock_size_scale_value_changed_cb` and `icon_size_scale_value_changed_cb`, prefs.js:276-297?

PR #2412, opened 2025-07-27
Ubuntu dock monitor awareness

bertramdryan:
###Problem
Currently, when an app is launched from the dock, it opens on the primary display, regardless of which monitor the dock interaction occurred on. This behavior can interrupt workflows—such as video calls or media playback—by unexpectedly shifting focus to another display.

### Solution
This PR modifies the activate(button) method in appIcon.js to capture the monitor where the click event originated and request the app to launch on that monitor. This aligns window placement with user intent, creating a more intuitive multi-display experience.

### Notes
Tested with various dock placements and multiple display configurations.

Behavior now matches user expectations similar to macOS, where new windows respect the context of user interaction.

Special thanks to Grok 4 for assisting with code suggestions.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -486,6 +486,9 @@ const DockAbstractAppIcon = GObject.registerClass({
         const event = Clutter.get_current_event();
         let modifiers = event ? event.get_state() : 0;
 
+        // Add this to get the clicked monitor index
+        const monitorIndex = this._dash.monitorIndex;
+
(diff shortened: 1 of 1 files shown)


3v1n0:
This branch should be rebased on master and dropped the ubuntu bits (so just cherry-pick your latest commit).

That said, I think it may be a bit controversial and it may require an option, in fact I do understand that we want to move a window opened from a specific launcher to the right monitor, but we should not move the windows around once opened IMHO.

3v1n0:
Mhmh, it seems like this needs a rebase ;)

Issue #2288, opened 2024-09-19
Dash not showing when applying pressure


mambucodev:
I updated today to gnome 47, and sometimes the dash doesn't come up when it's hidden. I've got intelli auto-hide activated when a focus windows is places on the dash area. I don´t know why but if I reload the extension or I disable and re-enable the intelli auto-hide option, it starts working again but then, after a while, it stops.

I'm on Arch Linux 6.10.10-arch1-1 with Gnome 47

Recording:
[Screencast From 2024-09-19 13-58-04.webm](https://github.com/user-attachments/assets/b6b04083-f8e0-4972-a5c6-47d0bce4a429)
PS: sorry but gnome recording doesn't show the mouse pointer, but I'm moving it up and down to the bottom of the screen to eventually show the dock. Also, the pressure in intelli auto-hide is set to `0`.

Neo-29:
confirmed here on archlinux too
Must tweak the settings to make the dock appear, then first reboot.... it doesn't work anymore

mrpeabody:
Full reset of dash-to-dock section in dconf fixes the problem, but yes, lots of issues, for example, dots/dashes no longer work in custom themes.

Neo-29:
> Full reset of dash-to-dock section in dconf fixes the problem, but yes, lots of issues, for example, dots/dashes no longer work in custom themes.

Even after reboot the computer?

mrpeabody:
> > Full reset of dash-to-dock section in dconf fixes the problem, but yes, lots of issues, for example, dots/dashes no longer work in custom themes.
> 
> Even after reboot the computer?

Correct! Reset /org/gnome/shell/extensions/dash-to-dock reboot, set it up and it works just fine.

Neo-29:
> > > Full reset of dash-to-dock section in dconf fixes the problem, but yes, lots of issues, for example, dots/dashes no longer work in custom themes.
> > 
> > 
> > Even after reboot the computer?
> 
> Correct! Reset /org/gnome/shell/extensions/dash-to-dock reboot, set it up and it works just fine.

Workaround confirmed working 👍

mambucodev:
> > > Full reset of dash-to-dock section in dconf fixes the problem, but yes, lots of issues, for example, dots/dashes no longer work in custom themes.
> > 
> > 
> > Even after reboot the computer?
> 
> Correct! Reset /org/gnome/shell/extensions/dash-to-dock reboot, set it up and it works just fine.

[...]

PR #2421, opened 2025-08-06
Fix: Intelligent Hide Not Working for All Positions

### LINKED ISSUE #2407: Issue with autohide
The auto hide options under intelligent auto hide arent working as expected. The dock appearance on mouse push to show works only when the dock is positioned on the left side of the screen and doesn't work when positioned at top/right/bottom.

Nauscar:
- Fixed pressure barrier for top and bottom positions
- Increased dwell edge to 2px

Fixes #2407

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -48,6 +48,7 @@ const {gettext: __} = Extension;
 
 const {signals: Signals} = imports;
 
+const DOCK_DWELL_EDGE_PX = 2
 const DOCK_DWELL_CHECK_INTERVAL = 100;
 const ICON_ANIMATOR_DURATION = 3000;
 const STARTUP_ANIMATION_TIME = 500;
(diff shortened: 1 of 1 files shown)


Nauscar:
Fixes #2288

docking.js
```diff
@@ -48,6 +48,7 @@ const {gettext: __} = Extension;
 
 const {signals: Signals} = imports;
 
+const DOCK_DWELL_EDGE_PX = 2
```
3v1n0, line 51: Maybe for another PR, but should we also scale this by the monitor scaling?

dmzoneill:
Tangentially related to #2617. This PR fixes pressure barrier geometry for top/bottom dock positions, which controls the hot edge used to reveal a hidden dock. The GNOME 50 bug in #2617 operates at a different level - Chrome actors are made invisible by the Shell during overview transitions, which pressure barriers cannot override. Still, correct barrier setup would be needed for any autohide recovery path once the underlying visibility issue is resolved.

PR #2431, opened 2025-08-29
Update zh_CN.po, 修正笔误

tigertall:


3v1n0:
Hard to tell for me whether this is better than before, but I accept votes :)

tigertall:
> Hard to tell for me whether this is better than before, but I accept votes :)

It seems the mistake just a typo due to the pinyin input method In Chinese。
"是" and "时" use the same intput key "SHI" in keyboard，sometimes we choose the wrong one。

Issue #2413, opened 2025-07-28
(Bug) : Dock isn't dodging windows that are tiled to the right side of screen.


jakegithub24:
# Specs :
- Ubuntu 24.04.02 LTS
- 6.14.04-24-generic
- GNOME 46.0
- Mutter
- 1920x1080p resolution

# Issue description :
I am facing this weird issue since last 6 months. I have installed this "Dash-to-dock" gnome extension as usual. It is working perfectly fine except this one issue. I use default window tiling present in Ubuntu. I prefer dock setting to intelligent autohide and dodge windows (Only focused application's windows) mode. It works good, but with one specific issue.

Whenever I try to tile any application window to the right side of the desktop/screen, the dock behaves weird. It hovers over the application window instead of hiding itself.

I have searched for solution many times, but I was not able to find one. I would really appreciate to hear on this. Thanks.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/3de0db2a-3fde-4e5a-823d-db419bc2e5de" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/965db8e5-ce49-4651-9d0f-22f7dc6510be" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b69e48ee-1dc3-4bf3-b8c6-d4f46cec85a0" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ae6f555b-048b-4ee2-b1cc-e7413fd2cc1f" />

vanvugt:
This may relate to the fact the window isn't moving so much as jumping when tiled. Maybe the dock isn't getting the right events to know about it.

Also, does ubuntu-dock have the same bug?

PixelcrafterEXE:
Ubuntu-dock has the same problem. When scaled far enough beyond the center-line the window is doged. It looks like the position to check is miscalculated somewhere. It hides only when something is within a bounding box with the size of the dock that starts at the left edge (not the center) of the screen.

nneto:
I'm experiencing this. I have an ultrawide monitor. When I snap a window to the left, the dock hides, when I press Super+RightArrow and move to the right, it appears. When I expand some amount of the window (it's more than half of the dock) then it hides again.

<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/a3515b6b-6b72-418d-82e5-0159a3d2d453" />

[...]

PR #2447, opened 2025-09-18
Fix Autohide Location calculation bug (Issue #2413)

PixelcrafterEXE:
Fix Autohide Location calculation bug (Issue #2413)

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -1245,9 +1245,10 @@ const DockedDash = GObject.registerClass({
     }
 
     _updateStaticBox() {
+        const [absX, absY] = this._box.get_transformed_position();
         this.staticBox.init_rect(
-            this.x + this._slider.x - (this._position === St.Side.RIGHT ? this._box.width : 0),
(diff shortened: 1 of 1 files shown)


3v1n0:
Seems fine, thanks!

Neo-29:
The issue here comes back randomly and i have to turn off/on intelligent hiding to get it back to work

Archlinux  gnome 49 up to date 

image

siddhpant:
This causes a regression for me. Dock on left side doesn't autohide. Have to do some weird random things to get it to hide.

The commit message also doesn't explain the issue. So best to revert.

siddhpant:
Okay, I'm not alone. See #2501

This must be reverted ASAP.

sergio-costas:
As commented in the issue, it seems that sometimes, the reported coordinates of the dock window are placed under the screen bottom (thus, it is reported as hidden), but the window dock is really over it (thus, it is visible). I don't know if the bug is in mutter, in clutter, or in ST.

viggy96:
Please revert this immediately, it breaks the autohide functionality.

PR #2460, opened 2025-11-04
Add Turkish translation for 'Pin to Dock'

Exinie:


3v1n0:
Please check if the new translation handles this properly.

Issue #1029, opened 2019-10-17
Optional feature: Bouncing icons


lastcosmonaut:
Make an option to enable bouncing icons.
When the option is set: a click on an icon should make the icon bounce up and down until the app is started.

acerspyro:
Can't be too complicated if GJS supports CSS3 transforms.

PR #2476, opened 2025-12-21
Implementing #1029, icons now bounce if toggled to do so in preferences.

justinnichols:
**DISCLAIMER**: I used GitHub Copilot to help implement this feature.  If this is not allowed or not wanted, please close this PR without merging and I will rewrite the logic myself and re-submit at a later date.

This PR implements the enhancement request in [#1029](https://github.com/micheleg/dash-to-dock/issues/1029).

**Features implemented**:
- Toggle preference for "Icons bounce when launching?" (Needs localization)
- When enabled, upon clicking an application icon that has no launched windows, the icon will bounce a minimum of 1 time and a maximum of 5 times.  If it is able to detect the application target is RUNNING, it will ease out the bounce back to the original location and will not perform another bounce.
- When disabled, the icons do not bounce.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ BASE_MODULES = extension.js \
 
 EXTRA_MODULES = \
                 appSpread.js \
+                bounceAnimation.js \
                 dash.js \
                 docking.js \
                 appIcons.js \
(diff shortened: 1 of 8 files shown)


justinnichols:
Typo in commit log, leaving as is, and will fix if required, but fixed title of this PR.

PR #2477, opened 2025-12-24
Fix crash when accessing disposed GProxyVolume on CD eject

jdblair:
## Summary

GNOME Shell crashes with heap corruption (SIGABRT/SIGSEGV) when ejecting CDs. The `MountableVolumeAppInfo` class accesses disposed `GProxyVolume` objects after gvfs has freed them.

## Root Cause

When gvfs disposes a volume on eject, `MountableVolumeAppInfo` continues to hold a direct reference to it. Calling any method on a disposed GObject causes heap corruption at the C level—before JavaScript exception handling can intervene.

## Solution

Instead of holding a direct reference to the `GProxyVolume` object (which can be disposed by gvfs at any time), cache the volume's identifiers (UUID, device path) at construction and look up the volume fresh from `VolumeMonitor` each time access is needed. If the volume is no longer in `VolumeMonitor`, it's been removed.

Key changes:
- Cache volume identifiers (UUID, unix-device, name) at construction
- Add `_safeVolume` getter that looks up volume by cached identifiers from `VolumeMonitor`
- Add `_safeMount` getter for consistent access pattern  
- Cache ID/name strings for use in error handlers when volume is disposed
- Add `invalidateVolume()` method called on volume removal
- Guard all volume/mount access points to use safe accessors
- Cancel async operations in `_onVolumeRemoved()` before cleanup

**Note:** This PR was updated from an earlier try/catch approach after testing revealed that disposed GObjects crash at the C level before JS exception handling. The current solution avoids holding direct references entirely.

## Testing

- Tested on Ubuntu 25.10 with GNOME Shell 49.0
- Verified CD eject/insert cycles no longer crash
- Errors are now caught gracefully

## Related

- Ubuntu bug: https://bugs.launchpad.net/ubuntu/+source/gnome-shell-extension-ubuntu-dock/+bug/2137078

### DIFF
--- a/locations.js
+++ b/locations.js
@@ -2,7 +2,6 @@
 
 import {
     Gio,
-    GioUnix,
     GLib,
     GObject,
     Shell,
@@ -111,7 +110,7 @@ export const LocationAppInfo = GObject.registerClass({
             GObject.ParamFlags.READWRITE,
             Gio.Cancellable.$gtype),
     },
-}, class LocationAppInfo extends GioUnix.DesktopAppInfo {
+}, class LocationAppInfo extends Gio.DesktopAppInfo {
     static get GJS_BINARY_PATH() {
         if (!this._gjsBinaryPath)
             this._gjsBinaryPath = GLib.find_program_in_path('gjs');


jdblair:
I closed this PR b/c I moved a little too quickly in submitting it! Specifically,

1) I need to test some more
2) Tip of master differs from what I'm running in Ubuntu 25.10 b/c it targets Gnome 50.

This is a real issue with removable media, and I will continue working on it.

jdblair:
The root cause and correct fix is in gvfs, not dash-to-dock. I've submitted the following to gvfs:

* https://gitlab.gnome.org/GNOME/gvfs/-/issues/829
* https://gitlab.gnome.org/GNOME/gvfs/-/merge_requests/289

PR #2494, opened 2026-02-09
fix: add null check in setUrgent and setUpdating methods

### LINKED ISSUE #2482: JS ERROR: TypeError: can't access property "updating", this._source is null
Ran `dbus-run-session gnome-shell --devkit --wayland` and got the following splat after exiting from activities pane.

```
JS ERROR: TypeError: can't access property "updating", this._source is null
setUpdating@file:///home/siddh/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIconIndicators.js:1109:9
destroy@file:///home/siddh/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIconIndicators.js:794:14
_decorateIcon/<@file:///home/siddh/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/appIconsDecorator.js:55:23
callback@file:///home/siddh/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/utils.js:199:17
clear@resource:///org/gnome/shell/ui/search.js:200:44
_clearDisplay/<@resource:///org/gnome/shell/ui/search.js:798:30
_clearDisplay@resource:///org/gnome/shell/ui/search.js:797:25
_reset@resource:///org/gnome/shell/ui/search.js:684:14
setTerms@resource:///org/gnome/shell/ui/search.js:749:18
_onTextChanged@resource:///org/gnome/shell/ui/searchController.js:248:29
reset@resource:///org/gnome/shell/ui/searchController.js:182:9
_onKeyPress@resource:///org/gnome/shell/ui/searchController.js:274:22
```

Version 102 on GNOME 49.2.

dmzoneill:
## Summary
- Adds null checks to prevent accessing `this._source` after it's been set to null
- Fixes race condition during cleanup when exiting activities pane
- Prevents JavaScript error: "can't access property 'updating', this._source is null"

## Fixes
Closes #2482

## Changes
Modified `setUrgent()` and `setUpdating()` methods in `appIconIndicators.js` to check if `this._source` is null before accessing it. This handles the case where these methods are called during the destroy sequence after the parent class has already nulled the source.

## Test plan
- Open GNOME Shell with dash-to-dock enabled
- Open activities pane
- Exit activities pane
- Verify no JavaScript errors are logged
- Check that icon indicators still function normally during normal operation

### DIFF
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -1096,7 +1096,7 @@ export class UnityIndicator extends IndicatorBase {
     }
 
     setUrgent(urgent) {
-        if (urgent || this._isUrgent !== undefined)
+        if (this._source && (urgent || this._isUrgent !== undefined))
             this._source.urgent = urgent;
 
         if (urgent)
@@ -1106,7 +1106,8 @@ export class UnityIndicator extends IndicatorBase {
     }
 
     setUpdating(updating) {
-        this._source.updating = updating;
+        if (this._source)
+            this._source.updating = updating;
     }
 
     _updateIconStyle() {


3v1n0:
Do you have a full trace of when the errors happen?

As ideally we should never call these functions if `this._source` is unset, and if this happens, then it means that we did not properly disconnect from signals.

dmzoneill:
the full trace is in the original issue #2482 - shows it happens during search controller reset when clearing display after exiting activities

PR #2508, opened 2026-02-16
Revert "Fix Autohide Location calculation bug (Issue #2413)"

### LINKED ISSUE #2501: Problem with auto hide dock (v103)
After updating to 103, auto hide dock does not work properly in maximized window and is not hidden.

siddhpant:
This reverts commit 3e0c24f7b09708757e70565a6e449b6143cd2b08.

That commit causes regression where the autohide is broken for regular cases.

Closes: #2501

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -1249,10 +1249,9 @@ const DockedDash = GObject.registerClass({
     }
 
     _updateStaticBox() {
-        const [absX, absY] = this._box.get_transformed_position();
         this.staticBox.init_rect(
-            absX,
-            absY,
+            this.x + this._slider.x - (this._position === St.Side.RIGHT ? this._box.width : 0),
+            this.y + this._slider.y - (this._position === St.Side.BOTTOM ? this._box.height : 0),
             this._box.width,
             this._box.height
         );


gnolooo:
Reverting this commit causes intellihide to malfunction, at least on my machine.

image

Specifically:

- when disabling this setting, the dock is always hidden (as expected)
- when enabling it, the dock should only hide if there is a window in its area, but instead it stays permanently visible.

gogi983:
Solution found here : 

https://github.com/micheleg/dash-to-dock/issues/2501#issuecomment-3909279105


> Maybe it'll be useful for someone. I also faced the same issue after the last update and have noticed, that the issue for me was that the box area used to calculate overlaps was positioned correctly on `x` axis, but not the `y` axis. Interestingly enough, when I reverted that fix the situation flipped: now it was the `x` axis that was off and `y` axis which was correct. So as a duck tape solution I ended up with that, which actually fixed both of the issues for me (which I guess was the initial intent of the broken fix).
> 
>         const [absX] = this._box.get_transformed_position();
> 
>         this.staticBox.init_rect(
>             absX,
>             // this.x + this._slider.x - (this._position === St.Side.RIGHT ? this._box.width : 0),
>             this.y + this._slider.y - (this._position === St.Side.BOTTOM ? this._box.height : 0),
>             this._box.width,
>             this._box.height
>         );

FTR: file to modify is Docking.js, from line 1252.
Location: /home//.local/share/gnome-shell/extensions/[dash-to-dock@micxgx.gmail.com](mailto:dash-to-dock@micxgx.gmail.com)

siddhpant:
That sounds like a random hack with no explanation.

First we should stop the new regression, then fix the existing problem.

siddhpant:
> Reverting this commit causes intellihide to malfunction, at least on my machine.

Does that happen on older version? If yes, then its a different bug.

siddhpant:
> that the issue for me was that the box area used to calculate overlaps was positioned correctly on x axis, but not the y axis. Interestingly enough, when I reverted that fix the situation flipped: now it was the x axis that was off and y axis which was correct.

`this.x` will be 0 in case of left, something in between in case of centre, and the max width in case of right. So the code subtracts the box width in case of right.

The logic is symmetrical, so we can't mix two solutions if we have to be correct. Otherwise it's just a hack for a specific use case.

siddhpant:
Note: For testing, use `dbus-run-session gnome-shell --devkit --wayland`.

3v1n0:
This was merged as part of 0f21b6b9baf504d6e6972e9ea8041240ceadfdc9 so closing, thanks though!

PR #2511, opened 2026-02-19
Fix autohide not working when panel mode is disabled

rodyoukai:
## Problem

When panel mode (dock-fixed) was disabled, the autohide functionality  would stop working. However, when panel mode was enabled, autohide worked  as expected.

## Root Cause

The issue was in the `_trackDock()` function. When the dock was not in  fixed (panel) mode, it was registered with `Main.layoutManager.addChrome(this)`  without any parameters. This prevented proper tracking by GNOME Shell's  Chrome system, which is necessary for hover events and autohide to function  correctly.

In contrast, when panel mode was enabled, the dock was registered with  proper parameters including `trackFullscreen: true`, which allowed the  system to track the dock's interaction correctly.

## Solution

Modified the `_trackDock()` function to always pass tracking parameters  to `Main.layoutManager.addChrome()`, regardless of the panel mode setting:

- When panel mode is **enabled**: `trackFullscreen: true, affectsStruts: true`
- When panel mode is **disabled**: `trackFullscreen: true, affectsStruts: false`

This ensures that GNOME Shell's Chrome system properly tracks the dock  in both modes, allowing hover detection and autohide to work consistently.

## Testing

- Autohide now works correctly when panel mode is disabled
- Autohide continues to work correctly when panel mode is enabled
- Panel mode behavior remains unchanged

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -452,17 +452,19 @@ const DockedDash = GObject.registerClass({
     }
 
     _trackDock() {
+        if (this.get_parent())
+            Main.layoutManager.removeChrome(this);
+            
         if (DockManager.settings.dockFixed) {
-            if (this.get_parent())
-                Main.layoutManager.removeChrome(this);
             Main.layoutManager.addChrome(this, {
                 trackFullscreen: true,
                 affectsStruts: true,
             });
         } else {
-            if (this.get_parent())
-                Main.layoutManager.removeChrome(this);
-            Main.layoutManager.addChrome(this);
+            Main.layoutManager.addChrome(this, {
+                trackFullscreen: true,
+                affectsStruts: false,
+            });
         }
     }
 
--- a/docking.js
+++ b/docking.js
@@ -258,8 +258,9 @@ const DockedDash = GObject.registerClass({
         // Create intellihide object to monitor windows overlapping
         this._intellihide = new Intellihide.Intellihide(this.monitorIndex);
 
-        // initialize dock state
-        this._dockState = State.HIDDEN;
+        // initialize dock state - must be consistent with slide_x initial value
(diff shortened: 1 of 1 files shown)


udhayakumar-in8 (changes requested):
changes work for me gnome latest arch
[docking.patch](https://github.com/user-attachments/files/25434333/docking.patch)
[intellihide.patch](https://github.com/user-attachments/files/25434334/intellihide.patch)

AmionSky:
The math seems to be wrong somewhere.

https://github.com/user-attachments/assets/50afd151-3de9-4995-b366-a3de30397e05

AmionSky:
In `_updateStaticBox()` the usage of `this.x` and `this.y` is not correct.

Using the previous absolute position is a solution, since the issue is the axis that is not modified for the hide.
```js
const [absX, absY] = this._box.get_transformed_position();

// Calculate the position based on dock position and monitor
switch (this._position) {
case St.Side.LEFT:
    staticX = this._monitor.x;
    staticY = absY;
    break;
case St.Side.RIGHT:
    staticX = this._monitor.x + this._monitor.width - width;
    staticY = absY;
    break;
case St.Side.TOP:
    staticX = absX;
    staticY = this._monitor.y;
    break;
case St.Side.BOTTOM:
    staticX = absX;
    staticY = this._monitor.y + this._monitor.height - height;
    break;
...
```

dmzoneill:
This PR's change to always pass trackFullscreen to addChrome() touches the same Chrome registration path that causes #2617. On GNOME 50, addChrome with affectsStruts: true causes panelBox.visible to be set to false during the overview-to-desktop transition, making both the top panel and dock invisible. Changing how _trackDock() registers the dock could alter or fix that behavior.

PR #2512, opened 2026-02-22
Fix overview state stuck at WINDOW_PICKER when disableOverviewOnStartup is enabled

vadimmaksimushkin:
**Initial problem**
There is a minor bug (more of inconvenience) right after the session start up. Using the 3 finger swipe the the application windows opens skipping the overview. After that it comes to its regular behavior

**One line solution**
When overview is disabled the state at start up is stuck at WINDOW_PICKER, this line just sets it HIDDEN


3v1n0:
Thanks!

PR #2516, opened 2026-02-25
Typo Fix: Added the missing `</kbd>` tag for the shell reload key sequence under the "Building" section of the READ.me document

xmers:
Typo fix: added the missing `</kbd>` tag for the shell reload key sequence under the "**Building**" section of the READ.me document.


sergio-costas:
Thanks!

PR #2519, opened 2026-02-28
Attempt at fixing #2501

### LINKED ISSUE #2501: Problem with auto hide dock (v103)

lemondead1:
It seems that recalculating the boundaries after the animation has played fixes dock being stuck in shown position. Fixes #2501

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -1258,7 +1258,7 @@ const DockedDash = GObject.registerClass({
         const absOffsetY = this._position === St.Side.BOTTOM || this._position === St.Side.TOP
             ? (1 - this._slider.slideX) * this._box.height : 0;
         this.staticBox.init_rect(
-            absX + absOffsetX * (this._position === St.Side.RIGHT ? 1 : -1),
+            absX + absOffsetX * (this._position === St.Side.LEFT ? 1 : -1),
             absY + absOffsetY * (this._position === St.Side.TOP ? 1 : -1),
             this._box.width,
             this._box.height


janga1:
Thanks @lemondead1, this fixed the issue for me.

lemondead1:
Fixed a small bug where the dock pops up for a moment when a window is closed. Can someone verify the new commit doesn't reintroduce #2413? On my machine it works fine

janga1:
> Fixed a small bug where the dock pops up for a moment when a window is closed. Can someone verify the new commit doesn't reintroduce #2413? On my machine it works fine

Tested your latest commit and the weird pop up is indeed gone. The dock seems to be working normally again. The tiling of windows also works fine on my machine (Fedora 43, GNOME 49.4). Well done.

schmauso:
I also tested your commits and the auto-hide feature is now working again. Seems fine on Gnome 49.4, GDM 49.2, Arch Linux 6.18.9-zen1-2-zen. Good work, hope for a fast merge.

yochananmarqos:
Running `make check` fails with eslint 9.33.0:

```
ESLINT_USE_FLAT_CONFIG=false eslint  .

/build/gnome-shell-extension-dash-to-dock/src/dash-to-dock/docking.js
   881:1   error  Trailing spaces not allowed                            no-trailing-spaces
   882:40  error  Missing semicolon                                      semi
  1256:1   error  This line has a length of 141. Maximum allowed is 110  max-len
  1256:13  error  'absOffsetX' is never reassigned. Use 'const' instead  prefer-const
  1256:41  error  Expected '===' and instead saw '=='                    eqeqeq
  1256:75  error  Expected '===' and instead saw '=='                    eqeqeq
  1257:1   error  This line has a length of 142. Maximum allowed is 110  max-len
  1257:13  error  'absOffsetY' is never reassigned. Use 'const' instead  prefer-const
  1257:41  error  Expected '===' and instead saw '=='                    eqeqeq
  1257:77  error  Expected '===' and instead saw '=='                    eqeqeq
  1259:49  error  Expected '===' and instead saw '=='                    eqeqeq
  1260:49  error  Expected '===' and instead saw '=='                    eqeqeq

✖ 12 problems (12 errors, 0 warnings)
  4 errors and 0 warnings potentially fixable with the `--fix` option.

(node:1077459) ESLintRCWarning: You are using an eslintrc configuration file, which is deprecated and support will be removed in v10.0.0. Please migrate to an eslint.config.js file. See https://eslint.org/docs/latest/use/configure/migration-guide for details. An eslintrc configuration file is used because you have the ESLINT_USE_FLAT_CONFIG environment variable set to false. If you want to use an eslint.config.js file, remove the environment variable. If you want to find the location of the eslintrc configuration file, use the --debug flag.
(Use `node --trace-warnings ...` to show where the warning was created)
make: *** [Makefile:158: check] Error 1
```

yochananmarqos:
@lemondead1 The tests pass now. :heavy_check_mark:

vanvugt:
The bug has been fixed by a different pull request. Thanks for your efforts and please continue to contribute wherever you can.

dyzevaar:
@vanvugt From what I can see, the “fix” you mention was essentially a revert of a previous commit rather than a real resolution of the underlying issue. This PR attempts to address the root cause and restore the intended behavior.

I understand everyone here is contributing their time voluntarily, and I appreciate that. At the same time, I think it’s still important to aim for a proper fix rather than just reverting changes when problems appear. Could you take another look at the approach in this PR and explain what specifically you think is incorrect about it?

vanvugt:
We need a stable product first that can be trusted even without understanding the source code. That is achieved by reverting to a known safe state as the first step.

~~Second, any solution that keeps `get_transformed_position` around is probably wrong, especially in multi-monitor layouts. Because that function returns global stage coordinates but the rest of the logic seems to work in local actor coordinates. The best way to test this is with two monitors and put the dock on the left of the right hand monitor for example.~~
I take that back, it looks like staticBox is used in global coordinates but that also means its x2/y2 fields are wrong. They are not "width" and "height" fields unless you assume autohide only ever works on the top or left of the top-left monitor. That seems like something that needs fixing.

Since the only remaining bug here is https://github.com/micheleg/dash-to-dock/issues/2413, I think fixing that should be the new goal. If this pull request is modified to do that then reopening it would make sense.

PR #2521, opened 2026-03-03
ok...mouse click can move the dock

adem1414:


adem1414:
make the dock movable by the mouse on opensuse leap

PR #2523, opened 2026-03-04
Fixes for GNOME Shell 50 (and 45, actually)

aleasto:


3v1n0:
Closing as handled via https://github.com/micheleg/dash-to-dock/pull/2546

PR #2524, opened 2026-03-04
Revert "Fix Autohide Location calculation bug (Issue #2413)"

aleasto:
`Clutter.Actor.get_transformed_position()` accounts for animations etc... which means that it cannot be used to compute the "theoretical" global position of the dock, which indeed might be part of an animation at the moment that the method is called.

This reverts commit 3e0c24f7b09708757e70565a6e449b6143cd2b08.

Closes: https://github.com/micheleg/dash-to-dock/issues/2501

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -1249,10 +1249,9 @@ const DockedDash = GObject.registerClass({
     }
 
     _updateStaticBox() {
-        const [absX, absY] = this._box.get_transformed_position();
         this.staticBox.init_rect(
-            absX,
-            absY,
+            this.x + this._slider.x - (this._position === St.Side.RIGHT ? this._box.width : 0),
+            this.y + this._slider.y - (this._position === St.Side.BOTTOM ? this._box.height : 0),
             this._box.width,
             this._box.height
         );


aleasto:
I see no evidence that 3e0c24f could possibly fix #2413
As far as I can see, both ways of getting the global position return roughly the same values, but the old one is always correct even during animations. I couldn't reproduce bug #2413 so I'm not sure what goes wrong.

vanvugt:
I think this is correct. The target box for intellihide should not simply slide offscreen, but should shrink to zero thickness. If it only slides offscreen then it's moving to a location where windows will never touch it and actually never trigger autohide. Remember this ClutterActorBox is defined as {x1, y1, x2, y2}, not {x, y, width, height} as the surrounding code might trick you into thinking.

~~Using `get_transformed_position` is probably also wrong because that returns stage-global coordinates (the full desktop and zero might not be the edge of the current monitor), whereas the old logic appears to be in actor-local coordinates. If you only have one monitor then you wouldn't notice that mistake.~~

Also, simply reverting a recent commit is safer than adding more complexity like #2519.

Issue #108, opened 2014-08-15
Option to not group windows


Majki:
I would love to see option for not grouping windows. Unity for Ubuntu groups them for example. And extra one small dot is not always most readable indicator of having 2 windows opened at the same time.

ticpu:
This demand really sounds Frippery Bottom Panel / like https://extensions.gnome.org/extension/3/bottom-panel/

gtirloni:
This would be awesome.

dinosaurfiles:
Yes, I would also like to have this feature

philroche:
+1 for this.

hitok:
+1

nediml:
+1

Hronom:
+1

chrisjbillington:
Frippery bottom panel is close to this, but does not support multiple monitors.

localdevjs:
Just adding my voice to this feature request, it would be an awesome enhancement!

sevenjay:
the great one could be

peb-adr:
+1

chrisjbillington:
Lol, dash to panel has this feature, but also merges with the top panel (optionally moving it to the bottom).

There is a feature request on dash to panel to make it use its own panel instead of merging with the existing one.

And there's this feature request to make dash to dock be more like a traditional panel.

If both are implemented, the projects will be getting pretty close to having the same feature set which would be pretty funny. Maybe in the long run they should be the same project.

lwbaqueros:
I agree, hopefully both come together, and users can configure both a dock and a top bar as part of just one gnome-extention. 
By the way, having `ungroup apps` option in Dash to Dock would be awesome to speed up productivity.

djboni:
+1
It would be nice to have the option to not group the windows, so we can click directly on what we need.
The possibility to reorder the group, or reorder all applications in the taskbar, would also be very welcome.

aindriu80:
Ungrouping applications would be a killer feature, right-clicking on a program to select the second instance is very uncomfortable. I've a dual monitor display and having one icon for several instances can be very confusing.

sumancbd:
+1
Please can we have this feature? Or is there any config I can change using dconf?
Thankyou

ant0nwax:
+1
Did microsoft actually add a feature like this to their current operating sXstem? Or did microsoft still not add a feature like this to windows 11?

surajhpatil:
+1 as well!

igiona:
[...]

PR #2526, opened 2026-03-06
settings: Add ability to Ungroup Applications

rasyidrafi:
As requested here, https://github.com/micheleg/dash-to-dock/issues/108
build and tested in my ubuntu 25.10 working well, basic ungrouping

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -1447,23 +1447,6 @@
                                 </layout>
                               </object>
                             </child>
-                            <child>
-                              <object class="GtkLabel">
-                                <property name="can_focus">0</property>
-                                <property name="hexpand">1</property>
-                                <property name="label" translatable="yes">Limits tooltip width for long window titles in both grouped and ungrouped modes.</property>
-                                <property name="wrap">1</property>
-                                <property name="xalign">0</property>
-                                <style>
-                                  <class name="dim-label"/>
-                                </style>
-                                <layout>
-                                  <property name="column">0</property>
-                                  <property name="row">1</property>
-                                  <property name="column-span">2</property>
-                                </layout>
-                              </object>
-                            </child>
                           </object>
                         </property>
                       </object>
--- a/Settings.ui
+++ b/Settings.ui
@@ -1032,6 +1032,17 @@
                                 </layout>
                               </object>
                             </child>
+                            <child>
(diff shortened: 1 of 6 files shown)


appIcons.js
```diff
@@ -119,11 +119,12 @@ const DockAbstractAppIcon = GObject.registerClass({
     },
 }, class DockAbstractAppIcon extends Dash.DashIcon {
     // settings are required inside.
-    _init(app, monitorIndex, iconAnimator) {
+    _init(app, monitorIndex, iconAnimator, window = null) {
         super._init(app);
 
         // a prefix is required to avoid conflicting with the parent class variable
         this.monitorIndex = monitorIndex;
+        this.window = window;
```
3v1n0, line 127: I'd call this `singleWindow` or something similar, to make it clear of what it is.

However,  I wonder if extending `DockAbstractAppIcon` with `DockWindowIcon` would be better instead of changing the logic of this one, that is clearly for apps only.

appIcons.js
```diff
@@ -950,10 +959,15 @@ const DockAbstractAppIcon = GObject.registerClass({
...
+    _init(app, monitorIndex, iconAnimator, window = null) {
+        super._init(app, monitorIndex, iconAnimator, window);
 
-        this._signalsHandler.add(tracker, 'notify::focus-app', () => this._updateFocusState());
+        if (window) {
+            this._signalsHandler.add(global.display, 'notify::focus-window',
+                () => this._updateFocusState());
+        } else {
+            this._signalsHandler.add(tracker, 'notify::focus-app', () => this._updateFocusState());
+        }
```
3v1n0, line 970: Ditto, no reason for having this while we should instead use another type.

Settings.ui
```diff
@@ -1447,23 +1447,6 @@
                                 </layout>
                               </object>
                             </child>
-                            <child>
-                              <object class="GtkLabel">
-                                <property name="can_focus">0</property>
-                                <property name="hexpand">1</property>
-                                <property name="label" translatable="yes">Limits tooltip width for long window titles in both grouped and ungrouped modes.</property>
```
3v1n0, line 1454: squash with previous commit

3v1n0:
Per se the change can be welcome by some people, and I'm not opposed, but I think we should try not to touch the app logic and rather use different code paths when this is used.

Issue #2255, opened 2024-07-22
Gnome-shell crashing when the dash-to-dock is set to autohide


petr-hybler-ofti:
When I set the dash-to-dock to autohide and I suspend the laptop, upon awaking it, the gnome shell crashing .. in the gnome-shell journalctl I can see this:

```
JS ERROR: TypeError: dockManager is null
                                          getApps@/usr/share/gnome-shell/extensions/ubuntu-dock@ubuntu.com/locations.js:1266:9
                                          getRunningApps@/usr/share/gnome-shell/extensions/ubuntu-dock@ubuntu.com/locations.js:1276:12
                                          _ensureLocations/<@/usr/share/gnome-shell/extensions/ubuntu-dock@ubuntu.com/docking.js:1714:56
                                          _create/object[name]@/usr/share/gnome-shell/extensions/ubuntu-dock@ubuntu.com/utils.js:296:68
                                          _syncRunningApplications@resource:///org/gnome/shell/misc/introspect.js:74:36
                                          IntrospectService/<@resource:///org/gnome/shell/misc/introspect.js:42:34
                                          updateSensitivity@resource:///org/gnome/shell/gdm/authPrompt.js:578:18
                                          _activateNext@resource:///org/gnome/shell/gdm/authPrompt.js:269:14
                                          _initInputRow/</<@resource:///org/gnome/shell/gdm/authPrompt.js:214:26
```
                                          
 ` Spurious clutter_actor_allocate called for actor 0x55655bb70430/<dashtodockContainer>[<Gjs_ubuntu-dock_ubuntu_com_docking_DashToDock>:0x55655bb70430] which isn't a des>`

`Object .Gjs_ubuntu-dock_ubuntu_com_dash_DockDash (0x55655bbd7330), has been already disposed — impossible to access it. This might be caused by the object having been `

Steps to repeat:

- enable Intelligent autohide
- suspend the laptop
- awake the laptop
- all pinned (favorite) apps are gone .. only "activities" icon is available 

When the activites icon is clicked, the system kinda frozes ...

Steps to fix:
- CTRL ALT F3 -> login as the same user ... run `sudo killall -1 gnome-shell` 
- CTRL ALT F2 -> when the gnomeshell restarts, it's all back to normal 

`$ gnome-shell --version
GNOME Shell 42.9
`
System: ubuntu 22.04.4 LTS
Windowing system: X11

PS: Wayland seems to be working fine tho

vanvugt:
[...]

PR #1366, opened 2020-12-31
Do not animate disposed icons

rhendric:
Fixes #1332.

Issue #1332, opened 2020-10-19
gnome-shell journal error flood when dash-to-dock is viewing 


akrava:
I've noticed, that extension is flooding too much in journalctl:
```
окт 19 23:41:58 UserPC gnome-shell[2975]: Object St.Bin (0x558698bafc10), has been already deallocated — impossible to set any property on it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
окт 19 23:41:58 UserPC gnome-shell[2975]: == Stack trace for context 0x55869671d900 ==
окт 19 23:41:58 UserPC gnome-shell[2975]: #0   7ffcec58aa60 b   /home/a_krava/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/docking.js:2036 (1b57f85acf88 @ 194)
окт 19 23:41:58 UserPC gnome-shell[2975]: Object St.Bin (0x558698bafc10), has been already deallocated — impossible to set any property on it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
окт 19 23:41:58 UserPC gnome-shell[2975]: == Stack trace for context 0x55869671d900 ==
окт 19 23:41:58 UserPC gnome-shell[2975]: #0   7ffcec58aa60 b   /home/a_krava/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/docking.js:2036 (1b57f85acf88 @ 194)
окт 19 23:41:58 UserPC gnome-shell[2975]: Object St.Bin (0x558698bafc10), has been already deallocated — impossible to set any property on it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
окт 19 23:41:58 UserPC gnome-shell[2975]: == Stack trace for context 0x55869671d900 ==
окт 19 23:41:58 UserPC gnome-shell[2975]: #0   7ffcec58aa60 b   /home/a_krava/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/docking.js:2036 (1b57f85acf88 @ 194)
окт 19 23:41:58 UserPC gnome-shell[2975]: Object St.Bin (0x558698bafc10), has been already deallocated — impossible to set any property on it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
[...]

PR #2533, opened 2026-03-10
Fix crash on disposed ScrollView in ensureActorVisibleInScrollView

### LINKED ISSUE #2532: Crash: ensureActorVisibleInScrollView accesses disposed ScrollView
## Bug description

`ensureActorVisibleInScrollView()` in `dash.js` crashes with a `TypeError` when the dash's `ScrollView` has already been destroyed (disposed).

The error manifests as repeated journal entries:

```
Object St.ScrollView (0x…), has been already disposed — impossible to get any property from it.
JS ERROR: TypeError: scrollView.vscroll is null
    ensureActorVisibleInScrollView@dash.js:1146
    _createAppItem/<@dash.js:569
```

## How to reproduce

1. Use Dash to Dock (or ubuntu-dock) alongside an alt-tab extension such as [Advanced Alt-Tab Window Switcher](https://extensions.gnome.org/extension/4412/advanced-alt-tab-window-switcher/)
2. Lock and unlock the screen (or trigger a suspend/resume cycle)
3. Use Alt-Tab to switch windows

The alt-tab extension calls `Main.activateWindow()`, which triggers the `notify::focused` signal on the `appIcon`. The signal handler in `_createAppItem()` then calls `ensureActorVisibleInScrollView(this._scrollView, item)` — but `this._scrollView` has already been disposed during the screen lock cycle.

## Root cause

The signal handlers connected in `_createAppItem()` (`clicked`, `key-focus-in`, `notify::focused`, `notify::urgent`) all reference `this._scrollView` through their closure, but they are **never disconnected** in `_onDestroy()`. When the dash is destroyed and recreated (e.g. after screen unlock), the old signal handlers remain active on appIcon objects that may still be alive, pointing to a now-disposed ScrollView.

## Related issues

- #1332 — Same class of bug (disposed `St.Bin` objects), fixed by PR #1366 which added disposal guards for icon animations
- #2255 — Crash with "DockDash has been already disposed" on autohide + suspend, same disposal pattern

## Proposed fix

PR incoming with a two-part fix:
1. **Root cause**: disconnect the 4 signal handlers in `_onDestroy()` so they cannot fire after the ScrollView is gone
2. **Defensive guard**: add a null/disposed check at the top of `ensureActorVisibleInScrollView()` as a safety net

## Environment

- GNOME Shell 46.0
- Ubuntu 24.04 (noble) with ubuntu-dock (dash-to-dock v90ubuntu3)
- Wayland session
- Advanced Alt-Tab Window Switcher v49.0

Benibur:
## Summary

Fixes #2532 — `ensureActorVisibleInScrollView()` crashes when the dash's `ScrollView` has been disposed, typically triggered by alt-tab extensions calling `Main.activateWindow()` after a screen lock/unlock cycle.

## Changes

Two-part fix in `dash.js`:

### 1. Root cause fix: disconnect signals in `_onDestroy()`

The `clicked`, `key-focus-in`, `notify::focused`, and `notify::urgent` signal handlers connected in `_createAppItem()` all call `ensureActorVisibleInScrollView(this._scrollView, ...)` but were never disconnected when the dash is destroyed. This means they could fire after `this._scrollView` is disposed.

- Added `this._appIconSignals` array to track signal connections
- Stored signal IDs when connecting in `_createAppItem()`
- Disconnect all tracked signals in `_onDestroy()`

### 2. Defensive guard in `ensureActorVisibleInScrollView()`

Added a null/disposed check at the function entry point, returning `[0, 0]` if the `scrollView` or `actor` is invalid. This follows the same pattern as PR #1366 which added disposal guards for icon animations.

## Context

This bug is the same class of disposed-object access that was fixed in:
- PR #1366 (disposed icon animation guards)
- Issue #1332 (disposed `St.Bin` flooding journal)
- Issue #2255 (disposed `DockDash` on autohide + suspend)

The crash is reliably triggered when using extensions that call `Main.activateWindow()` externally (e.g. Advanced Alt-Tab Window Switcher), combined with screen lock/unlock or suspend/resume cycles.

## Test plan

- [x] Syntax validated with `node --check`
- [ ] Verify no more `ScrollView disposed` errors in journal after lock/unlock + alt-tab
- [ ] Verify dock scroll-to-focused behavior still works normally
- [ ] Verify dock behavior on extension disable/enable cycle

### DIFF
--- a/dash.js
+++ b/dash.js
@@ -142,6 +142,7 @@ export const DockDash = GObject.registerClass({
         this._shownInitially = false;
         this._initializeIconSize(this.iconSize);
         this._signalsHandler = new Utils.GlobalSignalsHandler(this);
+        this._appIconSignals = [];
 
         this._separator = null;
 
@@ -313,6 +314,16 @@ export const DockDash = GObject.registerClass({
 
     _onDestroy() {
         this.iconAnimator.destroy();
+
+        // Disconnect scrollView-related signals to prevent access to disposed objects
+        for (const signal of this._appIconSignals) {
+            try {
+                signal.obj.disconnect(signal.id);
+            } catch (e) {
+                // appIcon may already be destroyed
+            }
+        }
+        this._appIconSignals = [];
 
         if (this._requiresVisibilityTimeout) {
             GLib.source_remove(this._requiresVisibilityTimeout);
@@ -524,11 +535,11 @@ export const DockDash = GObject.registerClass({
         item.setChild(appIcon);
 
         appIcon.connect('notify::hover', a => this._ensureItemVisibility(a));
-        appIcon.connect('clicked', actor => {
+        const clickedId = appIcon.connect('clicked', actor => {
             ensureActorVisibleInScrollView(this._scrollView, actor);
         });
 
-        appIcon.connect('key-focus-in', actor => {
+        const keyFocusId = appIcon.connect('key-focus-in', actor => {
             const [xShift, yShift] = ensureActorVisibleInScrollView(this._scrollView, actor);
 
             // This signal is triggered also by mouse click. The popup menu is opened at the original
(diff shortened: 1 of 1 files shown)


3v1n0:
> This concept looks promising, but I think that it would be better to do a class that we can reuse in other places, with a `connect` method that receives the object, the signal name and the callback, and does everything inside. What do you think?

We already have Utils.GlobalSignalsHandler, although ideally I want to migrate one day all this to use upstream signal tracker

3v1n0:
Closing, using instead https://github.com/micheleg/dash-to-dock/pull/2545

PR #2534, opened 2026-03-12
Initial Latvian translation

EriksRemess:
As title says - starting point of Latvian translation. Some strings could change over time.


3v1n0:
Thanks!

PR #2540, opened 2026-03-20
fix: drop direction argument from workspaceSwitcherPopup.display()

Daniel-Escamilla:
GNOME Shell dropped the direction parameter from WorkspaceSwitcherPopup.display() in this commit:
_https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/209d8c7f10b2753a0d3e665f56c4156d4fa03492_

Ubuntu-dock still passes direction as the first argument, which lands in activeWorkspaceIndex. Since direction is never a valid workspace index, no workspace dot is ever highlighted as active.

Fixes: _https://bugs.launchpad.net/ubuntu/+source/gnome-shell-extension-ubuntu-dock/+bug/1966681_

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -1408,7 +1408,7 @@ const DockedDash = GObject.registerClass({
 
                 // Do not show workspaceSwitcher in overview
                 if (!Main.overview.visible)
-                    Main.wm._workspaceSwitcherPopup.display(direction, ws.index());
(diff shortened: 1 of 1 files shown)


3v1n0:
The target branch is wrong though, so I'll adjust that in another PR.

PR #2542, opened 2026-03-21
po: Add Romanian translation

Rajat-XR:
Adds complete Romanian (ro) translation for all strings.
Plural forms follow the standard Romanian 3-form rule (nplurals=3).

Translated using the existing fr.po and de.po files as structural reference.


3v1n0:
Thanks!

PR #2552, opened 2026-03-25
intellihide: fix overlap check getting permanently stuck

### LINKED ISSUE #2551: Autohide randomly? stops working
The autohide function stops working at some point. The dock is always shown over the windows. I cannot tell what triggers it but it starts working again after opening the dock settings window. The issue is several weeks old, and I have it on 30+ laptops (school laptops) running Fedora 43 (some of them Silverblue). The rpm package from Fedora 43 doesn't have this problem, while the one for Fedora 44 does.

The issue appears to be widespread, judging by the comments at https://extensions.gnome.org/.

Alisson-Frota-Soares:
If _doCheckOverlap() throws inside the throttle timeout callback (e.g. due to a null workspace on transient windows in GNOME 48), _checkOverlapTimeoutId is never reset to 0. All subsequent _checkOverlap() calls short-circuit, permanently freezing the dock in its current visibility state.

Fixes:
- Wrap _doCheckOverlap() in try-catch inside timeout callback
- Guard against null workspace/meta_window in filter
- Guard against null compositor private in _windowCreated
- Prevent duplicate signal handlers in _addWindowSignals
- Clear stale timeout state in enable()

Fixes: #2551

### DIFF
--- a/intellihide.js
+++ b/intellihide.js
@@ -107,6 +107,11 @@ export class Intellihide {
     enable() {
         this._isEnabled = true;
         this._status = OverlapStatus.UNDEFINED;
+        this._checkOverlapTimeoutContinue = false;
+        if (this._checkOverlapTimeoutId > 0) {
+            GLib.source_remove(this._checkOverlapTimeoutId);
+            this._checkOverlapTimeoutId = 0;
+        }
         global.get_window_actors().forEach(function (wa) {
             this._addWindowSignals(wa);
         }, this);
@@ -128,11 +133,15 @@ export class Intellihide {
     }
 
     _windowCreated(display, metaWindow) {
-        this._addWindowSignals(metaWindow.get_compositor_private());
+        const dominated = metaWindow.get_compositor_private();
+        if (dominated)
+            this._addWindowSignals(dominated);
         this._doCheckOverlap();
     }
 
     _addWindowSignals(wa) {
+        if (this._trackedWindows.has(wa))
+            return;
         if (!this._handledWindow(wa))
             return;
         const signalId = wa.connect('notify::allocation', this._checkOverlap.bind(this));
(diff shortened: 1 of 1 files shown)


intellihide.js
```diff
@@ -107,6 +107,11 @@ export class Intellihide {
     enable() {
         this._isEnabled = true;
         this._status = OverlapStatus.UNDEFINED;
+        this._checkOverlapTimeoutContinue = false;
```
3v1n0, line 110: If you're removing the timeout, this call has no effect. Otherwise `_checkOverlapTimeoutContinue` is unset (thus false) by default.

Maybe you do want to remove the timeout on disable though.

intellihide.js
```diff
@@ -128,11 +133,15 @@ export class Intellihide {
     }
 
     _windowCreated(display, metaWindow) {
-        this._addWindowSignals(metaWindow.get_compositor_private());
+        const dominated = metaWindow.get_compositor_private();
```
3v1n0, line 136: If've a metaWindow, then we also have a the compositor private side.

intellihide.js
```diff
@@ -248,8 +261,12 @@ export class Intellihide {
     // Optionally skip windows of other applications
     _intellihideFilterInteresting(wa) {
         const metaWin = wa.get_meta_window();
+        if (!metaWin)
+            return false;
         const currentWorkspace = global.workspace_manager.get_active_workspace_index();
         const workspace = metaWin.get_workspace();
+        if (!workspace)
```
3v1n0, line 268: IIRC this can happen only if the window is being destroyed, so in such case we should not even be here.

dmzoneill:
Possibly related to #2617. The intellihide timeout fix here addresses a different layer (overlap checking) than the GNOME 50 visibility regression in #2617 (Chrome actors hidden by Shell during overview transitions). However, if both issues are present simultaneously, a stuck overlap check would prevent intellihide from ever calling _show() again, making the invisible dock completely unrecoverable without a restart.

PR #2558, opened 2026-03-30
intellihide: Fix startup race condition on GNOME 49 with XWayland apps

RickAnderson-hub:
## Problem

On GNOME 49, intellihide intermittently stops responding after login when XWayland applications (e.g. IntelliJ IDEA) are already open. The dock stays visible regardless of window overlap, but begins working correctly as soon as any new window is created (e.g. opening the Extensions app).

## Root causes

**Bug 1 — `_windowCreated` crashes on null compositor private**

`get_compositor_private()` can return `null` for XWayland windows whose actor isn't yet associated with the MetaWindow at signal-emission time. The null was passed directly into `_addWindowSignals`, which called `null.get_meta_window()` and threw a TypeError. Because the crash happened before `_doCheckOverlap()`, the overlap check was silently skipped for that event.

**Bug 2 — Startup race: XWayland windows missed by initial scan**

`enable()` runs a one-shot `_doCheckOverlap()` at initialisation. XWayland window actors can exist in `global.get_window_actors()` before their `MetaWindow` is fully associated. Windows missed at that moment are never re-checked, because no subsequent signal fires for a window that is already static and focused (`restacked` and `notify::focus-app` only fire on changes).

## Fix

- Null-check `get_compositor_private()` in `_windowCreated` before passing to `_addWindowSignals`; `_doCheckOverlap()` still runs regardless.
- Schedule a 500 ms deferred re-scan in `enable()` that picks up any actors whose `MetaWindow` was not ready during the initial pass, and tracks them for future `notify::allocation` events.
- Cancel the deferred timer in both `disable()` and `destroy()` to prevent leaks.

## Tests

Adds a Node.js test suite using `node:test` and `mock.module()` (no new dependencies). 49 tests covering both fixes, core overlap detection, window type filtering, all intellihide modes, throttling, and lifecycle behaviour.

```
make test
```

### DIFF
--- a/intellihide.js
+++ b/intellihide.js
@@ -67,6 +67,7 @@ export class Intellihide {
 
         this._checkOverlapTimeoutContinue = false;
         this._checkOverlapTimeoutId = 0;
+        this._deferredCheckId = 0;
 
         this._trackedWindows = new Map();
 
@@ -102,6 +103,7 @@ export class Intellihide {
 
         // Remove  residual windows signals
         this.disable();
+        this._cancelDeferredCheck();
     }
 
     enable() {
@@ -111,6 +113,22 @@ export class Intellihide {
             this._addWindowSignals(wa);
         }, this);
         this._doCheckOverlap();
+
+        // Deferred re-check: during startup or session restore, some window
+        // actors (especially XWayland) may not be fully ready yet, causing
+        // the initial overlap check to miss them. Re-scan after a delay to
+        // pick up any windows that became ready after the initial check.
+        this._cancelDeferredCheck();
+        this._deferredCheckId = GLib.timeout_add(
+            GLib.PRIORITY_DEFAULT, 500, () => {
+                this._deferredCheckId = 0;
+                global.get_window_actors().forEach(function (wa) {
+                    if (!this._trackedWindows.has(wa))
+                        this._addWindowSignals(wa);
+                }, this);
+                this._doCheckOverlap();
+                return GLib.SOURCE_REMOVE;
+            });
     }
 
     disable() {
@@ -125,10 +143,21 @@ export class Intellihide {
             GLib.source_remove(this._checkOverlapTimeoutId);
             this._checkOverlapTimeoutId = 0;
         }
+
+        this._cancelDeferredCheck();
+    }
+
+    _cancelDeferredCheck() {
+        if (this._deferredCheckId > 0) {
+            GLib.source_remove(this._deferredCheckId);
+            this._deferredCheckId = 0;
+        }
     }
 
     _windowCreated(display, metaWindow) {
-        this._addWindowSignals(metaWindow.get_compositor_private());
+        const wa = metaWindow.get_compositor_private();
+        if (wa)
+            this._addWindowSignals(wa);
         this._doCheckOverlap();
     }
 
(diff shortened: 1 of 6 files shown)


RickAnderson-hub:
@copilot apply changes based on the comments in [this thread](https://github.com/micheleg/dash-to-dock/pull/2558#pullrequestreview-4029227660)

tests/mocks/gi.js
```diff
@@ -0,0 +1,86 @@
...
+
+export const GLib = {
+    PRIORITY_DEFAULT: 0,
+    SOURCE_CONTINUE: true,
+    SOURCE_REMOVE: false,
+    timeout_add(priority, interval, callback) {
+        const id = _nextTimeoutId++;
+        _pendingTimeouts.set(id, {interval, callback});
+        return id;
+    },
```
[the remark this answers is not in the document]
RickAnderson-hub: @copilot apply changes based on [this feedback](https://github.com/micheleg/dash-to-dock/pull/2558#discussion_r3008594746)

intellihide.js
```diff
@@ -111,6 +113,22 @@ export class Intellihide {
             this._addWindowSignals(wa);
         }, this);
         this._doCheckOverlap();
+
+        // Deferred re-check: during startup or session restore, some window
+        // actors (especially XWayland) may not be fully ready yet, causing
+        // the initial overlap check to miss them. Re-scan after a delay to
+        // pick up any windows that became ready after the initial check.
+        this._cancelDeferredCheck();
+        this._deferredCheckId = GLib.timeout_add(
```
sergio-costas, line 122: Why do you use a timeout? Can you guarantee that after that time all the windows will be ready?

sergio-costas (changes requested):

intellihide.js
RickAnderson-hub: It’s a practical workaround for an inherent race condition. During session restore, XWayland apps can show up in `get_window_actors()` before their underlying MetaWindow is fully mapped and positioned, which causes the overlap geometry check to return incorrect results. Adding a delay gives the compositor time to finish initializing those actors.

That said, there’s no guarantee that all windows will be ready after this delay. The 500 ms value is just a best-effort heuristic, not a reliable signal. GNOME Shell doesn’t provide an API that indicates when all startup windows are ready. On slower hardware or when restoring many windows, 500 ms may still be insufficient; on faster systems, it’s likely longer than necessary. Signals like `window-added` or `map` on individual actors aren’t dependable either, since XWayland geometry can still be incorrect when those signals fire.

Even if the deferred check runs too early, the `_trackedWindows` mechanism ensures that any later geometry changes (such as moves or resizes) will trigger new overlap checks through the existing per-window signals. The timeout only needs to cover the narrow case where a window exists but hasn’t moved yet and its overlap was missed at startup. After that, normal signal-driven handling takes over.

sergio-costas: My point is: is there a way of detecting in the timeout callback that not all windows have been initialized and try again, until all them are?

RickAnderson-hub: "GNOME Shell doesn’t provide an API that indicates when all startup windows are ready."

sergio-costas: Yes, but can we check it somehow? Not a signal or a call that says "all windows are ready", but something like "this window is ready", that we can check in each window...?

PR #2559, opened 2026-04-01
Do not mark the whole Kazakh translation as fuzzy

AsciiWolf:
/cc @crayxt


crayxt:
I am ok with this (wasn't sure about that one).

PR #2565, opened 2026-04-14
docking: Avoid inserting ("tracking") docks during the login animation

vanvugt:
Up until GNOME 49 we could be confident that extensions were activated before the login animation, but that's not true in GNOME 50 (gnome-shell#8917). In GNOME 50 we get constructed _during_ the GNOME login animation so inserting docks at that moment creates a visible glitch.

Now we wait until the GNOME login animation is done before starting our own dock login animation. Or if we're not starting up then insert docks immediately.

Closes: https://bugs.launchpad.net/bugs/2146516

PR #2569, opened 2026-04-15
Fix delayed staged dock

3v1n0:
See commits for details, but since 8aa70efb6 we may have a situation in which the dock is added to the stage later than its initialization, while various code may do operations that require the dock to be staged.

So delay all these actions until we're good to go

/cc @vanvugt 

Fixes: #2565

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -421,25 +421,6 @@ const DockedDash = GObject.registerClass({
         this._slider.set_child(this._box);
         this._box.add_child(this.dash);
 
-        // Create and apply height/width constraint to the dash.
-        if (this._isHorizontal) {
-            this.connect('notify::width', () => {
-                this.dash.setMaxSize(this.width, this.height);
-            });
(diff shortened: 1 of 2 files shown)


vanvugt:
~~I feared this might happen, but what are the symptoms to look for?~~

Oh it's [this](https://github.com/micheleg/dash-to-dock/pull/2565#issuecomment-4247521577).

vanvugt:
Tested and confirmed it fixes the log spam.

The rest of the changes sound reasonable but I worry a little that with so many changes we may struggle to justify them to the release team in time for 26.04.

Side note: I wonder if it is now easier to fix [LP#2035304](https://bugs.launchpad.net/bugs/2035304)...

3v1n0:
> Side note: I wonder if it is now easier to fix LP#2035304

Oh maybe this helps already? As we set the style before showing in theory...

PR #2580, opened 2026-04-24
Use the DEFAULT value from GSettings

sergio-costas:
Currently, the DEFAULT setting for the number-of-open-windows indicator is broken. It should look like the DOTS, but the style is broken because it uses its own class/CSS (the dots are placed over the icon). All other styles work as expected.

This patch changes the behavior, and when the user sets the DEFAULT setting, it will read the default value in GSettigs and use that one. For example, Ubuntu uses the DOTS as default in GSettings, so it will use exactly the same code path as DOTS when choosing DEFAULT. Other distros can change that default, and the code will work as expected, using the default value set in GSettings.

### DIFF
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -58,11 +58,13 @@ export class AppIconIndicator {
             this._indicators.push(unityIndicator);
         }
 
-        switch (runningIndicatorStyle) {
-        case RunningIndicatorStyle.DEFAULT:
-            runningIndicator = new RunningIndicatorDefault(source);
-            break;
+        if (runningIndicatorStyle === RunningIndicatorStyle.DEFAULT) {
+            // If DEFAULT is choosen, use the GSettings default value
+            const defaultValue = settings.get_default_value('running-indicator-style');
+            runningIndicatorStyle = RunningIndicatorStyle[defaultValue.get_string()[0]];
+        }
 
+        switch (runningIndicatorStyle) {
         case RunningIndicatorStyle.DOTS:
             runningIndicator = new RunningIndicatorDots(source);
             break;
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -235,6 +235,7 @@ class RunningIndicatorBase extends IndicatorBase {
 
 // We add a css class so third parties themes can limit their indicator customization
 // to the case we do nothing
+/* eslint-disable no-unused-vars */
 class RunningIndicatorDefault extends RunningIndicatorBase {
     constructor(source) {
         super(source);
@@ -246,6 +247,7 @@ class RunningIndicatorDefault extends RunningIndicatorBase {
         super.destroy();
     }
 }
+/* eslint-enable no-unused-vars */
 
 const IndicatorDrawingArea = GObject.registerClass(
 class IndicatorDrawingArea extends St.DrawingArea {
--- a/schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.dash-to-dock.gschema.xml
@@ -92,7 +92,7 @@
       <description>FIXED: constant transparency. DYNAMIC: dock takes the opaque style only when windows are close to it.</description>
     </key>
     <key name="running-indicator-style" enum="org.gnome.shell.extensions.dash-to-dock.running-indicator-style">
-      <default>'DEFAULT'</default>
+      <default>'DOTS'</default>
       <summary>...</summary>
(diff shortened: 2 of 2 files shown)


vanvugt:
There's a CI failure;
```
/home/runner/work/dash-to-dock/dash-to-dock/appIconIndicators.js
  238:7  error  'RunningIndicatorDefault' is defined but never used. Allowed unused vars must match /(^unused|_$)/u  no-unused-vars
```
Also isn't this fix a little self referencial?
```
    <key name="running-indicator-style" enum="org.gnome.shell.extensions.dash-to-dock.running-indicator-style">
      <default>'DEFAULT'</default>
      <summary>...</summary>
      <description>DEFAULT: .... DOTS: ....</description>
    </key>
```
Default is still DEFAULT. If it's broken then we should either make DEFAULT usable or delete it.

sergio-costas:
@vanvugt Mmm... in Ubuntu, the default value is DOTS... I'll change it too, then.

sergio-costas:
About CI check... it doesn't work on 24.04, it's estlint version is too old... I'll try it in a VM.

appIconIndicators.js
```diff
@@ -58,11 +58,13 @@ export class AppIconIndicator {
...
 
-        switch (runningIndicatorStyle) {
-        case RunningIndicatorStyle.DEFAULT:
-            runningIndicator = new RunningIndicatorDefault(source);
-            break;
+        if (runningIndicatorStyle === RunningIndicatorStyle.DEFAULT) {
+            // If DEFAULT is choosen, use the GSettings default value
+            const defaultValue = settings.get_default_value('running-indicator-style');
+            runningIndicatorStyle = RunningIndicatorStyle[defaultValue.get_string()[0]];
+        }
```
vanvugt, line 65: According to the schema, the default value is DEFAULT so this wouldn't fix anything.
```
    <key name="running-indicator-style" enum="org.gnome.shell.extensions.dash-to-dock.running-indicator-style">
      <default>'DEFAULT'</default>
      <summary>...</summary>
      <description>DEFAULT: .... DOTS: ....</description>
    </key>
```

sergio-costas: Fixed.

vanvugt:
I still don't think this makes sense. Never using `RunningIndicatorDefault` means it shouldn't exist. But also `DEFAULT` can't be removed because it would break schema compatibility.

Since the only actual bug we've seen was overlap then why not just fix the overlap? Make DEFAULT a bottom dot without overlap like it seems was the original intention?

vanvugt:
We don't even need to fix the behaviour of the Default class. I think this alone would be fine as a first step:

```diff
-     <default>'DEFAULT'</default>
+     <default>'DOTS'</default>
```

sergio-costas:
My point is that having two different pieces of code for exactly the same is wrong.

vanvugt:
> My point is that having two different pieces of code for exactly the same is wrong.

I agree, but we might both be wrong until we can understand the original plan:

```js
// We add a css class so third parties themes can limit their indicator customization
// to the case we do nothing
class RunningIndicatorDefault extends RunningIndicatorBase {
```
Also they're not exactly the same if the default mode was meant to put dots in a different place from DOTS.

appIconIndicators.js
```diff
@@ -233,6 +235,7 @@ class RunningIndicatorBase extends IndicatorBase {
 
 // We add a css class so third parties themes can limit their indicator customization
 // to the case we do nothing
+/* eslint-disable no-unused-vars */
```
3v1n0, line 238: I'd rather remove unused things

3v1n0:
My feeling was that the default mode should just use tye upstream dot mode, so not many dots, just one.

appIconIndicators.js
vanvugt: It shouldn't be unused because:
* It's there by design.
* The DEFAULT enum can't be removed (for compatibility).

vanvugt:
> My feeling was that the default mode should just use tye upstream dot mode, so not many dots, just one.

Sounds right. But also fixing the positioning to not draw on top of icons.

sergio-costas:
> My feeling was that the default mode should just use tye upstream dot mode, so not many dots, just one.

Hmm... that makes sense... I'll check it, then.

sergio-costas:
Yes, you are right: in DEFAULT mode there is a single dot. It's not the same than DOTS.

sergio-costas:
Ok, closing this one in favor of https://github.com/micheleg/dash-to-dock/pull/2581

PR #2581, opened 2026-04-28
appIcons: Fix DEFAULT dot position

sergio-costas:
Until now, the dot vertical position was -12, inherited from code from Gnome Shell. But with the current layout, that value is wrong and must be zero to ensure that the dot is placed outside the icon, below it.

This patch fixes it.

This patch only affects the DEFAULT mode. All other modes continue to work as before.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -363,6 +363,7 @@ export const DockAbstractAppIcon = GObject.registerClass({
         super._updateDotStyle();
         const themeNode = this._dot.get_theme_node();
         this._dot.translationX = themeNode.get_length('offset-x');
+        this._dot.translationY = 0;
     }
 
     _addUrgentWindow(window) {


appIcons.js
```diff
@@ -363,6 +363,7 @@ export const DockAbstractAppIcon = GObject.registerClass({
         super._updateDotStyle();
         const themeNode = this._dot.get_theme_node();
         this._dot.translationX = themeNode.get_length('offset-x');
+        this._dot.translationY = 0;
     }
```
vanvugt, line 367: I'm thinking we should just drop the whole theme node because it's not providing us with any information other than offset-y:

From gnome-shell:
```css
    // running app dot
    .app-grid-running-dot {
      // manually position the dot within the dash item
      @if $contrast == 'high' {
        offset-y: -$dash_padding - 1px; // don't draw dot directly on inset
      } @else {
        offset-y: -$dash_padding;
      }
    }
```
and
```js
    _updateDotStyle() {
        const themeNode = this._dot.get_theme_node();
        this._dot.translationY = themeNode.get_length('offset-y');
    }
```

vanvugt: Oh we can't nullify it because that would break `_updateDotStyle`.

vanvugt:
Tested. Looks good, thanks.

PR #2584, opened 2026-05-02
Cycle or minimize feature

lullptr:
Feature: cycle windows and minimize at end.

Behaviour: windows are cycled normally. When the cycle reaches the end, all the windows in the group are minimized together.

This is something I find useful, so sharing in case it's helpful to others.

Logic and Code is adapted from dash-to-panel

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -56,14 +56,15 @@ const clickAction = Object.freeze({
     MINIMIZE: 1,
     LAUNCH: 2,
     CYCLE_WINDOWS: 3,
-    MINIMIZE_OR_OVERVIEW: 4,
-    PREVIEWS: 5,
-    MINIMIZE_OR_PREVIEWS: 6,
-    FOCUS_OR_PREVIEWS: 7,
-    FOCUS_OR_APP_SPREAD: 8,
-    FOCUS_MINIMIZE_OR_PREVIEWS: 9,
-    FOCUS_MINIMIZE_OR_APP_SPREAD: 10,
-    QUIT: 11,
+    CYCLE_OR_MINIMIZE: 4,
+    MINIMIZE_OR_OVERVIEW: 5,
+    PREVIEWS: 6,
+    MINIMIZE_OR_PREVIEWS: 7,
+    FOCUS_OR_PREVIEWS: 8,
+    FOCUS_OR_APP_SPREAD: 9,
+    FOCUS_MINIMIZE_OR_PREVIEWS: 10,
+    FOCUS_MINIMIZE_OR_APP_SPREAD: 11,
+    QUIT: 12,
 });
 
 const scrollAction = Object.freeze({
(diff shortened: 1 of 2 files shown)


appIcons.js
```diff
@@ -616,6 +617,25 @@ export const DockAbstractAppIcon = GObject.registerClass({
...
                 break;
 
+            case clickAction.CYCLE_OR_MINIMIZE:
+                if (!Main.overview.visible) {
+                    if (this.focused ||
+                        (recentlyClickedApp === this.app &&
+                            recentlyClickedAppWindows[recentlyClickedAppIndex % recentlyClickedAppWindows.length] === 'MINIMIZE'
+                        ) &&
+                        !hasUrgentWindows) {
+                        // reversed=false, shouldMinimize=true
```
sergio-costas: I think that this should be put this way for clarity

```suggestion
                    if ((this.focused ||
                        (recentlyClickedApp === this.app &&
                            recentlyClickedAppWindows[recentlyClickedAppIndex % recentlyClickedAppWindows.length] === 'MINIMIZE'
                        )) &&
                        !hasUrgentWindows) {
                        // reversed=false, shouldMinimize=true
```

appIcons.js
sergio-costas: What about putting both cases here, instead of duplicating all the code?

sergio-costas (changes requested):

appIcons.js
```diff
@@ -56,14 +56,15 @@ const clickAction = Object.freeze({
...
-    FOCUS_MINIMIZE_OR_APP_SPREAD: 10,
-    QUIT: 11,
+    CYCLE_OR_MINIMIZE: 4,
+    MINIMIZE_OR_OVERVIEW: 5,
+    PREVIEWS: 6,
+    MINIMIZE_OR_PREVIEWS: 7,
+    FOCUS_OR_PREVIEWS: 8,
+    FOCUS_OR_APP_SPREAD: 9,
+    FOCUS_MINIMIZE_OR_PREVIEWS: 10,
+    FOCUS_MINIMIZE_OR_APP_SPREAD: 11,
```
3v1n0: I need to double check, but changing the mapping here does not break previous installations?

appIcons.js
lullptr: How about this code below for merging both cases? also let me know if I need to commit first.

```javascript
case clickAction.CYCLE_WINDOWS:
case clickAction.CYCLE_OR_MINIMIZE: {
    const shouldMinimize = buttonAction === clickAction.CYCLE_OR_MINIMIZE;
    
    const shouldCycle =
        this.focused ||
            (
                shouldMinimize &&
                recentlyClickedApp === this.app &&
                recentlyClickedAppWindows[
                    recentlyClickedAppIndex %
                    recentlyClickedAppWindows.length
                ] === 'MINIMIZE'
            );

    if (!Main.overview.visible) {
        if (shouldCycle && !hasUrgentWindows) {
            this._cycleThroughWindows(false, shouldMinimize);
        } else {
            // Activate the first window
            const [w] = windows;
            Main.activateWindow(w);
        }
    } else {
        this.app.activate();
    }
    break;
}
```

appIcons.js
lullptr: Good catch, I think your assumption is correct.
What do you think about adding the new action before `QUIT`?

```javascript
const clickAction = Object.freeze({
    SKIP: 0,
    MINIMIZE: 1,
    LAUNCH: 2,
    CYCLE_WINDOWS: 3,
    MINIMIZE_OR_OVERVIEW: 4,
    PREVIEWS: 5,
    MINIMIZE_OR_PREVIEWS: 6,
    FOCUS_OR_PREVIEWS: 7,
    FOCUS_OR_APP_SPREAD: 8,
    FOCUS_MINIMIZE_OR_PREVIEWS: 9,
    FOCUS_MINIMIZE_OR_APP_SPREAD: 10,
    CYCLE_OR_MINIMIZE: 11,
    QUIT: 12,
});
```

appIcons.js
sergio-costas: Looks good!

appIcons.js
```diff
@@ -890,6 +906,8 @@ export const DockAbstractAppIcon = GObject.registerClass({
         if (appWindows.length < 1)
             return;
 
+        if (shouldMinimize) appWindows.push('MINIMIZE');
+
```
3v1n0, line 910: Nah I don't like this...

PR #2596, opened 2026-05-22
Makefile: Remember gschemas.compiled for local installs

### LINKED ISSUE #2595: make install no longer installs gschemas.compiled, and the extension can't load
```
(gnome-shell:48607): GNOME Shell-CRITICAL **: 15:05:45.071: Extension dash-to-dock@micxgx.gmail.com: GLib.FileError: Failed to open file “/home/dan/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/schemas/gschemas.compiled”: open() failed: No such file or directory

Stack trace:
  getSettings@resource:///org/gnome/shell/extensions/sharedInternals.js:101:53
  DockManager@file:///home/dan/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/docking.js:1711:42
  enable@file:///home/dan/.local/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/extension.js:11:23
  ...
```

This is a regression from https://github.com/micheleg/dash-to-dock/commit/1ad2628e32e596b32fbdb1bcf3d1088f7eb50033

vanvugt:
Or else the extension won't load at all.

Closes: https://github.com/micheleg/dash-to-dock/issues/2595
Fixes: 1ad2628e32 ("Makefile: Do not ship settings schema on zip file")

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,6 +107,7 @@ ifeq ($(INSTALLTYPE),system)
 	cp -r ./schemas/*.gschema.xml $(SHARE_PREFIX)/glib-2.0/schemas
 	cp -r ./_build/locale/* $(SHARE_PREFIX)/locale
 endif
+	cp schemas/gschemas.compiled $(INSTALLBASE)/$(INSTALLNAME)/schemas/
 	-rm -fR _build
 	echo done
 


Makefile
3v1n0, line 116: I'm fine to add this back, but we need to filter it out in the zip file.

sergio-costas, line 116:
```suggestion
	zip -qr "$(UUID)$(VSTRING).zip" . -x schemas/gschemas.compiled
```

sergio-costas (dismissed):
Suggestion based on review.

Makefile
vanvugt: Yes it looks like that problem exists, but it probably shouldn't exist. Why isn't `zip-file` based on the install tree instead?

vanvugt: Sounds like a bandaid to a problem that shouldn't exist so I would like to resolve [the other discussion](https://github.com/micheleg/dash-to-dock/pull/2596#discussion_r3296827619) instead.

vanvugt: Oh I am also modifying the local install tree. I'll just modify the fix to not appear in the zip.

vanvugt: I've pushed a different solution.

sergio-costas (approved):
Yes, it makes more sense this way.

PR #2600, opened 2026-05-26
Makefile: Ensure schemas dir exists

aminvakil:
Since https://github.com/micheleg/dash-to-dock/commit/2bc44a6addc1c58360b12038ef5a0bec14403548 building package fails as directory does not exist.

https://github.com/aminvakil/aur/actions/runs/26433076564/job/77810069385

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,6 +107,7 @@ ifeq ($(INSTALLTYPE),system)
 	cp -r ./schemas/*.gschema.xml $(SHARE_PREFIX)/glib-2.0/schemas
 	cp -r ./_build/locale/* $(SHARE_PREFIX)/locale
 endif
+	mkdir -p $(INSTALLBASE)/$(INSTALLNAME)/schemas/
 	cp schemas/gschemas.compiled $(INSTALLBASE)/$(INSTALLNAME)/schemas/
 	-rm -fR _build
 	echo done


aminvakil:
Fixes https://github.com/micheleg/dash-to-dock/issues/2602.

vanvugt:
As mentioned in #2602, I don't think this is required. The `_build` rule is already guaranteed to make the schemas subdirectory and the `install-local` rule is guaranteed to copy it. It sounds like you were starting with an unclean source tree so I think the bug will go away after you do `make clean`.

aminvakil:
No, this is a clean build, this is getting tested in GHA every night

https://github.com/aminvakil/aur/actions/workflows/schedule.yml

and starts to fail exactly since https://github.com/micheleg/dash-to-dock/commit/2bc44a6addc1c58360b12038ef5a0bec14403548 commit.

This is how package is getting built:
https://github.com/aminvakil/aur/blob/master/gnome-shell-extension-dash-to-dock-git/PKGBUILD

vanvugt:
Oh I see, you are using `$(INSTALLTYPE),system`

aminvakil:
> Oh I see, you are using `$(INSTALLTYPE),system`

Ah, so my fix should not be in `install-local` then?

vanvugt:
Replaced by #2605.

PR #2605, opened 2026-06-01
Don't install gschemas.compiled for $(INSTALLTYPE)==system

### LINKED ISSUE #2602: Build breaks after commit 2bc44a6 from two days ago
https://github.com/micheleg/dash-to-dock/commit/2bc44a6addc1c58360b12038ef5a0bec14403548

On line 106 you introduced an if statement I think, this leads to a directory non-existing error in a certain case. My build commands:

```bash
export SASS=sassc
cd "git-checkout"
make VERSION="1.2.3.4" _build
```

Error:

```
cp: cannot create regular file '/tmp/makepkg.builddir/gnome-shell-extension-dash-to-dock-git/pkg/gnome-shell-extension-dash-to-dock-git/usr/share/gnome-shell/extensions/dash-to-dock@micxgx.gmail.com/schemas/': Not a directory
make: *** [Makefile:108: install-local] Error 1
```

vanvugt:
Because $(INSTALLTYPE)==system starts by deleting the target directory so our copy would fail.

Closes: https://github.com/micheleg/dash-to-dock/issues/2602
Fixes: 2bc44a6add ("Makefile: Remember gschemas.compiled for local installs")

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,8 +106,9 @@ ifeq ($(INSTALLTYPE),system)
 	mkdir -p $(SHARE_PREFIX)/glib-2.0/schemas $(SHARE_PREFIX)/locale
 	cp -r ./schemas/*.gschema.xml $(SHARE_PREFIX)/glib-2.0/schemas
 	cp -r ./_build/locale/* $(SHARE_PREFIX)/locale
-endif
+else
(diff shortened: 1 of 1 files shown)


aminvakil:
Continuation of https://github.com/micheleg/dash-to-dock/pull/2600:

Package gets built correctly with this branch.

PR #2606, opened 2026-06-03
appIcons: Fix DEFAULT dot position

sergio-costas:
Until now, the dot vertical position was -12, inherited from code from Gnome Shell. But with the current layout, that value is wrong and must be zero to ensure that the dot is placed outside the icon, below it.

This patch fixes it.

[LP: #2155129](https://bugs.launchpad.net/ubuntu/+source/gnome-shell-ubuntu-extensions/+bug/2155129)


vanvugt:
It might have been simpler to say this was a cherry-pick of 85d74e08d4c156053972e61f2c28d853118e380e. Also I wouldn't stop to ask for reviews on such cherry-picks.

PR #2607, opened 2026-06-03
Explain why "scroll over icon" doesn't work

### LINKED ISSUE #2374: Desktop scrolling is not working
Scrolling through workspaces on the dock stopped working

sergio-costas:
When the "Fixed icon size" setting in the first tab is enabled, the "Scroll action" in the third tab does nothing when scrolling over an application icon, although it still works when using it over the "Show applications" icon.

This PR adds a notification to inform the user about this and avoid misunderstandings and bug reports like https://github.com/micheleg/dash-to-dock/issues/2374 .

Closes: https://github.com/micheleg/dash-to-dock/issues/2374

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -1763,6 +1763,20 @@
                               </object>
                             </child>
                             <child>
+                              <object class="GtkLabel" id="note_about_fixed_size_icon">
+                                <property name="margin_top">5</property>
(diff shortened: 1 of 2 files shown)


vanvugt:
It should probably say `Closes: https://github.com/micheleg/dash-to-dock/issues/2374` even if there's nothing to fix.

PR #2609, opened 2026-06-04
Fix dock refresh

icegood:
during mount/unmount of removable stick

===

i just did to make i work for me. how exactly it should be done - have no idea and 
will not even try to understand code. Generated by AI


vanvugt:
We'll need an explanation of what was wrong and what the fix does.

Also I suspect this fixes some known bug, but which?

3v1n0:
Not acceptable in the current state.

Do proper atomic commits with explaination and feel free to re-open this in case.

PR #2467, opened 2025-11-27
Reimplement Clutter.Color.Shade in RunningIndicatorMetro

alexcanepa:
Fixes #2461. 

It seems that the `_bodyColor` returned by `get_theme_node` is now `Cogl.Color` rather than `Clutter.Color`. 

This PR re-implements [the functionality of `Clutter.Color.shade`](https://github.com/ebassi/clutter/blob/1227f7c489e838c963aaf76328d4b381c51f9fe1/clutter/clutter-color.c#L376-L392) to define `blackenedColor` and `darkenedColor` as before.

Tested without issues on GNOME 49.2.

PR #2615, opened 2026-06-23
Community Edition v106: merge 5 bug fixes + dock margin feature

OtavioAVBelchior:
Bug fixes (upstream PRs):
- intellihide: fix overlap check permanently stuck on GNOME 48 (PR #2552)
  - try-catch in timeout callback, null guards for workspace/compositor
  - prevent duplicate signal handlers, clear stale timeout on enable()
- Fix autohide not working when panel mode is disabled (PR #2511)
  - addChrome now always passes trackFullscreen regardless of dock-fixed
- Fix 2-4px icon offset in fixed dock mode (PR #2506)
  - consolidate margin onto #dash directly instead of child elements
- Fix Metro running indicator crash on GNOME 49 (PR #2467)
  - Clutter.Color.shade removed; reimplemented via Cogl.Color.init_from_hsl
- Prevent wl-clipboard from appearing in dock on Wayland (PR #2244)

New feature:
- Dock margin size customization (PR #2390)
  - Settings slider (0-300 px) when panel/extended mode is active
  - New GSettings key: dock-margin-size
  - Enabled only when extend-height is on

Version bumped to 106.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,53 +1,87 @@
-# Dash to Dock
+# Dash to Dock — Community Edition
+
 ![screenshot](https://github.com/micheleg/dash-to-dock/raw/master/media/screenshot.jpg)
 
-## A dock for the GNOME Shell
-This extension enhances the dash moving it out of the overview and transforming it in a dock for an easier launching of applications and a faster switching between windows and desktops without having to leave the desktop view.
+> A maintained fork of [dash-to-dock](https://github.com/micheleg/dash-to-dock) incorporating community bug fixes and feature contributions that have been awaiting review upstream.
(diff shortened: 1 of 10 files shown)


dmzoneill:
This bundles #2511 and #2552, both of which relate to #2617. The addChrome fix from #2511 changes the same Chrome registration path that causes panelBox to become invisible on GNOME 50, and the intellihide fix from #2552 prevents the overlap check from freezing, which would otherwise block visibility recovery. Neither fully addresses the GNOME 50 regression in #2617, but both improve the surrounding robustness.

PR #2645, opened 2026-07-10
Add per-monitor dock position overrides

unixman69:
## What this does

With multi-monitor docks enabled, every dock currently sits on the same edge (`dock-position`) of its monitor. This PR lets each monitor override that: in **Position and size**, below *Show the dock on*, there is one row per connected monitor with a dropdown — **Follow global setting / Top / Right / Bottom / Left**.

Motivating setup: a laptop panel on the left of an external monitor. The dock on the laptop makes sense on the left edge, but on the external monitor the right edge is where the user actually is.

## Design

- New `monitor-positions` key of type `a{ss}`, mapping **monitor connector names** (e.g. `DP-1`) to `TOP`/`RIGHT`/`BOTTOM`/`LEFT`. Keying by connector (same approach as `preferred-monitor-by-connector`) means overrides survive unplug/replug and never leak onto other monitors. The default `{}` changes no behavior.
- `Utils.getPosition()` gains an optional `monitorIndex` argument, threaded through from all position-dependent code paths (dock placement, theming, running-indicator sides, app icon menus, window previews, tooltips). Callers that pass nothing behave exactly as before.
- The parsed map comes from `DockManager.settings.monitorPositions` via the existing `_mapSettingsValues()` machinery, so there is no new settings plumbing, caching, or cleanup; docks are rebuilt on `changed::monitor-positions` like the other layout keys.
- A monitor index of -1 ("no monitor", e.g. from `findIndexForActor`) never matches an override, including overrides whose connector is currently unplugged (`get_monitor_for_connector` also returns -1 for those).
- Overrides are applied as-is, without the RTL left/right swap: the user picks a physical edge of a specific monitor. This is documented in the schema description.
- The existing *Position on screen* label is renamed to **Global position on screen** so its relationship with the per-monitor rows is explicit. (This changes one translatable string; existing translations of it will fall back to English until catalogs are updated.)
- Drive-by fix: `ThemeManager._adjustTheme` passed a stray `settings` argument to the previously parameterless `Utils.getPosition()`.

## Testing

Developed and daily-driven on GNOME Shell 50.3 (Wayland, CachyOS) on a two-monitor setup (laptop eDP-1 dock on the left, external DP-4 dock on the right), exercising dock placement, intellihide, theming/transparency, app icon menus, window previews and tooltips on both monitors, plus override add/change/remove from the prefs UI and monitor hotplug.

## Disclosure

This code was written by AI (Anthropic Claude, model Claude Fable 5, `claude-fable-5`, running in Claude Code) under my direction and review, and is stated as such in the commit message. Happy to adjust anything — naming, schema shape, UI placement — to fit the project's preferences.

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -556,7 +556,7 @@
                                 <property name="can_focus">0</property>
                                 <property name="hexpand">1</property>
                                 <property name="xalign">0</property>
-                                <property name="label" translatable="yes">Position on screen</property>
+                                <property name="label" translatable="yes">Global position on screen</property>
 
                               </object>
                             </child>
--- a/appIconIndicators.js
+++ b/appIconIndicators.js
@@ -149,7 +149,7 @@ class RunningIndicatorBase extends IndicatorBase {
     constructor(source) {
         super(source);
 
-        this._side = Utils.getPosition();
+        this._side = Utils.getPosition(this._source.monitorIndex);
         this._dominantColorExtractor = new DominantColorExtractor(this._source.app);
         this._signalsHandler.add(this._source, 'notify::running', () => this.update());
         this._signalsHandler.add(this._source, 'notify::focused', () => this.update());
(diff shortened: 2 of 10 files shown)


3v1n0:
I don't think the added complexity is worth to maintain at the moment, unless there are many users affected by this.

Issue #2377, opened 2025-03-21
Issue from #2145 persists; focus, minimize or app spread/preview issues


vari4bled:
**Prerequisite**

-     Behavior > Click action > "Focus, minimize or preview" or "Focus, minimize or app spread" is set
-     There is only one window visible and focused on the desktop (either only one window is open, or all other applications/windows are minimized)
-     Click on the focused window's dock icon to minimize. Then, click on the icon again to focus the window.

**Expected Behavior**

-     When the app is minimized, the dock icon's "highlight" state is removed. (App is not focused anymore)
-     When the icon is clicked again, the window focuses and becomes visible again.

**Bugged Behavior**

-     The icon stays in the 'window is focused' state, as made apparent by the icon remaining highlighted.
-     When the icon is clicked again, the minimized window is not focused. Window does not become visible.

**What I tried**

-     Disabling all other extensions, then restarting computer
-     Resetting dash to dock settings, only applying the click action behavior

**Any error logs?**

-     Looking glass: No
-     journalctl: No

**What am I running?**
dash-to-dock v100(also checked earlier versions)
dconf dump /org/gnome/shell/extensions/dash-to-dock/
```
[/]
apply-custom-theme=false
background-opacity=0.80000000000000004
click-action='focus-minimize-or-previews'
dash-max-icon-size=64
dock-position='BOTTOM'
height-fraction=0.90000000000000002
preferred-monitor=-2
preferred-monitor-by-connector='DP-2'
transparency-mode='DYNAMIC'
```

**The reason:** 
When you minimize with no apps in focus(everything is minimized/show desktop) the last minimized app remains in focus(this.focused = 1) , so Main.activateWindow(w); wont trigger. Sometimes this takes 2 attempts to minimize the same app with nothing else showing for some reason for the focused to not flip. It would be great if minimized state would be tracked with a variable. With a dirty solution I got the expected behavior:

**Dirty solution**
```
            case clickAction.FOCUS_MINIMIZE_OR_PREVIEWS:
                if (this.focused && !hasUrgentWindows) {
                    if (windows.length > 1 || modifiers || button !== 1)
                        this._windowPreviews();
                    else if (!Main.overview.visible) {
                        this._minimizeWindow();
                        this.focused = 0;
                        }
                } else {
                    // Activate the first window
                    const [w] = windows; 
                    Main.activateWindow(w);
                    this.focused = 1;
                }
                break;

```
**Issues with dirty solution:**
[...]

PR #2647, opened 2026-07-12
appIcons: fix stale cached focus state breaking click actions

### LINKED ISSUE #2468: Left click to minimize/focus doesn't work with a single window open
As the title says, if you have only a single window open in a workspace and it's minimized, you can't focus it using the dock icon. It doesn't react to getting clicked on at all. Occasionally, it also won't let you minimize the window through the dock instead.

https://github.com/user-attachments/assets/b619c923-8981-44f0-adba-12ca1bff3fc0
(I'm attempting to focus the window by clicking on it's dock icon in the video)

This bug appeared after I upgraded to Fedora 43 and Gnome 49 (I have a multi-monitor setup). I'm currently running version 102.

I've tried downgrading back to versions 101 and 100, but the bug persists in both of them. I've also tried resetting the settings, but that didn't fix it either.

### My settings
```
[/]
animation-time=0.20000000000000001
apply-custom-theme=false
autohide-in-fullscreen=false
background-opacity=0.80000000000000004
click-action='minimize'
custom-background-color=false
custom-theme-shrink=false
dash-max-icon-size=44
dock-fixed=false
dock-position='BOTTOM'
extend-height=false
height-fraction=0.90000000000000002
hide-delay=1.3877787807814457e-17
hide-tooltip=false
icon-size-fixed=false
intellihide-mode='ALL_WINDOWS'
middle-click-action='previews'
multi-monitor=true
preferred-monitor=-2
preferred-monitor-by-connector='DP-3'
pressure-threshold=0.0
preview-size-scale=0.0
shift-click-action='minimize'
shift-middle-click-action='launch'
show-apps-at-top=true
show-mounts=false
show-mounts-network=false
show-mounts-only-mounted=true
show-trash=false
```

### Errors/Warnings
The ```journalctl /usr/bin/gnome-shell -f``` output is this:
```
Nov 27 23:49:47 potato gnome-shell[8088]: Can't update stage views actor dashtodockContainer [Gjs_dash-to-dock_micxgx_gmail_com_docking_DashToDock] is on because it needs an allocation.
Nov 27 23:49:47 potato gnome-shell[8088]: Can't update stage views actor unnamed [Gjs_dash-to-dock_micxgx_gmail_com_docking_DashSlideContainer] is on because it needs an allocation.
Nov 27 23:49:47 potato gnome-shell[8088]: Can't update stage views actor dashtodockBox [StBoxLayout] is on because it needs an allocation.
Nov 27 23:49:47 potato gnome-shell[8088]: Can't update stage views actor bms-dash-backgroundgroup [MetaBackgroundGroup] is on because it needs an allocation.
Nov 27 23:49:47 potato gnome-shell[8088]: Can't update stage views actor bms-dash-blurred-widget [StWidget] is on because it needs an allocation.
Nov 27 23:49:47 potato gnome-shell[8088]: Can't update stage views actor dashtodockContainer [Gjs_dash-to-dock_micxgx_gmail_com_docking_DashToDock] is on because it needs an allocation.
Nov 27 23:49:47 potato gnome-shell[8088]: Can't update stage views actor unnamed [Gjs_dash-to-dock_micxgx_gmail_com_docking_DashSlideContainer] is on because it needs an allocation.
Nov 27 23:49:47 potato gnome-shell[8088]: Can't update stage views actor dashtodockBox [StBoxLayout] is on because it needs an allocation.
```

unixman69:
## What this fixes

Fixes #2468 and #2377: click actions (most visibly `minimize`) intermittently do nothing, until the app loses and regains focus.

## Root cause

The click actions rely on the icon's cached `focused` flag, computed in `_updateFocusState()` as `tracker.focus_app === this.app && this.running`. During app startup the tracking events can arrive in an order that leaves it permanently stale:

1. The icon is created while the app is `STARTING` (`running = false`).
2. `notify::focus-app` fires when the new window takes focus — `tracker.focus_app` is right, but `running` is still `false`, so `focused` is computed `false`.
3. `notify::state` fires for `STARTING → RUNNING`, but its handler only called `_updateRunningState()` — nothing re-evaluates `focused`.

The icon now believes the app is unfocused. Clicking it takes the activate branch, which is a no-op on an already-focused window, and since that doesn't change `tracker.focus_app`, no event ever heals the flag. Only a real focus change (switching to another window) fixes it — matching the reproduction steps in both issues.

## The fix

- **Root cause**: recompute the focus state in the `notify::state` handler, since `focused` depends on `running`.
- **Defense in depth**: refresh the icon's cached state once at the top of `activate()`, so any other stale path (e.g. the minimized-single-window case reported in #2468, and the workaround suggested in the #2377 discussion) self-corrects at click time. This is one extra state computation per click, so the cost is negligible.
- **Drive-by fix** in the same code path: the double-click detection in the `MINIMIZE` action tested `Clutter.EventType.CLUTTER_BUTTON_PRESS`, which doesn't exist in GJS (the member is `BUTTON_PRESS`), so it was always falsy and `event.get_click_count()` was never called. It now checks the current event's type.

## Testing

Bug reproduced reliably on GNOME Shell 50.3 (Wayland, two-monitor setup) with `click-action=minimize`: open GNOME Settings for the first time, click its dock icon — nothing happens until you switch focus away and back, after which minimize works. The patched build is shipping in a fork I maintain and is being verified on the same setup; I'll report back on this PR if anything shakes out.

## Disclosure

This code was written by AI (Anthropic Claude, model Claude Fable 5, `claude-fable-5`, running in Claude Code) under my direction and review, and is stated as such in the commit message.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -137,7 +137,13 @@ export const DockAbstractAppIcon = GObject.registerClass({
         }
 
         this._signalsHandler.add(this.app, 'windows-changed', () => this._updateWindows());
-        this._signalsHandler.add(this.app, 'notify::state', () => this._updateRunningState());
+        this._signalsHandler.add(this.app, 'notify::state', () => {
+            this._updateRunningState();
+            // The focus state depends on the running state, and may have
+            // been computed while the app was still starting: recompute it
+            // now that the state changed (e.g. STARTING -> RUNNING).
+            this._updateFocusState();
+        });
         this._signalsHandler.add(global.display, 'window-demands-attention', (_dpy, window) =>
             this._onWindowDemandsAttention(window));
         this._signalsHandler.add(global.display, 'window-marked-urgent', (_dpy, window) =>
@@ -541,6 +547,12 @@ export const DockAbstractAppIcon = GObject.registerClass({
             break;
         }
 
+        // The cached running/focused/urgent state can be stale if the
+        // tracking events arrived in an unlucky order while the app was
+        // starting: recompute it at click time so the action always
+        // matches the actual state (see issues #2377 and #2468).
+        this._updateState();
+
         // We check if the app is running, and that the # of windows is > 0 in
         // case we use workspace isolation.
         const windows = this.getInterestingWindows();
@@ -565,7 +577,7 @@ export const DockAbstractAppIcon = GObject.registerClass({
                         // minimize all windows on double click and always in
                         // the case of primary click without additional modifiers
                         let clickCount = 0;
-                        if (Clutter.EventType.CLUTTER_BUTTON_PRESS)
+                        if (event?.type() === Clutter.EventType.BUTTON_PRESS)
                             clickCount = event.get_click_count();
                         const allWindows = (button === 1 && !modifiers) || clickCount > 1;
                         this._minimizeWindow(allWindows);


unixman69:
Follow-up on testing: the patch has now been running on GNOME Shell 50.3 (Wayland, two-monitor setup) via the fork, and the reproduction from #2468/#2377 is confirmed fixed — a freshly started, focused app now minimizes on the first click of its dock icon, with no regressions observed in the other click actions so far.

appIcons.js
```diff
@@ -565,7 +577,7 @@ export const DockAbstractAppIcon = GObject.registerClass({
                         // minimize all windows on double click and always in
                         // the case of primary click without additional modifiers
                         let clickCount = 0;
-                        if (Clutter.EventType.CLUTTER_BUTTON_PRESS)
+                        if (event?.type() === Clutter.EventType.BUTTON_PRESS)
```
3v1n0, line 580: This is valid, but needs a separated commit

3v1n0: Actually no, `get_click_count` does not exists since 2021 (muttter commit 3d37602c5455b834c20d0ddb4e1ce8614a467af2), so this is just dead code.

appIcons.js
```diff
@@ -137,7 +137,13 @@ export const DockAbstractAppIcon = GObject.registerClass({
         }
 
         this._signalsHandler.add(this.app, 'windows-changed', () => this._updateWindows());
-        this._signalsHandler.add(this.app, 'notify::state', () => this._updateRunningState());
+        this._signalsHandler.add(this.app, 'notify::state', () => {
+            this._updateRunningState();
+            // The focus state depends on the running state, and may have
+            // been computed while the app was still starting: recompute it
+            // now that the state changed (e.g. STARTING -> RUNNING).
```
3v1n0, line 144: Sure, then we should just do:

```js
_updateFocusState() {
        this._updateRunningState();
        this.focused = tracker.focus_app === this.app && this.running;
    }
```

No?

PR #2649, opened 2026-07-14
metadata: Enable support for GNOME 51

vanvugt:
It appears to work with 51.alpha so let's get started.


sergio-costas (approved):
Let's go!

PR #2651, opened 2026-07-22
appIcons: fix double-click detection and GNOME 50 click-count removal

unixman69:
The minimize-all-on-double-click branch checked
Clutter.EventType.CLUTTER_BUTTON_PRESS, which is not a real enum member and is therefore always falsy, so the click count was never read and the double-click path was dead. Icon activation is delivered through St.Button's clicked signal on button release, so the current event is a BUTTON_RELEASE; read the click count from press and release events.

Additionally, GNOME 50 removed Clutter.Event.get_click_count(), where the unconditional call threw a TypeError that aborted the whole click handler and silently broke every click action (including minimize). Call it optionally so GNOME 50 falls back to single-click semantics while older shells keep full double-click detection.

### DIFF
--- a/appIcons.js
+++ b/appIcons.js
@@ -565,8 +565,19 @@ export const DockAbstractAppIcon = GObject.registerClass({
                         // minimize all windows on double click and always in
                         // the case of primary click without additional modifiers
                         let clickCount = 0;
-                        if (Clutter.EventType.CLUTTER_BUTTON_PRESS)
-                            clickCount = event.get_click_count();
+                        // Icon activation is driven by St.Button's clicked
+                        // signal, emitted on button release, so the current
+                        // event is a BUTTON_RELEASE: the previous check for
+                        // the non-existent Clutter.EventType.CLUTTER_BUTTON_PRESS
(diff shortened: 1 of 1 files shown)


appIcons.js
```diff
@@ -565,8 +565,19 @@ export const DockAbstractAppIcon = GObject.registerClass({
...
+                        // the non-existent Clutter.EventType.CLUTTER_BUTTON_PRESS
+                        // was always falsy, leaving the click count at zero.
+                        // Read it from press and release events, and call
+                        // get_click_count() optionally since GNOME 50 removed
+                        // it from Clutter.Event (there a plain click still
+                        // minimizes, only double-click detection is skipped).
+                        const eventType = event?.type();
+                        if (eventType === Clutter.EventType.BUTTON_PRESS ||
+                            eventType === Clutter.EventType.BUTTON_RELEASE)
+                            clickCount = event.get_click_count?.() ?? 0;
```
3v1n0, line 580: As said this API does not exist anymore.

While this may need a fix, this is not the right one.

3v1n0:
Closing as it was handled via #2685 and #2703

Issue #2535, opened 2026-03-13
Unusual popup when using wl-copy (wl-clipboard)


rijwan-dev:
Whenever I use wl-clipboard a window is popup and the dock appears for a moment in response to that. wl-clipboard create a gui window for it's functioning. see [Issue](https://github.com/bugaevc/wl-clipboard/issues/90), this issue might be discussed multiple time here as well.......

While the window is fast and not visible. But dock appears again and again whenever I yank something in my nvim editor the dock popups even when I am in full screen.
Is it possible to make the dash-to-dock aware of wl-clipboard window and ignore it completely and not draw it on the dock?
If it is possible I kindly request, please do so. 😢

pradt2:
I'm guessing it's because wl-clipboard shows as a new icon in the dock, which would make sense since to Dash to Dock it's technically  just another newly open app. I experience the same behaviour using another clipboard manager - windows11-clipboard-manager

My workaround is to pin the clipboard to the dock. This stops the dock from appearing when I open the clipboard manager.

3v1n0:
I think those windows should use the [wayland tagging](https://wayland.app/protocols/xdg-toplevel-tag-v1#xdg_toplevel_tag_manager_v1:request:set_toplevel_tag) to provide us some information to ignore them.

Given they cannot define the windows type, they could be tagged so that we and mutter in general can recognize and threat them accordingly.

So please, open bugs upstream about.

rijwan-dev:
> I'm guessing it's because wl-clipboard shows as a new icon in the dock, which would make sense since to Dash to Dock it's technically just another newly open app. I experience the same behaviour using another clipboard manager - windows11-clipboard-manager
> 
> My workaround is to pin the clipboard to the dock. This stops the dock from appearing when I open the clipboard manager.

And how did you pinned? The icon appear so rapid and disappear. Even you run wl-clipboard in loop it'll disappear rapidly and unable to pin.

pradt2:
In my case, as long as I keep the clipboard manager window open, the icon stays in the dock, and I can just navigate to the dock, right-click on the icon, and pin it

NN1985:
FYI, I fixed this temporarily as a workaround by swapping wl-clipboard for 'xclip'.

PR #2652, opened 2026-07-22
intellihide: ignore wl-clipboard helper windows

unixman69:
wl-clipboard maps a short-lived invisible window on every clipboard access. It briefly becomes the top window and blips the focus-app tracking, so the focused window is filtered out of the overlap set and the dock flashes in and out whenever an application polls the selection (for example a terminal running a tool that watches the clipboard). Exclude such helper windows from intellihide entirely, like the existing DING and DropDownTerminal special cases.


3v1n0:
See previous discussions: https://github.com/micheleg/dash-to-dock/issues/2535#issuecomment-4244000138 and https://github.com/micheleg/dash-to-dock/pull/2244

PR #2667, opened 2026-08-19
perf: skip redundant dock geometry updates

Dank-del:
Avoids recalculating the dock static box, Intellihide target, and visible-desktop margins when repeated allocation notifications report identical geometry. This preserves behavior while reducing avoidable work during layout churn. Validated with ESLint and a complete local extension build.

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -262,6 +262,7 @@ const DockedDash = GObject.registerClass({
         // this store size and the position where the dash is shown;
         // used by intellihide module to check window overlap.
         this._staticBox = new Clutter.ActorBox();
+        this._staticBoxGeometry = null;
 
         // Initialize pressure barrier variables
         this._canUsePressure = false;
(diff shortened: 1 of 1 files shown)


vanvugt:
Sounds reasonable, but I also wonder if it's a case of premature optimization. We're assuming  that avoiding:
```
this._intellihide.updateTargetBox(this._staticBox);
this._updateVisibleDesktop();
```
has a meaningful performance benefit. Has there been an observed or measured benefit of the change?

docking.js
```diff
@@ -1260,12 +1261,19 @@ const DockedDash = GObject.registerClass({
...
-        );
+        const x = this.x + this._slider.x -
+            (this._position === St.Side.RIGHT ? this._box.width : 0);
+        const y = this.y + this._slider.y -
+            (this._position === St.Side.BOTTOM ? this._box.height : 0);
+        const {width, height} = this._box;
+        const geometry = this._staticBoxGeometry;
+
+        if (geometry && geometry.x === x && geometry.y === y &&
+            geometry.width === width && geometry.height === height)
```
vanvugt, line 1272: I would have kept `geometry` and `_staticBoxGeometry` cast as `Clutter.ActorBox` so we can use `clutter_actor_box_equal` here, but won't block on it.

Issue #1906, opened 2023-01-24
when shutdown primary monitor switch to second monitor


i74ifa:
> Distro: fedora 37
> GNOME: 42
> dask-to-dock: latest version v75

i have laptop with monitor and i installed the dash-to-dock 
configure shothe dock on monitor but if shutdown monitor the dock switch to screen laptop 
and when trun on monitor not switch dock it

Issue #2335, opened 2024-11-22
Dock only appears on one monitor on boot, and on a single-monitor system it doesn't show.


KatieTheDev:
When booting into Fedora Silverblue with Dash to Dock and Dash to Panel, the dock only shows on my secondary monitor. On my laptop with only one screen, it doesn't show at all. If I disable and re-enable the Dash to Dock extension, it shows correctly. This is really annoying.

dukrous:
This is also happening to me on a vanilla Arch install and Dash to Dock v99 on a single monitor system. Once I disable and re-enable, it shows up again and works flawlessly.

KatieTheDev:
Me too. 
-------- Original Message --------On 12/27/24 19:41, Javier Heredia  wrote:
This is also happening to me on a vanilla Arch install and Dash to Dock v99 on a single monitor system. Once I disable and re-enable, it shows up again and works flawlessly.

—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>
[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/micheleg/dash-to-dock/issues/2335#issuecomment-2564147421",
"url": "https://github.com/micheleg/dash-to-dock/issues/2335#issuecomment-2564147421",
"name": "View Issue"
},
"description": "View this Issue on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

sergio-costas:
Mmm... let me see if I get it right:

- Do you have both Dash to dock and Dash to panel extensions enabled at the same time?
- Are you using in both tests the same laptop, having it connected to an external monitor during the first case, and without it in the second case?
- Are you rebooting between cases, or just disconnect the external monitor?

dukrous:
I only have Dash to Dock enabled, so no to the first two questions. I am rebooting between cases. I did find a workaround and that was enable "show on all monitors".

vanvugt:
Does this command give results that make sense to your setup?
```
gsettings list-recursively org.gnome.shell.extensions.dash-to-dock | grep preferred-monitor
```
It should show `preferred-monitor` as `-2` and `preferred-monitor-by-connector` as the Wayland/DRM connector ID, which you can check using `wayland-info` or `weston-info`.

[...]

Issue #2009, opened 2023-05-16
Feature Request: Add option to show dock always on the main display


dominikzogg:
It would be great if i could chose to put the dock always on the main display (defined in gnome)

cassidyjames:
it looks like this feature is _supposed_ to exist (based on the `/org/gnome/shell/extensions/dash-to-dock/preferred-monitor-by-connector` GSetting), but the setting is immediately overridden at runtime with the current display's connector.

This means, for example, that if I connect my laptop to an external `DP-1` display that is set as the primary display by GNOME, the dock remains on the `eDP-1` internal display because the setting was overridden by Dash to Dock at runtime. If I change the setting to `primary`, it is immediately overridden by the _current_ primary display connector, `DP-1`.

when I disconnect, Dash to Dock updates the setting to the new primary display, `eDPI-1` and thus correctly appears on the internal display. However, when I reconnect to the external display and GNOME correctly updates the primary display, Dash to Dock remains on the non-primary `eDP-1` laptop display.

It appears Dash to Dock is over-agressively overriding `/org/gnome/shell/extensions/dash-to-dock/preferred-monitor-by-connector` instead of allowing it to remain `primary`.

dominikzogg:
@cassidyjames i got a theory: do you also got >=2 external monitor of the same type and series?

i am not sure about the code but this looks like there can be more than one matches:
https://github.com/micheleg/dash-to-dock/blob/master/prefs.js#L121

dominikzogg:
https://github.com/micheleg/dash-to-dock/pull/2054

dominikzogg:
@cassidyjames my PR doesn't solve anything, but makes it worse

kaimast:
I can confirm this behavior. This makes it very annoying to use dash to dock with a laptop.

6XGate:
I suspect these issues are related:
#1906
#1958

This seems to be a bit of an issue as of late and I've had to use Dash-to-Panel with setting to mimic Dash-to-Dock.

Issue #1958, opened 2023-03-04
Randomly on a wrong monitor when waking from sleep.


HubKing:
It happens randomly but sometimes when waking up from sleep, DtD is displayed on a wrong monitor. I have to open the settings and change "Show the dock on". 

I don't know why this happens exactly, but my guess is that sometimes the primary monitor gets initialised a little bit later when waking up from sleep, and at that very moment DtD enumerates the monitor and wrongly thinks that there is only the secondary monitor and displays itself on that monitor.

Can't there be a workaround for this, adding a checkbox like "Always show on the primary monitor", and when the primary monitor changes, DtD automatically moves itself to the primary monitor?

6XGate:
I'm having the same issue. Looking at the DConf settings, the old means to control this has been deprecated and no longer used behaves as it use to. The new means is suppose to support a `primary` value, but it won't stick and just gets transformed to the connection of the primary monitor. This means that when your displays change again, it will move again.

Seems there is something causing it to react to the `preferred-monitor-by-connector` and transforms it rather than leaving it on the default value.

HubKing:
I have been changing monitor settings/disabling extensions to figure out the cause of some problem, and it seems now that DtD ALWAYS move to the secondary monitor when waking up from sleep. That is, the bug is always reproducible. 

If you want to investigate the cause of this unwanted move, maybe this is a good chance. I don't know where DtD is outputting logs, so I checked `journalctl /usr/bin/gnome-shell`, but I am not sure sure which lines are from DtD. If you tell me what I should do to find out the cause, please let me know, and I will do those things and report the result.

pled:
Same issue here : always moved secondary monitor after resume.
Debian bookworm, Gnome Shell 43.3, D2D v79, did not occur before.

6XGate:
I also want to add this issue affects both Xorg and Wayland.

rocchidavide:
Any updates about this bug?

hyperdotbat:
Same issue (Arch, GNOME 47.4), the thing is I actually have `Show on All monitors` checked, and looking at the config in Extensions shows that apparently the correct one is selected and is just not respected, this started happening seemingly out of nowhere, I didnt even add new Extensions so there shouldnt be any conflicts, unchecking (doesnt matter if I check it back or not) fixes it until the next reboot / sign out
[...]

PR #2671, opened 2026-09-04
Add option to hide dock when preferred monitor is unavailable

djrarky:
## Summary

Add an optional **"Hide when this monitor is unavailable"** setting for single-monitor dock configurations.

Currently, when the preferred monitor is disconnected or disabled, Dash to Dock falls back to the primary monitor. This adds an option to instead hide the dock until the preferred monitor becomes available again.

The option defaults to disabled, so existing behavior is unchanged.

## Motivation

This is useful for setups where the dock should only appear on a specific display.

For example, on a laptop with an external dock:

- With the laptop display enabled, the dock can be shown on the laptop display.
- When the lid is closed, that display becomes unavailable.
- Instead of moving the dock to an external monitor, the dock can remain hidden.
- When the laptop display becomes available again, the dock is recreated there automatically.

The behavior is based only on monitor availability and does not require any laptop-, lid-, or docking-specific logic.

## Changes

- Add a `hide-if-monitor-unavailable` setting, disabled by default.
- Add a **"Hide when this monitor is unavailable"** checkbox to the monitor settings.
- Do not fall back to the primary monitor when the preferred monitor is unavailable and the option is enabled.
- Recreate the dock automatically when the preferred monitor becomes available again.
- Keep the configured monitor visible in the preferences selector while it is inactive.
- Disable the option when **"Show on all monitors"** is enabled.

## Behavior

| Preferred monitor | Option disabled | Option enabled |
| --- | --- | --- |
| Available | Show dock on preferred monitor | Show dock on preferred monitor |
| Unavailable | Fall back to primary monitor | Hide dock |

## Compatibility

The new setting defaults to `false`, preserving the current behavior for existing users.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -1,7 +1,34 @@
 // -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
 
-import {DockManager} from './docking.js';
+import {DockManager as BaseDockManager} from './docking.js';
+import {Utils} from './imports.js';
 import {Extension} from './dependencies/shell/extensions/extension.js';
+
+class DockManager extends BaseDockManager {
+    _bindSettingsChanges() {
+        super._bindSettingsChanges();
+        this._signalsHandler.add(
+            this._settings,
+            'changed::hide-if-monitor-unavailable',
+            this._toggle.bind(this));
+    }
+
+    _createDocks() {
+        if (this.settings.hideIfMonitorUnavailable &&
+            !this.settings.multiMonitor &&
+            this.settings.preferredMonitor === -2 &&
+            this.settings.preferredMonitorByConnector !== 'primary') {
+            const monitorManager = Utils.getMonitorManager();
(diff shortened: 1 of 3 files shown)


vanvugt:
@djrarky, this sounds like it relates to one or more of #2335, #1958, #2009, #1906. Please check those to see which ones you think are resolved by this pull request.

djrarky:
It's not directly related to an issue, at least none of those I've seen. Just a preference for the dock to be on my laptop monitor and not the big displays when connected to my dock and the lid is shut.

vanvugt:
> It's not directly related to an issue, at least none of those I've seen. Just a preference for the dock to be on my laptop monitor and not the big displays when connected to my dock and the lid is shut.

I'm fairly confident it does resolve at least some of those issues. So we should confirm which ones can be declared closed here.

djrarky:
It resolves #1906, #1958. 
It doesn't fix the others as this isn't set to follow primary display - it's set to follow a single display.

PR #2673, opened 2026-09-07
Fix GNOME 51.rc compatibility

### LINKED ISSUE #2670: Error with gnome 51.rc
Fedora 45 pre release with gnome 51.rc show this error: "ImportError: Unable to load file from: resource:///org/gnome/shell/ui/pointerWatcher.js"

vanvugt:
The `PointerWatcher` class and `vertical` property no longer exist in 51.rc.

Closes #2670 and more

### DIFF
--- a/docking.js
+++ b/docking.js
@@ -16,7 +16,6 @@ import {
     Layout,
     Main,
     OverviewControls,
-    PointerWatcher,
     SwitcherPopup,
     Workspace,
     WorkspacesView,
@@ -512,11 +511,7 @@ const DockedDash = GObject.registerClass({
         // Remove existing barrier
         this._removeBarrier();
 
-        // Remove pointer watcher
-        if (this._dockWatch) {
-            PointerWatcher.getPointerWatcher()._removeWatch(this._dockWatch);
-            this._dockWatch = null;
-        }
+        this._removeDockWatch();
 
         if (this._optionalScrollWorkspaceSwitchDeadTimeId) {
             GLib.source_remove(this._optionalScrollWorkspaceSwitchDeadTimeId);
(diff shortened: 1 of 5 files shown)


docking.js
```diff
@@ -922,9 +920,9 @@ const DockedDash = GObject.registerClass({
         if (this._autohideIsEnabled &&
             (!Utils.supportsExtendedBarriers() ||
              !DockManager.settings.requirePressureToShow)) {
-            const pointerWatcher = PointerWatcher.getPointerWatcher();
-            this._dockWatch = pointerWatcher.addWatch(
-                DOCK_DWELL_CHECK_INTERVAL, this._checkDockDwell.bind(this));
+            this._dockWatch = Utils.getCursorTracker().connect(
+                'position-invalidated',
+                () => this._checkDockDwell(...global.get_pointer()));
```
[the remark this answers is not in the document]
vanvugt: I had the same concern. Copilot wrote that code.

3v1n0:
We should not remove code that worked in previous versions upstream though, so things should be adjusted depending their existence.

vanvugt:
> We should not remove code that worked in previous versions upstream though, so things should be adjusted depending their existence.

Nothing is removed. Only "moved".

docking.js
```diff
@@ -922,15 +926,26 @@ const DockedDash = GObject.registerClass({
...
 
-    _checkDockDwell(x, y) {
+    _checkDockDwellLater(x, y) {
+        this._checkDockDwellPos = {x, y};
+
+        if (this._checkDockDwellId > 0)
+            return;
+
+        this._checkDockDwellId = GLib.timeout_add(GLib.PRIORITY_DEFAULT,
+            DOCK_DWELL_CHECK_INTERVAL, this._checkDockDwellNow.bind(this));
```
3v1n0:
```suggestion
            DOCK_DWELL_CHECK_INTERVAL, () => this._checkDockDwellNow(x, y));
```

And drop `this._checkDockDwellPos`

docking.js
```diff
@@ -972,6 +987,9 @@ const DockedDash = GObject.registerClass({
             this._cancelDockDwell();
             this._dockDwelling = false;
         }
+
+        this._checkDockDwellId = 0;
+        return GLib.SOURCE_REMOVE;
```
3v1n0: Make this part of the arrow function where the timeout is called, I'd prefer to keep functions independent from the fact they're used in a source or not

PR #2678, opened 2026-09-14
Fix per-monitor rendering on mixed-DPI display sets

### LINKED ISSUE #435: Dock does not scale itself correctly when moved to a monitor with different DPI
I'm using the master branch. I have 2 monitors: My primary one is a 3200x1800 one and Gnome scales everything to 2x there. My secondary one is a 1920x1200 and Gnome uses 1x scale there. However, if I move the dock there, it is still displayed at a 2x scale, instead of 1x.

To make this easier to grasp, I have attached 2 screenshots, with the dock and the settings window side-by-side. The settings window is scaled correctly, according to the DPI of the monitor. The dock is not.

Normal DPI monitor (secondary):
96dpi

HiDPI monitor (primary):
192dpi

valentin-huebner:
## Summary

Fixes #435.

This change scales the components owned by each Dash-to-Dock instance according
to its target monitor. It requests icons at the monitor's native scale and
applies corresponding dimensions to dock styling, badges, scroll fades, and
detached tooltip labels.

The ideal fix would be for GNOME Shell/St to support per-monitor theme contexts,
allowing actors to resolve theme dimensions and rendering resources using the
scale of their stage view. Currently, `St.ThemeContext` is global to the Shell
stage and follows the primary monitor, so a dock displayed on another monitor
inherits the primary monitor's scale.

## Changes

- Request application icons at the target monitor's native scale.
- Apply monitor-specific sizing to dock padding, spacing, margins, badges, and
  scroll fades.
- Scale detached tooltip labels for their corresponding monitor while keeping tooltip positioning in native actor coordinates.

## Testing

Tested on:

- Ubuntu 24.04
- GNOME Shell 46
- Wayland
- One 100% monitor and one 200% monitor
- Dock configured to appear on all monitors

Verified with either monitor selected as primary:

- Dock size remains consistent between monitors.
- Icons are rendered sharply.
- Icon button padding remains proportional.
- Tooltips have the correct size and position.

The upstream port also passes:

- Dart Sass compilation
- `make _build`
- ESLint for all changed JavaScript files

## Architectural limitation

This is necessarily a component-level workaround. Theme dimensions inherited
from GNOME Shell cannot be corrected for an entire dock subtree because St does
not expose a per-subtree or per-stage-view `ThemeContext`.

The stylesheet currently covers Mutter's integer monitor scale factors from
1 through 4. Components added in the future may need their theme-owned metrics
included in the same mechanism.

Feedback on whether this workaround is appropriate upstream would be
appreciated.


3v1n0:
If any  we should rely on `clutter_actor_get_resource_scale` but IIRC there are no cases were we need to do manual intervention, do we?

